PR #9, opened 2020-03-30
Do deeper traversal

### LINKED ISSUE #5: [bug] walk_expression is only called for top level expressions.
`walk_expression` is not being called for function arguments, if statement tests, variable assignments, switch cases and all other non top level expressions.

The same is true for many other properties.

lucacasonato:
Fixes #5 

There are some annoyances with how Rust's default trait method inheritance works. Implementing the one of AstTraverser methods in the rule will override the default trait method, and thus stop walking. Because Rust has no way to call the default trait method from an implementor (https://www.reddit.com/r/rust/comments/3uvfgx/how_to_call_default_trait_method/), the implementor will have to copy and paste the default trait code into their method (or we change the structure of AstTraverser to have a `walk_*` and `super_walk_*` method, where the default of `walk_*` is to call `super_walk_*`), to not stop the traversal. This is especially annoying for expressions, because they can be deeply nested in each other.

Dupe of #7

### DIFF
--- a/src/traverse.rs
+++ b/src/traverse.rs
@@ -165,12 +165,33 @@ pub trait AstTraverser {
     self.walk_identifier(ident)
   }
 
-  fn walk_identifier(&self, ident: Ident) {}
+  fn walk_identifier(&self, ident: Ident) {
+    if let Some(type_ann) = ident.type_ann {
+      self.walk_ts_type_ann(type_ann);
+    }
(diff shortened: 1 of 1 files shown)


bartlomieju:
Thanks @lucacasonato but I completely replace `AstTraverse` with `swc_ecma_visit` in #15

PR #11, opened 2020-03-31
add noEmptyFunction

zhmushan:


src/rules/no_empty_function.rs
```diff
@@ -0,0 +1,26 @@
...
+
+impl NoEmptyFunction {
+  pub fn new(context: Context) -> Self {
+    Self { context }
+  }
+}
+
+impl AstTraverser for NoEmptyFunction {
+  fn walk_fn_decl(&self, fn_decl: FnDecl) {
+    if fn_decl.function.body.unwrap().stmts.is_empty() {
```
bartlomieju, line 18: It will crash if `function.body` is None

zhmushan: so should I ban both "empty body" and "no body"?

PR #21, opened 2020-04-02
[WIP] scopes 2

bartlomieju:


src/rules/no_func_assign.rs
```diff
@@ -0,0 +1,98 @@
...
+impl LintRule for NoFuncAssign {
+  fn new() -> Box<Self> {
+    Box::new(NoFuncAssign)
+  }
+
+  fn lint_module(&self, context: Context, module: swc_ecma_ast::Module) {
+    let mut scope_manager = ScopeVisitor::default();
+    scope_manager.visit_module(&module, &module);
+    let mut visitor = NoFuncAssignVisitor::new(context, scope_manager);
+    visitor.visit_module(&module, &module);
```
bartlomieju, line 23: What I want to achieve here is: do first pass in `ScopeVisitor` to discover all scopes, bindings, references, etc. 
Then do second pass with `NoFuncAssignVisitor` that is `scopeful` - ie. each `visit_*` method would get "current" scope when calling `self.scope_manager.get_current_scope()`. I'm completely out of ideas how to compose those two visitors, @kdy1 please advise.

kdy1: The first way is to use a ring buffer. This was very easy to implement, but it requires visitors to visit blocks and functions in the same order.

The second way is using hashmap with a hygiene context of a span as a key. Using hygiene information as a key means that it is preserved even if range of span changes or some nodes are dropped


That's all I can think of right now. I'll leave more comments if I remember more

bartlomieju: @kdy1 sorry for slow response; I'd like to follow up on it. 

Using hygiene context seems like a better solution for the future (for example auto fixes for lint errors would change the spans). 

Does it mean that I should use [`SyntaxContext`](https://docs.rs/swc_common/0.5.12/swc_common/struct.SyntaxContext.html) as a key in the hash map or am I beating around the wrong bush?

kdy1: @bartlomieju I think I failed to understand requirements, but you can store it on span using `span.apply_mark(mark)` where Span and Mark are from swc_common.

I think using SyntaxContext as a key would work well, but you have to associate the node with marker anyway, and the recommended way to associate the node with syntax context is `span.apply_mark(mark)`.

bartlomieju: @kdy1 thanks for quick response; I'll go through `Resolver` and `Hygiene` in SWC to get better grasp on how they work.

PR #22, opened 2020-04-02
basic ignore directives

### LINKED ISSUE #2: MVP rule ignores
The only way to ignore a lint rule should be inline comment:
```
// dlint-ignore explicitFunctionReturnType
function fooBar() {
  ...
}

// dlint-ignore noExplicitAny explicitFunctionReturnType
function fizzBuzz(a: any) {
  ...
}
```

bartlomieju:
This PR adds simple handling of ignore directives.

I had no better idea so for now it's `// dlint-ignore rule1 rule2`

Closes #2


ry:
I think this would be better
```
// deno:ignore rule1
```

bartlomieju:
@ry I agree, changed to `deno:ignore`

PR #24, opened 2020-04-03
Add getter-return and no-setter-return rules

eliassjogreen:


src/rules/no_setter_return.rs
```diff
@@ -0,0 +1,137 @@
...
+  }
+};
+
+class b {
+  set setterA() {
+    return;
+  }
+  private set setterB() {
+    return;
+  }
```
bartlomieju, line 106: a) Setters should have argument
b) I think this example shouldn't cause error:
```ts
private set setterB(val) {
    if (typeof val === "string") {
      this.stringVal = val;
      return;
    }
    this.numericVal = val;
}
```

bartlomieju:
@eliassjogreen I think we could land this PR

eliassjogreen:
Alright, I have reopened the pr and fixed the conflicts

PR #26, opened 2020-04-24
format and colorize resut

disizali:
Change the result for a better and clearer display

from this ->

before

to this ->

after

colorized and formatted into a table

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,6 +38,18 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


bartlomieju:
@disizali thank you for the PR, but I'm not gonna merge it. I want to get more lint rules working and get good diagnostics from them before introducing additional crates for colorization (we also need to keep Deno in mind and preferably use same dependencies). Let's revisit this topic in some time, thanks again!

Issue #28, opened 2020-05-05
Missing documentation on why not eslint


Raynos:
Would it be possible to document why eslint is not used for linting typescript used in `deno` ?

I would understand the desire to write a linter in rust if `eslint` is deemed too slow.

I would also understand that the deno community does not wish to rely on a binary like `eslint` that only works with `node dist/eslint.js` ( That being I said i see `eslint` as a third_party module ( https://github.com/denoland/deno_third_party/tree/4a3ade332261afb8fcb8b364e59d3cca7c975d36/node_modules/eslint ) )

`eslint` and `typescript-eslint` ( https://github.com/typescript-eslint/typescript-eslint ) have written a lot of linting rules that can be used to provide linting benefits.

It is possible to "bundle up" eslint into a single standalone CLI linter with no configuration file, for example ( https://github.com/Raynos/tsdocstandard ).

bartlomieju:
> Would it be possible to document why eslint is not used for linting typescript used in `deno` ?

It is; this project is still work-in-progress and it is not yet used in deno codebase.

> I would understand the desire to write a linter in rust if eslint is deemed too slow.

:+1: that's the reason for `deno_lint`

> eslint and typescript-eslint ( https://github.com/typescript-eslint/typescript-eslint ) have written a lot of linting rules that can be used to provide linting benefits.

I agree, we plan to implement `recommended` set of rules from eslint.

> It is possible to "bundle up" eslint into a single standalone CLI linter with no configuration file, for example ( https://github.com/Raynos/tsdocstandard ).

I can't really say; haven't tried.

Raynos:
Opened a PR https://github.com/denoland/deno_lint/pull/29 that mentions why `deno_lint` > `eslint`.

PR #29, opened 2020-05-05
Document goal of project

Raynos:
> I would understand the desire to write a linter in rust if eslint is deemed too slow.

> 👍 that's the reason for deno_lint

Closes https://github.com/denoland/deno_lint/issues/28

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,10 @@
 **NOTE**
 Very much work-in-progress
 
+## Goal
+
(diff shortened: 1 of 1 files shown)


bartlomieju:
@Raynos thanks, but I don't want to add such claims to README before confirmation. Once we set up benchmarks to compare against ESLint speed we can revisit the topic.

Raynos:
Thanks.

PR #30, opened 2020-05-14
add eqeqeq rule

g-plane:
This rule will check `==` and `!=`, but there are exceptions: when comparing with `null`, these operators are allowed.


g-plane:
@bartlomieju Done.

Issue #1, opened 2020-03-28
MVP lint rules


bartlomieju:
Decide on list of lint rules that are required for MVP linter.

Good idea to go through recommended sets and compile a list:
- [`eslint`](https://eslint.org/docs/rules/)
- [`typescript-eslint`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.json)

lucacasonato:
I think we should start with all of the recommended rules in the `possible errors` section of `eslint`. These are lints that prevent runtime errors, so they are most useful. Those would be:

- [ ] for-direction
- [ ] getter-return
- [ ] no-async-promise-executor
- [ ] no-compare-neg-zero
- [ ] no-cond-assign
- [ ] no-constant-condition
- [ ] no-control-regex
- [ ] no-debugger
- [ ] no-dupe-args
- [ ] no-dupe-else-if
- [ ] no-dupe-keys
- [ ] no-duplicate-case
- [ ] no-empty
- [ ] no-empty-character-class
- [ ] no-ex-assign
- [ ] no-extra-boolean-cast
- [ ] no-extra-semi
- [ ] no-func-assign
- [ ] no-import-assign
- [ ] no-inner-declarations
- [ ] no-invalid-regexp
- [ ] no-irregular-whitespace
- [ ] no-misleading-character-class
- [ ] no-obj-calls
- [ ] no-prototype-builtins
- [ ] no-regex-spaces
- [ ] no-setter-return
- [ ] no-sparse-arrays
- [ ] no-unexpected-multiline
- [ ] no-unreachable
- [ ] no-unsafe-finally
- [ ] no-unsafe-negation
- [ ] use-isnan
- [ ] valid-typeof

Then later we should implement the `possible errors` recommended rules for typescript and then the remaining recommended `eslint` and typescript specific rules. It would be awesome if we could have support for `eslint:recommended` and `@typescript-eslint/recommended` out of the box.

bartlomieju:
> I think we should start with all of the recommended rules in the possible errors section of eslint. These are lints that prevent runtime errors, so they are most useful. Those would be:

> Then later we should implement the possible errors recommended rules for typescript and then the remaining recommended eslint and typescript specific rules. It would be awesome if we could have support for eslint:recommended and @typescript-eslint/recommended out of the box.

@lucacasonato thanks for compiling the list, presented approach seems to be reasonable. 

[...]

PR #31, opened 2020-05-15
Adding no-dupe-keys Rule

DomParfitt:
Implementation of the [no-dupe-keys](https://eslint.org/docs/rules/no-dupe-keys) rule as per #1.

### DIFF
--- a/src/rules/no_dupe_keys.rs
+++ b/src/rules/no_dupe_keys.rs
@@ -49,9 +49,6 @@ impl Visit for NoDupeKeysVisitor {
       }
     }
 
-    println!("Keys: {:?}", keys);
-    println!("Duplicates: {:?}", duplicates);
-
     if !duplicates.is_empty() {
       self.context.add_diagnostic(
         obj_lit.span,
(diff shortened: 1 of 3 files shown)


src/rules/no_dupe_keys.rs
```diff
@@ -0,0 +1,280 @@
...
+impl NoDupeKeysVisitor {
+  pub fn new(context: Context) -> Self {
+    Self { context }
+  }
+}
+
+impl Visit for NoDupeKeysVisitor {
+  fn visit_object_lit(&mut self, obj_lit: &ObjectLit, _parent: &dyn Node) {
+    let mut keys: HashSet<JsWord> = HashSet::new();
+    let mut duplicates: BTreeSet<JsWord> = BTreeSet::new();
```
bartlomieju, line 40: Nitpick - use `String` instead of `JsWord`

DomParfitt: No problem, I'll put the change in now.

For my own understanding, what's the benefit of `String` over `JsWord`?

bartlomieju: Purely cosmetic; one less foreign type that needs to be looked up

PR #34, opened 2020-05-16
Adding no-unsafe-finally Rule

DomParfitt:
Adds [no-unsafe-finally](https://eslint.org/docs/rules/no-unsafe-finally) as per #1.

### DIFF
--- a/src/rules/no_unsafe_finally.rs
+++ b/src/rules/no_unsafe_finally.rs
@@ -94,6 +94,109 @@ let foo = function() {
   }
 
   #[test]
+  fn it_passes_for_a_return_within_a_function_in_a_finally_block() {
+    test_lint(
+      "no_unsafe_finally",
+      r#"
+let foo = function() {
+  try {
+    return 1;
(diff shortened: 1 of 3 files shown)


src/rules/no_unsafe_finally.rs
```diff
@@ -0,0 +1,240 @@
...
+impl Visit for NoUnsafeFinallyVisitor {
+  fn visit_try_stmt(&mut self, try_stmt: &TryStmt, _parent: &dyn Node) {
+    if let Some(finally_block) = &try_stmt.finalizer {
+      // Convenience function for providing different diagnostic message
+      // depending on statement type
+      let add_diagnostic = |stmt_type: &str| {
+        self.context.add_diagnostic(
+          finally_block.span,
+          "noUnsafeFinally",
+          format!("Unsafe usage of {}Statement", stmt_type).as_str(),
```
bartlomieju, line 39: This message is not very helpful, but it's ok for the first pass

PR #35, opened 2020-05-17
add "no-throw-literal" rule

g-plane:
Ref: https://eslint.org/docs/rules/no-throw-literal

### DIFF
--- a/src/rules/no_throw_literal.rs
+++ b/src/rules/no_throw_literal.rs
@@ -1,0 +1,103 @@
+// Copyright 2020 the Deno authors. All rights reserved. MIT license.
+use super::Context;
+use super::LintRule;
+use swc_atoms::js_word;
+use swc_ecma_ast::{Expr, ThrowStmt};
+use swc_ecma_visit::Node;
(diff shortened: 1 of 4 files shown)


src/rules/no_throw_literal.rs
```diff
@@ -0,0 +1,103 @@
...
+  use super::*;
+  use crate::test_util::test_lint;
+  use serde_json::json;
+
+  #[test]
+  fn check_literal() {
+    test_lint(
+      "no_throw_literal",
+      "throw 'kumiko'",
+      vec![NoThrowLiteral::new()],
```
bartlomieju, line 63: Can you add a few more tests cases for bad usage of throw?

g-plane: Bad usage? Such as other types of literals?

bartlomieju: Yes, a number, null, etc.

g-plane: Done.

PR #36, opened 2020-05-17
add "default-param-last" rule

g-plane:
Ref: https://eslint.org/docs/rules/default-param-last

### DIFF
--- a/src/rules/default_param_last.rs
+++ b/src/rules/default_param_last.rs
@@ -1,0 +1,70 @@
+// Copyright 2020 the Deno authors. All rights reserved. MIT license.
+use super::Context;
+use super::LintRule;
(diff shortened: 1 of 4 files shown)


src/rules/default_param_last.rs
```diff
@@ -0,0 +1,70 @@
...
+mod tests {
+  use super::*;
+  use crate::test_util::test_lint;
+  use serde_json::json;
+
+  #[test]
+  fn default_param_last_test() {
+    test_lint(
+      "default_param_last",
+      "function fn(a = 2, b) {}",
```
bartlomieju, line 57: Please add more test cases for failing scenarios and at least one test case for success scenario

PR #38, opened 2020-05-19
Adding valid-typeof Rule

DomParfitt:
Adds [https://eslint.org/docs/rules/valid-typeof](valid-typeof) rule as per #1.

@bartlomieju the ESLint version of this rule has an option - `requireStringLiterals` - that can be provided to it which acts like a strict-mode of sorts. For example, by default the rule will pick up the following as a lint error due to the misspelling of "undefined":
```javascript
typeof foo === "undefnied"
```
The correct version of the above code being:
```javascript
typeof foo === "undefined"
```
If the `requireStringLiterals` option is not set (or set to `false`) then the following code is also acceptable but is considered a lint error if `requireStringLiterals` is `true`:
```javascript
typeof foo === undefined
```
I have not implemented the functionality provided by `requireStringLiterals` as I don't think it's currently possible in `deno_lint` to provide options to rules and doing so probably falls out of the scope of implementing this particular rule.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,12 @@ jobs:
           rustup component add clippy
           rustup component add rustfmt
 
+      - name: Cache
+        uses: actions/cache@v1
+        with:
+          path: target/
+          key: ${{ matrix.config.os }}-${{ hashFiles('Cargo.lock') }}
+
       - name: Environment
         run: |
           echo ::set-env name=GH_ACTIONS::1
--- a/README.md
+++ b/README.md
@@ -19,7 +19,9 @@ Very much work-in-progress
 - `noEmptyInterface`
 - `noEval`
 - `noExplicitAny`
+- `noNewSymbol`
 - `noSparseArray`
+- `noThrowLiteral`
 - `noVar`
 - `singleVarDeclarator`
 - `useIsNaN`
--- a/README.md
+++ b/README.md
@@ -28,12 +28,12 @@ Very much work-in-progress
 Only single line ignores are supported:
 
 ```ts
-// deno:ignore noExplicitAny
+// deno-lint-ignore noExplicitAny
 function foo(): any {
   // ...
 }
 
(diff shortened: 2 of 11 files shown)


bartlomieju:
> Adds [https://eslint.org/docs/rules/valid-typeof](valid-typeof) rule as per #1.
> 
> @bartlomieju the ESLint version of this rule has an option - `requireStringLiterals` - that can be provided to it which acts like a strict-mode of sorts. For example, by default the rule will pick up the following as a lint error due to the misspelling of "undefined":
> 
> ```js
> typeof foo === "undefnied"
> ```
> 
> The correct version of the above code being:
> 
> ```js
> typeof foo === "undefined"
> ```
> 
> If the `requireStringLiterals` option is not set (or set to `false`) then the following code is also acceptable but is considered a lint error if `requireStringLiterals` is `true`:
> 
> ```js
> typeof foo === undefined
> ```
> 
> I have not implemented the functionality provided by `requireStringLiterals` as I don't think it's currently possible in `deno_lint` to provide options to rules and doing so probably falls out of the scope of implementing this particular rule.

@DomParfitt sounds good. There are currently no options, I for now I don't even consider adding them. If your solution uses more "strict" way of ESLint then I'm good with that :+1:

DomParfitt:
@bartlomieju It currently implements the less strict version from ESLint (which I think is the default for ESLint) but it wouldn't take me much work for this rule to implement the stricter version as the default if you'd prefer.

bartlomieju:
> @bartlomieju It currently implements the less strict version from ESLint (which I think is the default for ESLint) but it wouldn't take me much work for this rule to implement the stricter version as the default if you'd prefer.

Great, sounds good to me 👍

DomParfitt:
@bartlomieju I've made the changes so the rule is now performing the stricter check if you want to re-check the code. 

The behaviour now is that it is only valid to compare a `typeof` with another `typeof` (i.e. `typeof foo === typeofbar`) or a specific set of strings (e.g. `"undefined"`, `"object"`, etc.).

PR #44, opened 2020-05-22
Adds no-cond-assign Rule

DomParfitt:
Adds [no-cond-assign](https://eslint.org/docs/rules/no-cond-assign) as per #1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,12 @@ Very much work-in-progress
 - `eqeqeq`
 - `explicitFunctionReturnType`
 - `noAsyncPromiseExecutor`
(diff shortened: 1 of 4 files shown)


g-plane:
However, the code will also be detected, but it should pass:

```js
if ((x = y)) {}
```

DomParfitt:
> 
> 
> However, the code below will also be detected, but it should pass:
> 
> ```js
> if ((x = y)) {}
> ```

@g-plane, I've added an additional test for this code but it is passing as expected without any modification to the rule implementation. Not sure if I'm missing something, what made you think this case wasn't working as expected?

PR #45, opened 2020-05-22
Adds no-empty Rule

DomParfitt:
Adds [no-empty](https://eslint.org/docs/rules/no-empty) as per #1.

### DIFF
--- a/src/rules/no_empty.rs
+++ b/src/rules/no_empty.rs
@@ -1,5 +1,6 @@
-use super::{LintRule, Context};
-use swc_ecma_ast::{Module, BlockStmt};
+// Copyright 2020 the Deno authors. All rights reserved. MIT license.
+use super::{Context, LintRule};
+use swc_ecma_ast::{BlockStmt, Module};
 use swc_ecma_visit::{Node, Visit};
 
 pub struct NoEmpty;
@@ -54,7 +55,7 @@ if (foo) {
 }
       "#,
       vec![NoEmpty::new()],
-      json!([])
+      json!([]),
     )
   }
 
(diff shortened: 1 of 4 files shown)


DomParfitt:
> 
> 
> @DomParfitt looks great, please add more test cases covering:
> 
> ```
> - while
> - switch
> - try/catch
> - try/catch/finally
> ```

@bartlomieju thanks! I've added tests for the above cases and also for `do-while` and `for` blocks

src/rules/no_empty.rs
```diff
@@ -0,0 +1,314 @@
...
+  fn visit_block_stmt(&mut self, block_stmt: &BlockStmt, _parent: &dyn Node) {
+    if block_stmt.stmts.is_empty()
+      && !block_stmt.contains_comments(&self.context)
+    {
+      self.context.add_diagnostic(
+        block_stmt.span,
+        "noEmpty",
+        "Empty block statement",
+      );
+    }
```
bartlomieju, line 39: There should be call to `swc_ecma_visit::visit_block_stmt()` here; otherwise if block is non-empty its contents won't be visited

DomParfitt: @bartlomieju do you mean to check for nested empty blocks? For example:
```javascript
if (foo) {
  if (bar) {
  }
}
```

bartlomieju: @DomParfitt yes, I have a suspicion that your example wouldn't be picked up right now

DomParfitt: Yep, you're right. Just added a test for the above and it currently doesn't report any linting errors. I'll add a check for nested blocks in.

PR #53, opened 2020-05-29
Basic scope analysis, noFuncAssign, noClassAssign

bartlomieju:
This commit adds basic scope analysis and two lint rules
that require such analysis:
- noFuncAssign
- noClassAssign

Scope analysis visits only a few nodes and needs to be
finished.

### DIFF
--- a/src/scopes.rs
+++ b/src/scopes.rs
@@ -110,10 +110,8 @@ impl ScopeManager {
     self.scope_stack.pop();
   }
 
-  pub fn get_root_scope(&mut self) -> &mut Scope {
-    self
-      .get_scope_mut(*self.scope_stack.first().unwrap())
-      .unwrap()
+  pub fn get_root_scope(&self) -> &Scope {
+    self.get_scope(*self.scope_stack.first().unwrap()).unwrap()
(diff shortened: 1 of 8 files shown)


src/scopes.rs
```diff
@@ -0,0 +1,386 @@
...
+  Switch,
+  With,
+  Catch,
+}
+
+#[derive(Clone, Debug)]
+pub struct Scope {
+  pub kind: ScopeKind,
+  pub id: u32,
+  pub parent_id: Option<u32>,
```
bartlomieju, line 59: It would be much better to hold reference to parent: `Option<&Scope>`

src/scopes.rs
```diff
@@ -0,0 +1,386 @@
...
+  pub span: Span,
+  pub child_scopes: Vec<u32>,
```
bartlomieju, line 61: Ditto, it'd be better to own child scopes: `Vec<Scope>`

bartlomieju:
This PR doesn't contain full implementation but I'm going to land it anyway. I'll be gradually working towards full scope analysis.

PR #56, opened 2020-05-30
Add requireYield rule

bartlomieju:


src/rules/require_yield.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    swc_ecma_visit::visit_fn_expr(self, fn_expr, parent);
+
+    self.check_function(&fn_expr.function);
+  }
+
+  fn visit_class_method(&mut self, class_method: &ClassMethod, parent: &dyn Node) {
+    eprintln!("visit class method");
+    if class_method.function.is_generator {
+      self.yield_stack.push(0);
```
bartlomieju, line 88: Possibly bug in SWC https://github.com/swc-project/swc/issues/808

PR #61, opened 2020-05-30
feat: add constructor_super rule

Qu4k:


Qu4k:
Now the this implementation passes all ESlint tests, not considering those that require code paths

PR #62, opened 2020-05-31
show code snippets

### LINKED ISSUE #47: Show code snippets in report output
Current report is very minimalistic:
```
$ target/debug/dlint test.ts
error: `any` type is not allowed (noExplicitAny) at ./test.ts:6:14
error: `var` keyword is not allowed (noVar) at ./test.ts:12:0
error: Variables shouldn't be deleted (noDeleteVar) at ./test.ts:14:0
```

It should be changed to contain a snippet of code that shows where the problem occurs (much like ESLint):
```
$ target/debug/dlint test.ts
error: `any` type is not allowed (noExplicitAny) at ./test.ts:6:14
1| function foo(a: any) {
 |                  ^
2|   //
3| }
```

@disizali had a PR with colorful output at https://github.com/denoland/deno_lint/pull/26, but I'd rather not introduce new crates for the sake of formatting right now. 

Instead we should copy-paste color helpers from `deno` and construct diagnostic messages manually.

I expect that `Linter` struct will be extended to be able to extract code snippets from from SWC'c source map.

disizali:
fixes #47 
new result : 
```
problem.ts =>
  1| var varName = "var value";
     ^
  (noVar) `var` keyword is not allowed
  18| debugger;
      ^
  (noDebugger) `debugger` statement is not allowed
problem2.ts =>
  1| const sparse = [1,,3];
                    ^
  (noSparseArray) Sparse arrays are not allowed
Found 3 problems
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,7 +42,7 @@ fn main() {
       report_util::report_filename(&file_diagnostics[0].location.filename);
       for d in file_diagnostics.iter() {
         report_util::report_line_src(d.location.line, &d.line_src);
-        report_util::place_glyph(d.location.line, d.location.col) ;   
+        report_util::place_glyph(d.location.line, d.location.col);   
         report_util::report_error(&d.code, &d.message);
       }
     }
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,7 +42,7 @@ fn main() {
       report_util::report_filename(&file_diagnostics[0].location.filename);
       for d in file_diagnostics.iter() {
         report_util::report_line_src(d.location.line, &d.line_src);
-        report_util::place_glyph(d.location.line, d.location.col);   
+        report_util::place_glyph(d.location.line, d.location.col);
         report_util::report_error(&d.code, &d.message);
       }
     }
--- a/src/main.rs
+++ b/src/main.rs
@@ -41,9 +41,9 @@ fn main() {
     if !file_diagnostics.is_empty() {
       report_util::report_filename(&file_diagnostics[0].location.filename);
       for d in file_diagnostics.iter() {
+        report_util::report_error(&d.code, &d.message);
         report_util::report_line_src(d.location.line, &d.line_src);
         report_util::place_glyph(d.location.line, d.location.col);
-        report_util::report_error(&d.code, &d.message);
       }
     }
   }
--- a/src/main.rs
+++ b/src/main.rs
@@ -41,9 +41,9 @@ fn main() {
     if !file_diagnostics.is_empty() {
       report_util::report_filename(&file_diagnostics[0].location.filename);
       for d in file_diagnostics.iter() {
-      report_util::report_line_src(d.location.line, &d.line_src);
-      report_util::place_glyph(d.location.line, d.location.col) ;   
-      report_util::report_error(&d.code, &d.message);
+        report_util::report_line_src(d.location.line, &d.line_src);
+        report_util::place_glyph(d.location.line, d.location.col) ;   
+        report_util::report_error(&d.code, &d.message);
       }
     }
   }
--- a/src/main.rs
+++ b/src/main.rs
@@ -39,11 +39,19 @@ fn main() {
 
     error_counts += file_diagnostics.len();
     if !file_diagnostics.is_empty() {
-      report_util::report_filename(&file_diagnostics[0].location.filename);
       for d in file_diagnostics.iter() {
         report_util::report_error(&d.code, &d.message);
+        report_util::report_location(
+          &file_diagnostics[0].location.filename,
+          d.location.line,
+          d.location.col,
+        );
         report_util::report_line_src(d.location.line, &d.line_src);
-        report_util::place_glyph(d.location.line, d.location.col);
+        report_util::place_glyphes(
+          d.location.line,
+          d.location.col,
+          d.snippet_length,
+        );
       }
     }
   }
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,8 +18,7 @@ fn main() {
 
   let file_names: Vec<String> = args[1..].to_vec();
 
-  let mut diagnostics = vec![];
-
+  let mut error_counts = 0;
   for file_name in file_names {
     let source_code =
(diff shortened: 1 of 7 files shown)


src/test_util.rs
```diff
@@ -14,6 +14,25 @@ pub fn test_lint(
   let diagnostics = linter
     .lint(filename.to_string(), source_code.to_string(), rules)
     .expect("Failed to lint");
-  let serialized_diagnostics = serde_json::to_value(diagnostics).unwrap();
+
+  // TODO(@disizali) refactor this to a better approach.
+  // it's a temporary solution for ignoring line_src field.
```
disizali, line 19: we could write `"line_src"` field on every single expected_diagnostic on each test, but it's redundant IMHO; 
because we do line number assertion anyway.

bartlomieju: @disizali don't worry about testing utility - I'll intend to rewrite it shortly. Any hacks in here are fine 👍

src/main.rs
```diff
@@ -64,16 +63,18 @@ fn main() {
...
+    if !file_diagnostics.is_empty() {
+      eprintln!("{} =>", file_diagnostics[0].location.filename);
+      for d in file_diagnostics.iter() {
+        eprintln!("  {}| {}", d.location.line, d.line_src);
+        eprintln!(
+          "  {}^",
+          " ".repeat(d.location.line.to_string().len() + d.location.col + 2)
+        );
+        eprintln!("  ({}) {}", d.code, d.message);
+      }
```
bartlomieju, line 76: Cool :+1: can you keep the old formatting as was and additionally display code snippet on the next line:

Eg:
```
// now:
▶ target/debug/dlint ../deno/std/http/server.ts
../deno/std/http/server.ts =>
  93|       } catch {}
                    ^
  (noEmpty) Empty block statement
  111|     while ((await body.read(buf)) !== null) {}
                                                   ^
  (noEmpty) Empty block statement
  120|   constructor(public listener: Listener) {}
                                                ^
  (noEmpty) Empty block statement
Found 3 problems
```

```
// new:
▶ target/debug/dlint ../deno/std/http/server.ts
error: Empty block statement (noEmpty) at ../deno/std/http/server.ts:93:0
93|       } catch {}
                    ^
...
Found 3 problems
```

I'll be refactoring AST parser soon so I'll figure a way to obtain code snippet plus surrounding lines

src/main.rs
```diff
@@ -64,16 +63,18 @@ fn main() {
...
-        "error: {} ({}) at {}:{}:{}",
-        d.message, d.code, d.location.filename, d.location.line, d.location.col
-      );
+    error_counts += file_diagnostics.len();
+    if !file_diagnostics.is_empty() {
+      eprintln!("{} =>", file_diagnostics[0].location.filename);
+      for d in file_diagnostics.iter() {
+        eprintln!("  {}| {}", d.location.line, d.line_src);
+        eprintln!(
+          "  {}^",
```
bartlomieju, line 72: Please make this glyph in red color

src/main.rs
```diff
@@ -64,16 +63,18 @@ fn main() {
...
+          " ".repeat(d.location.line.to_string().len() + d.location.col + 2)
+        );
+        eprintln!("  ({}) {}", d.code, d.message);
```
bartlomieju, line 75: please print "code" in grey color

src/main.rs
bartlomieju: Ah I see that you show all errors from a single file collapsed together. That's probably better approach

bartlomieju:
@disizali looking great!

Two nitpicks:
1. remove `=>` after filename - not really relevant
2. I think it would be more readable to display code and message first and problematic line after:
```
  (noEmpty) Empty block statement
  93|       } catch {}
                    ^
```

src/main.rs
```diff
@@ -33,16 +37,15 @@ fn main() {
...
-        d.message, d.code, d.location.filename, d.location.line, d.location.col
-      );
+    error_counts += file_diagnostics.len();
+    if !file_diagnostics.is_empty() {
+      report_util::report_filename(&file_diagnostics[0].location.filename);
+      for d in file_diagnostics.iter() {
+        report_util::report_line_src(d.location.line, &d.line_src);
+        report_util::place_glyph(d.location.line, d.location.col);
+        report_util::report_error(&d.code, &d.message);
+      }
```
nayeemrmn: This is something that bugs me about ESLint's output. Something like
```
../deno/std/http/server.ts =>
  93|       } catch {}
                    ^
  (noEmpty) Empty block statement
```
doesn't have anything to ctrl + click in an integrated terminal to take you straight to the offending line.
If there is an error at `std/http/server.ts`, line 93, column 16, it's conventional to output the following adjacently: `std/http/server.ts:93:16`. Clippy and other Rust errors do this correctly:
```
error: unused import: `std::path::Path`
  --> cli/main.rs:96:5
   |
96 | use std::path::Path;
   |     ^^^^^^^^^^^^^^^
   |
```

bartlomieju: @nayeemrmn I agree, that's been bugging me as well - having a way to link to problematic location is great. It seems like Rust's output gets it right; @disizali sorry to bother, but maybe you could try above format?

disizali: It's my pleasure, I will definitely do it.
I think rust output is really telling too,
I try to be like that

disizali:
thanks for your complete guidance @bartlomieju & @nayeemrmn.

The output of the report now looks like this:

image

bartlomieju:
@disizali thank you! I changed formatting a bit:
Screenshot 2020-06-01 at 12 31 58

disizali:
@bartlomieju  It's great that we have such a beautiful and clear output now.

I'm thinking about `hints` for the next step , it'll be awesome and tricky in the js world 💯.

bartlomieju:
> @bartlomieju It's great that we have such a beautiful and clear output now.
> 
> I'm thinking about `hints` for the next step , it'll be awesome and tricky in the js world 💯.

@disizali yes, `hints/advice` will definitely be useful and we should add them - but first I need to refactor testing setup

PR #71, opened 2020-06-01
refactor: Testing utils

lucacasonato:
@bartlomieju If this seems good then I'll move all of the other tests over to the new utils. I did ultimately keep a line number version of the utils because it seems useful for the multi line tests (the comment related ones for example).

### DIFF
--- a/src/rules/ban_ts_ignore.rs
+++ b/src/rules/ban_ts_ignore.rs
@@ -46,42 +46,31 @@ impl LintRule for BanTsIgnore {
 #[cfg(test)]
 mod tests {
   use super::*;
-  use crate::test_util::test_lint;
-  use serde_json::json;
+  use crate::test_util::*;
 
   #[test]
   fn ban_ts_ignore() {
-    test_lint(
-      "ban_ts_ignore",
+    assert_lint_err_on_line::<BanTsIgnore>(
       r#"
 // @ts-ignore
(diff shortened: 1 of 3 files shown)


src/rules/ban_ts_ignore.rs
```diff
@@ -46,42 +46,31 @@ impl LintRule for BanTsIgnore {
...
-      "ban_ts_ignore",
+    assert_lint_err_on_line::<BanTsIgnore>(
       r#"
 // @ts-ignore
 function foo() {
   // pass
 }
-
+    "#,
+      "banTsIgnore",
```
bartlomieju, line 60: Passing expected code seems superfluous if we pass `BanTsIgnore` struct - this should probably be automated by adding `code()` method to `LintRule` trait; that way we can discover this code automatically. Fine for now, I'll see to it when refactoring `LintRule` trait.

src/test_util.rs
```diff
@@ -38,3 +38,90 @@ pub fn test_lint(
...
+    assert!(diagnostics.is_empty());
+  }
+}
+
+#[allow(dead_code)]
+pub fn assert_lint_err<T: LintRule + 'static>(
+  source: &str,
+  code: &str,
+  col: usize,
+) {
```
bartlomieju, line 89: 👍

PR #73, opened 2020-06-01
use the default trait

utam0k:
This is not that much of a benefit for now, but it's more common and better to use the default trait.


bartlomieju:
@utam0k thank you for opening the PR - it's certainly better to use `Default` trait. That said, `Linter` will be undergoing changes and refactors in coming days and won't have `default()` method anymore. Closing without merge.

PR #75, opened 2020-06-01
add LintRule.code()

bartlomieju:


bartlomieju:
@lucacasonato PTAL I simplified tests a bit

PR #81, opened 2020-06-02
add no_empty_character_class rule

disizali:
this PR is going to add :
+ `no empty character class rule`
+ `3 tests about it`

### DIFF
--- a/src/rules/no_empty_character_class.rs
+++ b/src/rules/no_empty_character_class.rs
@@ -35,10 +35,25 @@ impl NoEmptyCharacterClassVisitor {
 impl Visit for NoEmptyCharacterClassVisitor {
   fn visit_regex(&mut self, regex: &Regex, _parent: &dyn Node) {
     let regex_literal = format!("/{}/", regex.exp);
-    let rule_regex =
-      regex::Regex::new(r"^/([^\[]|\.|\[([^\\\]]|\\.)+\])*/[gimuys]*$")
-        .unwrap();
-    if !rule_regex.is_match(&regex_literal) {
+    lazy_static! {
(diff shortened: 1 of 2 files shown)


src/rules/no_empty_character_class.rs
```diff
@@ -0,0 +1,72 @@
...
+  pub fn new(context: Context) -> Self {
+    Self { context }
+  }
+}
+
+impl Visit for NoEmptyCharacterClassVisitor {
+  fn visit_regex(&mut self, regex: &Regex, _parent: &dyn Node) {
+    let regex_literal = format!("/{}/", regex.exp);
+    let rule_regex =
+      regex::Regex::new(r"^/([^\[]|\.|\[([^\\\]]|\\.)+\])*/[gimuys]*$")
```
bartlomieju, line 39: make this regex a `lazy_static` and please add comment describing this regex as well as linking to eslint rule

src/rules/no_empty_character_class.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+  use crate::test_util::*;
+
+  #[test]
+  fn no_empty_character_class_match() {
+    assert_lint_err::<NoEmptyCharacterClass>(r#"/^abc[]/.test("abcdefg");"#, 0);
```
bartlomieju, line 58: Can you port some more test cases from eslint?

disizali: done

src/rules/no_empty_character_class.rs
disizali: thanks for the notice , done!

PR #98, opened 2020-06-03
add no array constructor rule

disizali:
this PR is going to add :
+ `no array constructor rule`
+ `12 tests about it`


src/rules/no_array_constructor.rs
```diff
@@ -0,0 +1,107 @@
...
+  use super::*;
+  use crate::test_util::*;
+
+  #[test]
+  fn no_array_constructor_call_valid() {
+    assert_lint_ok_n::<NoArrayConstructor>(vec![
+      "Array(x)",
+      "Array(9)",
+      "Array.foo()",
+      "foo.Array()",
```
bartlomieju, line 86: The same rule applies in `typescript_eslint` - in that case it recognizes `Array` constructor with type parameter (`Array<T>`) please handle that situation.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-array-constructor.md

disizali: done

PR #104, opened 2020-06-03
Kebab case everything

### LINKED ISSUE #82: Lint rule code - camelCase or kebab-case
Currently rule codes are written using camelCase:
- `noExplicitAny`
- `noEval`
- `getterReturn`

In ESLint rule codes are written using kebab-case:
- `no-explicit-any`
- `no-eval`
- `getter-return`

Ignore directives are currently spelled out using kebab-case (`deno-lint-ignore` which follows convention from Deno for formatting ignores - `deno-fmt-ignore`).

We should make a decision one way or another and unify all codes.

Probably the best to go with kebab-case as it's already established convention in ESLint

lucacasonato:
Closes #82


src/rules/use_isnan.rs
```diff
@@ -54,7 +54,7 @@ impl Visit for UseIsNaNVisitor {
         if is_nan_identifier(&ident) {
           self.context.add_diagnostic(
             bin_expr.span,
-            "useIsNaN",
+            "use-is-na-n",
```
disizali: Why is only this one different from eslint ?

lucacasonato: Fixed already

PR #117, opened 2020-06-05
add prefer-namespace-keyword rule

disizali:
this PR is going to add :

+ `prefer-namespace-keyword rule`
+ `7 tests about it`


src/rules/prefer_namespace_keyword.rs
```diff
@@ -0,0 +1,99 @@
...
+
+pub struct PreferNamespaceKeyword;
+
+impl LintRule for PreferNamespaceKeyword {
+  fn new() -> Box<Self> {
+    Box::new(PreferNamespaceKeyword)
+  }
+
+  fn code(&self) -> &'static str {
+    "preferNamespaceKeyword"
```
bartlomieju, line 17: `prefer-namespace-keyword`

src/rules/prefer_namespace_keyword.rs
```diff
@@ -0,0 +1,99 @@
...
+      .source_map
+      .span_to_snippet(mod_decl.span)
+      .expect("error in load snippet");
+
+    if let Some(capt) = KEYWORD.captures(&snippet) {
+      let keyword = capt.name("keyword").unwrap().as_str();
+      if keyword == "module" && !mod_decl.global {
+        self.context.add_diagnostic(
+          mod_decl.span,
+          "preferNamespaceKeyword",
```
bartlomieju, line 61: `prefer-namespace-keyword`

src/rules/prefer_namespace_keyword.rs
disizali: done

src/rules/prefer_namespace_keyword.rs
disizali: thanks for the notice , done 👍

PR #127, opened 2020-06-06
Add no-misused-new rule

ardyfeb:


src/rules/mod.rs
```diff
@@ -29,6 +29,7 @@ mod no_eval;
 mod no_ex_assign;
 mod no_explicit_any;
 mod no_func_assign;
+mod no_misused_new;
```
disizali, line 32: can you add this rule in `fn get_all_rules` ?

```rust
no_misused_new::NoMisusedNew::new(),
```

src/rules/no_misused_new.rs
```diff
@@ -0,0 +1,134 @@
+use super::Context;
```
disizali, line 1: add copyright

src/rules/no_misused_new.rs
```diff
@@ -0,0 +1,134 @@
...
+    Box::new(NoMisusedNew)
+  }
+
+  fn lint_module(&self, context: Context, module: Module) {
+    let mut visitor = NoMisusedNewVisitor::new(context);
+    visitor.visit_module(&module, &module);
+  }
+
+  fn code(&self) -> &'static str {
+    "noMisusedNew"
```
disizali, line 25: `no-misused-new`

src/rules/no_misused_new.rs
```diff
@@ -0,0 +1,134 @@
...
+          if let Expr::Ident(ident) = &*signature.key {
+            if JsWord::from("constructor") == ident.sym
+              && signature.type_ann.is_some()
+              && self
+                .match_parent_type(&n.id, &signature.type_ann.as_ref().unwrap())
+            {
+              // constructor
+              self.context.add_diagnostic(
+                signature.span,
+                "noMisusedNew",
```
disizali, line 67: ditto

src/rules/no_misused_new.rs
```diff
@@ -0,0 +1,134 @@
...
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+  use crate::test_util::*;
+
+  #[test]
+  fn no_misused_new_test() {
```
disizali, line 112: can you separate valid and invalid tests into two function and add more tests for each of them, please ?
Like :
```ts
declare abstract class C {
  foo() {}
  get new();
  bar();
}
```

src/rules/mod.rs
ardyfeb: I forgot add that after fixing conflict my forked repo

src/rules/no_misused_new.rs
```diff
@@ -0,0 +1,212 @@
...
+      "#,
+      3,
+      4,
+    );
+
+    assert_lint_err_on_line::<NoMisusedNew>(
+      r#"
+class C {
+    new(): C;
+}
```
bartlomieju, line 175: Please add a test case with nested class:
```
class A {
  foo() {
    class C {
      new(): C;
    }
  }
}
```

It might not get caught at the moment

src/rules/no_misused_new.rs
```diff
@@ -0,0 +1,212 @@
...
+          // new
+          self.context.add_diagnostic(
+            method.span,
+            "no-misused-new",
+            "Class cannot have method named `new`.",
+          );
+        }
+      }
+    }
+  }
```
bartlomieju, line 131: There should probably be call to `swc_ecma_visit::visit_class_decl(self, expr, parent)` otherwise nested classes won't get inspected

src/rules/no_misused_new.rs
```diff
@@ -0,0 +1,212 @@
...
+            self.context.add_diagnostic(
+              signature.span,
+              "no-misused-new",
+              "Interfaces cannot be constructed, only classes",
+            );
+          }
+        }
+        _ => {}
+      }
+    }
```
bartlomieju, line 110: Ditto

src/rules/no_misused_new.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    false
+  }
+
+  fn is_constructor_keyword(&self, ident: &Ident) -> bool {
+    JsWord::from("constructor") == ident.sym
+  }
+}
```
bartlomieju, line 52: Cleanly done 👍

bartlomieju:
Thanks for the review @disizali

src/rules/no_misused_new.rs
ardyfeb: Thanks @bartlomieju

src/rules/no_misused_new.rs
ardyfeb: Got it

PR #138, opened 2020-06-09
add ban-ts-comment rule

disizali:
this PR is going to add :
+ `ban-ts-comment rule`
+ `25 tests about it`

_ps: `@ts-check` is allowed by default on tslint._

### DIFF
--- a/src/rules/ban_ts_comment.rs
+++ b/src/rules/ban_ts_comment.rs
@@ -14,7 +14,7 @@ impl BanTsComment {
 
     lazy_static! {
       static ref BTC_REGEX: regex::Regex =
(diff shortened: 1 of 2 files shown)


src/rules/ban_ts_comment.rs
```diff
@@ -0,0 +1,140 @@
...
+      r#"if (false) {
+// @ts-expect-error: Unreachable code error
+console.log('hello');
+}"#,
+      2,
+      0,
+    );
+    assert_lint_err_on_line::<BanTsComment>(
+      r#"if (false) {
+// @ts-nocheck: Unreachable code error
```
bartlomieju, line 133: Comments with description should be enabled by default - we use quite a few in Deno

disizali: it's ok now, just warn about `// @ts-[directive]`.

PR #167, opened 2020-06-14
Add no-await-in-loop rule

magurotuna:
This PR is to add [no-await-in-loop - Rules - ESLint](https://eslint.org/docs/rules/no-await-in-loop#top) rule.
Since this rule is not listed as `recommended`, I added it only to `get_all_rules`.

### DIFF
--- a/src/rules/no_await_in_loop.rs
+++ b/src/rules/no_await_in_loop.rs
@@ -1,0 +1,428 @@
+// Copyright 2020 the Deno authors. All rights reserved. MIT license.
+use super::Context;
+use super::LintRule;
+use swc_common::Span;
+use swc_ecma_ast::{
+  ArrowExpr, AwaitExpr, DoWhileStmt, ForInStmt, ForOfStmt, ForStmt, Function,
+  WhileStmt,
+};
+use swc_ecma_visit::{Node, Visit};
+
+pub struct NoAwaitInLoop;
+
+impl LintRule for NoAwaitInLoop {
+  fn new() -> Box<Self> {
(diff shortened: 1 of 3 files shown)


src/rules/no_await_in_loop.rs
```diff
@@ -0,0 +1,428 @@
...
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+  use crate::test_util::*;
+
+  #[test]
+  fn no_await_in_loop_valid() {
```
disizali, line 210: can you add this source to valid tests, please?
```ts
async function myLoop() {
  for (let i = await something(); i < n; i++) {}
}
```
This is considered an error,
That shouldn't be the case.

disizali, line 210: The same goes for this source :
```ts
async function myLoop() {
  for (const thing of await anotherThing) {
  // also :
  // for (let thing in await anotherThing) {}
    await something();
  }
}
```

magurotuna: Thank you for the review!
Oh, I missed that. I'll fix it.

magurotuna: Thanks again. I'll fix as well.

magurotuna:
I've added more valid testcases:

- `await` in the initialization of `for` statement
- `await` in the right-hand of `for-of` statement
- `await` in the right-hand of `for-in` statement
- `await` in the right-hand of `for-await-of` statement

and fixed the implementation to pass the tests.

PR #177, opened 2020-06-18
Add no-dupe-class-members rule

magurotuna:
~This is WORK IN PROGRESS.~
(UPDATE) Completed.

This PR is to add [no-dupe-class-members](https://eslint.org/docs/2.0.0/rules/no-dupe-class-members) rule.
Almost all tests are brought from [eslint/no-dupe-class-members.js at master · eslint/eslint](https://github.com/eslint/eslint/blob/master/tests/lib/rules/no-dupe-class-members.js), and I implemented the code to pass the tests.

### DIFF
--- a/src/rules/no_dupe_class_members.rs
+++ b/src/rules/no_dupe_class_members.rs
@@ -1,0 +1,702 @@
+// Copyright 2020 the Deno authors. All rights reserved. MIT license.
+use super::Context;
+use super::LintRule;
+use std::cmp::Ordering;
+use std::collections::BTreeMap;
+use swc_common::Span;
+use swc_ecma_ast::{
+  BigInt, Bool, Class, ClassMethod, ComputedPropName, Expr, Ident, Lit,
+  MethodKind, Null, Number, PropName, Str, Tpl,
+};
+use swc_ecma_visit::Node;
+use swc_ecma_visit::Visit;
+
+pub struct NoDupeClassMembers;
+
+impl LintRule for NoDupeClassMembers {
+  fn new() -> Box<Self> {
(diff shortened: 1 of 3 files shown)


src/rules/no_dupe_class_members.rs
```diff
@@ -0,0 +1,702 @@
...
+      "#,
+      vec![(3, 2), (4, 2)],
+    );
+
+    assert_lint_err_on_line_n::<NoDupeClassMembers>(
+      r#"
+class Foo {
+  set bar(value: number) {}
+  bar() {}
+}
```
bartlomieju, line 697: @magurotuna great tests, can you also add a test with nested class to make sure it's caught?

```ts
class Bar {
  bar() {
    class Foo {
      set bar(value: number) {}
      bar() {}
    }
  }
}
```

src/rules/no_dupe_class_members.rs
```diff
@@ -0,0 +1,702 @@
...
+      Expr::Tpl(Tpl {
+        ref quasis,
+        ref exprs,
+        ..
+      }) if exprs.is_empty() => {
+        quasis.iter().next().map(|q| q.raw.value.to_string())?
+      }
+      _ => return None,
+    },
+  };
```
bartlomieju, line 130: Nice :+1: I could be probably useful in other rules. If so we'll need to move it to other place, but looks good for now 👍

src/rules/no_dupe_class_members.rs
magurotuna: Thanks for pointing out! I added more tests to see nested class.

PR #180, opened 2020-06-19
add no-const-assign rule

disizali:
this PR is going to add:
- `no-const-assign rule`
- `23 tests about it`

### DIFF
--- a/src/rules/no_const_assign.rs
+++ b/src/rules/no_const_assign.rs
@@ -138,7 +138,7 @@ mod tests {
       for (const x in [1,2,3]) { foo(x); }
       for (const x of [1,2,3]) { foo(x); }
       const x = {key: 0}; x.key = 1;
-
+      if (true) {const a = 1} else { a = 2};
       // ignores non constant.
       var x = 0; x = 1;
       let x = 0; x = 1;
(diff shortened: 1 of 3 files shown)


src/rules/no_const_assign.rs
```diff
@@ -0,0 +1,178 @@
...
+    assert_lint_err::<NoConstAssign>("for (const i = 0; i < 10; ++i) {}", 26);
+    assert_lint_err::<NoConstAssign>(
+      "const x = 0; while (true) { x = x + 1; }",
+      28,
+    );
+    assert_lint_err_n::<NoConstAssign>(
+      "const x = 0  ; x = 1; x = 2;",
+      vec![15, 22],
+    );
+  }
```
bartlomieju, line 177: Great tests! However it seems that scope analysis is still not fully done so I'd suggest to add test cases in all types of scopes. If some of them don't catch this violation - it's fine let's leave them commented out and start uncommenting as we progress on scope analysis. WDYT @disizali?

disizali: I added new tests, they all worked as expected.
Since this rule respects all types of scopes and doesn't need to know the exact type of scope, the current state of the scope is sufficient for this one IMHO.

bartlomieju: Great to hear that!

Issue #160, opened 2020-06-12
Finish scope analysis


bartlomieju:

disizali:
@bartlomieju Do you plan to work on this soon?
I can work on some parts if you don't have enough time ...

bartlomieju:
@disizali go ahead, this week I'm working mostly on Deno.

A good start would be to:
- ensure all types of scopes are created when needed
- all bindings are added for each type of scope - warning: this is quite complex for `Pat` (func arguments)

After we got those two working we could think about discovering references to bindings.

bartlomieju:
@disizali I just checked and it seems that `Loop` scope is never constructed - ie. `for (...)`, `for .. of ...` and `for .. in ..`. So tests added in #180 might not catch everything. I'll try to handle `Loop` scope this weekend

disizali:
@bartlomieju Before you get started, I had an active branch about adding loop scoops.
I'll open a PR in a few minutes.

bartlomieju:
Done in #267

PR #192, opened 2020-06-22
scopes: arguments, objects and array inner functions 

### LINKED ISSUE #188: False positives in no-const-assign
Trying to run latest code on `std/` in `deno` repository I get:
```
(no-const-assign) Reassigning constant variable is not allowed
  actual = await buf.peek(1);
  ~~~~~~~~~~~~~~~~~~~~~~~~~~
    at /Users/biwanczuk/dev/deno/test_util/../std/io/bufio_test.ts:274:2

(no-const-assign) Reassigning constant variable is not allowed
  actual = await buf.peek(1);
  ~~~~~~~~~~~~~~~~~~~~~~~~~~
    at /Users/biwanczuk/dev/deno/test_util/../std/io/bufio_test.ts:278:2

(no-const-assign) Reassigning constant variable is not allowed
  actual = await buf.peek(1);
  ~~~~~~~~~~~~~~~~~~~~~~~~~~
    at /Users/biwanczuk/dev/deno/test_util/../std/io/bufio_test.ts:282:2

(no-const-assign) Reassigning constant variable is not allowed
  actual = await buf.peek(2);
  ~~~~~~~~~~~~~~~~~~~~~~~~~~
    at /Users/biwanczuk/dev/deno/test_util/../std/io/bufio_test.ts:286:2

(no-const-assign) Reassigning constant variable is not allowed
  actual = await buf.peek(4);
  ~~~~~~~~~~~~~~~~~~~~~~~~~~
    at /Users/biwanczuk/dev/deno/test_util/../std/io/bufio_test.ts:294:2

(no-const-assign) Reassigning constant variable is not allowed
  actual = await buf.peek(0);
  ~~~~~~~~~~~~~~~~~~~~~~~~~~
    at /Users/biwanczuk/dev/deno/test_util/../std/io/bufio_test.ts:301:2

(no-const-assign) Reassigning constant variable is not allowed
    eventsFired = [];
    ~~~~~~~~~~~~~~~~
    at /Users/biwanczuk/dev/deno/test_util/../std/node/events_test.ts:28:4
```

CC @disizali

disizali:
fixes #188.
On the way to completing #160.

object/array literal inner `Function` ,`Method`,`Getter`,`Setter` scopes as following :
```ts
let obj = {
  get getter(){} // Getter Scope
  set setter(){} // Setter Scope
  method(){} // Method Scope
}

let array = [
  function inArrayFunction(){} // Function scope
]
```

support `object prop call` and `new object` functions args as following :
```ts
Deno.test("arg test" , function(){}) // Function Scope
new Something(function(){}); // Function Scope
```


src/scopes.rs
```diff
@@ -35,12 +35,15 @@ pub enum ScopeKind {
...
   Function,
+  Method,
   Block,
   Loop,
   Class,
   Switch,
   With,
   Catch,
+  Getter,
+  Setter,
```
bartlomieju, line 46: Do we really need new scope types? Shouldn't `ScopeKind::Function` be enough?

src/scopes.rs
```diff
@@ -772,6 +930,198 @@ switch (foo) {
...
+      ()=>{const b;},
+      [
+        function nested() { const c;}
+      ],
+      {
+        innerMethod(){
+          const d;
+        }
+      }
+    ]
```
bartlomieju, line 1070: Nice 👍

src/scopes.rs
disizali: your right, **we** don't need them for the current rules, actually.
Done 👍 
stay minimal for now.


_I'm just curious about doing something in the future, Shouldn't we care about plugin development and analyzing all kinds of scopes? My opinion is that we should proceed according to the needs_

bartlomieju: > I'm just curious about doing something in the future, Shouldn't we care about plugin development and analyzing all kinds of scopes? 

I'm not sure - ESLint doesn't recognize `Getter/Setter` scopes so maybe it's critical

> My opinion is that we should proceed according to the needs

I agree

PR #195, opened 2020-06-22
scopes: import binding

disizali:
Ref #160 
This is a prerequisite for `no-import-assign` rule.

### DIFF
--- a/src/scopes.rs
+++ b/src/scopes.rs
@@ -22,6 +22,7 @@ pub enum BindingKind {
   Param,
   Class,
   CatchClause,
+  Import,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


src/scopes.rs
```diff
@@ -1106,6 +1124,67 @@ switch (foo) {
...
+    let source_code = r#"
+    import defaultExport1 from "module-name";
+    import * as namespaced1 from "module-name";
+    import { export1 } from "module-name";
+    import { export2 as alias1 } from "module-name";
+    import { export3 , export4 } from "module-name";
+    import { export5 , export6 as alias2} from "module-name";
+    import defaultExport2, { export7 } from "module-name";
+    import defaultExport3, * as namespaced2 from "module-name";
+    import "module-name";
```
bartlomieju, line 1141: I think that this rule should also catch violations for bindings in `export { a, b, c } from "./asdf.ts"` - if I'm not mistaken there's no separate rule for exports

disizali: in export-from  declerations are not new bindings in the working script.
```ts
// export
export { a } from './a.ts';
a = 2; // 2:1 error 'a' is not defined. (no-undef)

// import
import { a } from './a.ts';
a = 2; // 2:1 error 'a' is read-only. (no-import-assign)
```

did I miss something? or a more specific example, please?

bartlomieju: Oooops, you're absolutely right @disizali, my bad

PR #203, opened 2020-06-28
Add no-shadow-restricted-names rule

ardyfeb:


src/rules/no_shadow_restricted_names.rs
```diff
@@ -0,0 +1,286 @@
...
+              self.check_shadowing(&assign.key);
+            }
+            ObjectPatProp::Rest(rest) => self.check_pat(&rest.arg),
+            ObjectPatProp::KeyValue(key_value) => {
+              self.check_pat(&key_value.value);
+            }
+          }
+        }
+      }
+      _ => {}
```
disizali, line 99: it would be great if we could recognize the rest patterns like ```const [...undefined] = []```.
can you add `Rest` arm, please?

ardyfeb: Done, thanks @disizali. i also add more tests

src/rules/no_shadow_restricted_names.rs
```diff
@@ -0,0 +1,307 @@
...
+    match ident.sym.to_string().as_str() {
+      "undefined" | "NaN" | "Infinity" | "arguments" | "eval" => true,
+      _ => false,
+    }
+  }
+
+  fn check_pat(&self, pat: &Pat, check_scope: bool) {
+    match pat {
+      Pat::Ident(ident) => {
+        if &ident.sym.to_string() == "undefined" && check_scope {
```
bartlomieju, line 57: Please add a comment explaining comparison with "undefined"

PR #208, opened 2020-07-02
add ban-types rule

disizali:
this PR is going to add
- `ban-types rule`
- `24 tests about it`

### DIFF
--- a/src/rules/ban_types.rs
+++ b/src/rules/ban_types.rs
@@ -74,10 +74,10 @@ impl Visit for BanTypesVisitor {
       return;
     }
     self.context.add_diagnostic(
-        ts_type_lit.span,
-        "ban-types",
-        BANNED_TYPES[5].1, // `Object` message
-      );
+      ts_type_lit.span,
+      "ban-types",
+      BANNED_TYPES[5].1, // `Object` message
+    );
   }
   fn visit_ts_keyword_type(
     &mut self,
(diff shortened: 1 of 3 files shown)


src/rules/ban_types.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+const BANNED_TYPES: [(&str, &str); 6] = [
+  ("String", "Use `string` instead"),
+  ("Boolean", "Use `boolean` instead"),
+  ("Number", "Use `number` instead"),
+  ("Symbol", "Use `symbol` instead"),
+  ("Function", "Define the function shape Explicitly."),
+  ("Object", "if you want a type meaning `any object` use `Record<string, unknown>` instead,
+or If you want a type meaning `any value`, you probably want `unknown` instead."),
```
bartlomieju, line 42: nitpick: `If` shouldn't be capitalized

src/rules/ban_types.rs
```diff
@@ -0,0 +1,169 @@
...
+      for element in ts_type_lit.members.iter() {
+        self.visit_ts_type_element(element, ts_type_lit);
+      }
+      return;
+    }
+    self.context.add_diagnostic(
+        ts_type_lit.span,
+        "ban-types",
+        "if you want a type meaning `any object` use `Record<string, unknown>` instead,
+or If you want a type meaning `any value`, you probably want `unknown` instead.",
```
bartlomieju, line 80: Ditto, also, already defined in `BANNED_TYPES` array - use that instead?

disizali: of course, thanks

Issue #86, opened 2020-06-02
typescript-eslint recommended rules


bartlomieju:
Tracking issues for recommended rules of typescript-eslint

https://github.com/typescript-eslint/typescript-eslint/blob/v3.1.0/packages/eslint-plugin/src/configs/recommended.ts

- [x] adjacent-overload-signatures <i>(easy)</i>
- [x] ban-ts-comment
- [x] ban-types
- [x] explicit-module-boundary-types <i>(easy)</i>
- [x] no-array-constructor
- [x] no-empty-function (needs to be fixed, false positives for constructors)
- [x] no-empty-interface
- [x] no-explicit-any
- [x] no-extra-non-null-assertion
- [x] no-extra-semi <i>(easy)</i>
- [x] no-inferrable-types <i>(easy)</i>
- [x] no-misused-new
- [x] no-namespace
- [x] no-non-null-asserted-optional-chain
- [x] no-non-null-assertion
- [x] no-this-alias
- [x] no-unused-vars <i>(hard, needs scope analysis)</i>
- [x] <s>no-var-requires</s>
- [x] prefer-as-const
- [x] prefer-namespace-keyword
- [x] triple-slash-reference

lucacasonato:
I don't think `no-var-requires` makes sense for us: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-var-requires.md

bartlomieju:
I've marked this issue with `0.2.0` milestone. I want to implement remaining rules this week so they will be included in `deno v1.1`

disizali:
@bartlomieju
i'm almost done with `no-this-alias` and `ban-ts-comment`,
Also `no-inferrable-types` is my next goal.
you can start the rest first.

iykekings:
I'm working on `no-non-null-asserted-optional-chain` and `no-non-null-assertion`.

bartlomieju:
@iykekings are you still working on `no-non-null-asserted-optional-chain`?

iykekings:
Yes, had plenty on my plate at work. I will try to finish it ASAP

magurotuna:
I'm working on `adjacent-overload-signatures`.

humancalico:
@bartlomieju It seems `no-inferrable-types` is done. It's not updated in your comment

qu4k:
Taking `explicit-module-boundary-types`

bartlomieju:
Done with #267

PR #161, opened 2020-06-12
Feat: add no-non-null-assertion rule

iykekings:

PR #241, opened 2020-07-25
add no_non_null_asserted_optional_chain rule

divy-work:
Adds https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-non-null-asserted-optional-chain.md

Ref: #86


divy-work:
Oh completely forgot to see #161 ! My bad :sweat_smile: 

Keeping this opened just in case.

iykekings:
😢 This wouldn't pass the tests in eslint for this rules. I would've let you override #161 if it did.

divy-work:
Yes @iykekings I was about to implment them but it seems your PR works better. I'm gonna close this one. :smile:

PR #243, opened 2020-07-26
Add no-unexpected-multiline rule

qkniep:
This PR adds the [`no-unexpected-multiline`](https://eslint.org/docs/rules/no-unexpected-multiline) rule.
Updating swc is required to handle optional chaining (in ES2020) correctly, as is also marked through TODO comments.

### DIFF
--- a/src/rules/no_unexpected_multiline.rs
+++ b/src/rules/no_unexpected_multiline.rs
@@ -1,0 +1,322 @@
+// Copyright 2020 the Deno authors. All rights reserved. MIT license.
+use super::Context;
+use super::LintRule;
+use crate::swc_common::{BytePos, Span, Spanned};
+use crate::swc_ecma_ast;
+use regex::Regex;
+use swc_ecma_visit::Node;
+use swc_ecma_visit::Visit;
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+pub struct NoUnexpectedMultiline;
+
+impl LintRule for NoUnexpectedMultiline {
+  fn new() -> Box<Self> {
+    Box::new(NoUnexpectedMultiline)
+  }
+
+  fn code(&self) -> &'static str {
+    "no-unexpected-multiline"
+  }
+
+  fn lint_module(&self, context: Arc<Context>, module: &swc_ecma_ast::Module) {
+    let mut visitor = NoUnexpectedMultilineVisitor::new(context);
+    visitor.visit_module(module, module);
+  }
+}
+
+struct NoUnexpectedMultilineVisitor {
+  context: Arc<Context>,
+}
+
+impl NoUnexpectedMultilineVisitor {
+  pub fn new(context: Arc<Context>) -> Self {
+    Self { context }
+  }
+
+  fn check_for_break_after(&self, outer: Span, after: Span, msg: &str) {
+    lazy_static! {
+      static ref MSG_MAP: HashMap<&'static str, &'static str> = {
+        let mut map = HashMap::new();
+        map.insert(
+          "call",
+          "Unexpected newline between function and ( of function call",
+        );
+        map.insert(
+          "member",
+          "Unexpected newline between object and [ of property access",
+        );
+        map.insert(
+          "div",
+          "Unexpected newline between numerator and division operator",
+        );
+        map
+      };
+    }
+    let source_map = &self.context.source_map;
+    let before_paren = outer.trim_start(after).unwrap();
+    let temp_span = source_map
+      .span_take_while(before_paren, |c| *c == ')' || c.is_whitespace());
+    let left_paren = before_paren.trim_start(temp_span).unwrap();
+    let line1 = source_map.lookup_char_pos(after.hi()).line;
+    let line2 = source_map.lookup_char_pos(left_paren.lo()).line;
+    if line1 != line2 {
+      self.context.add_diagnostic(
+        left_paren.with_lo(left_paren.lo() + BytePos(1)),
+        "no-unexpected-multiline",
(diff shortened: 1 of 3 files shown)


src/rules/no_unexpected_multiline.rs
```diff
@@ -0,0 +1,322 @@
...
+          "member",
+          "Unexpected newline between object and [ of property access",
+        );
+        map.insert(
+          "div",
+          "Unexpected newline between numerator and division operator",
+        );
+        map
+      };
+    }
```
bartlomieju, line 57: Nitpick: `lazy_static` should be placed after the imports

src/rules/no_unexpected_multiline.rs
```diff
@@ -0,0 +1,322 @@
...
+impl Visit for NoUnexpectedMultilineVisitor {
+  fn visit_bin_expr(
+    &mut self,
+    bin_expr: &swc_ecma_ast::BinExpr,
+    _parent: &dyn Node,
+  ) {
+    lazy_static! {
+      static ref SLASH_AND_FLAGS: regex::Regex =
+        Regex::new(r"^/[gimsuy]+(?:[\W].*)?$").unwrap();
+    }
```
bartlomieju, line 84: Ditto

src/rules/no_unexpected_multiline.rs
```diff
@@ -0,0 +1,322 @@
...
+    call_expr: &swc_ecma_ast::CallExpr,
+    _parent: &dyn Node,
+  ) {
+    if let swc_ecma_ast::ExprOrSuper::Expr(expr) = &call_expr.callee {
+      self.visit_expr(expr, _parent);
+    }
+    for arg in &call_expr.args {
+      self.visit_expr(&arg.expr, _parent);
+    }
+    // TODO: handle ES2020 optional chaining
```
bartlomieju, line 124: What's exactly missing here? SWC has `Expr::OptChain` which gives [`OptChainExpr`](https://docs.rs/swc_ecma_ast/0.25.0/swc_ecma_ast/struct.OptChainExpr.html), I think that could be used?

src/rules/no_unexpected_multiline.rs
```diff
@@ -0,0 +1,322 @@
...
+    );
+    assert_lint_err_on_line::<NoUnexpectedMultiline>(
+      r#"const x = aaaa<
+  test
+>/*
+test
+*/`foo`"#,
+      5,
+      3,
+    );
```
bartlomieju, line 320: Nice tests, are they ported from ESLint?

qkniep: In part, yes. That last one for example is from ESLint. Probably would not have come up with this by myself.

src/rules/no_unexpected_multiline.rs
qkniep: Specifically I would need the ability to differentiate between what's before and after the `?.` operator. For example, according to ESLint in the following the first is lint free and the second has an unexpected-multiline:
```
foo.bar?.
(a);

foo.bar
(a)?.b;
```
In SWC's `OptChainExpr` in both cases the value of `expr` would be the whole expression.

bartlomieju: I'm still not entirely grasping the problem... 😒  but it seems the problem lies in SWC after all. @qkniep could you provide example what would work in this case?

CC @kdy1 for opinion

kdy1: I think OptChainExpr should have a field like `question_dot_token: Span`.
Will adding it fix the issue?

bartlomieju: Adding `question_dot_token: Span would allow to check the line this token is on, so it should be possible to compare lines between statements in the above example. It could work 👍

bartlomieju:
@qkniep I did a mistake upgrading `dprint-typescript-plugin` so unfortunately you'll have to wait a day or two for `swc_ecma_ast 0.26.0` which includes span containing `?.` token. With it, it should be possible to fix the problem you described above.

kdy1:
> @qkniep I did a mistake upgrading `dprint-typescript-plugin` so unfortunately you'll have to wait a day or two for `swc_ecma_ast 0.26.0` which includes span containing `?.` token. With it, it should be possible to fix the problem you described above.

I've published 0.26.0

https://crates.io/crates/swc_ecma_ast

bartlomieju:
> > @qkniep I did a mistake upgrading `dprint-typescript-plugin` so unfortunately you'll have to wait a day or two for `swc_ecma_ast 0.26.0` which includes span containing `?.` token. With it, it should be possible to fix the problem you described above.
> 
> I've published 0.26.0
> 
> https://crates.io/crates/swc_ecma_ast

Thanks @kdy1 unfortunately I have to wait for another release of `dprint-typescript-plugin` because we reexport a few `swc_` crates from there (to ensure we got only a single version of each crate in `deno` binary). Hopefully it soon won't be required as it's a lot of gymnastics for each upgrade 😆

bartlomieju:
@qkniep I did upgrade to latest `swc_ecma_ast`. You should be able to check if `?.` token is on the same line using [`OptChainExpr.question_dot_token`](https://docs.rs/swc_ecma_ast/0.26.0/swc_ecma_ast/struct.OptChainExpr.html)

qkniep:
I figured out how to handle optional chaining with this rule, in the end the best solution I found is not using the `question_dot_token` span. I am really sorry for the additional work my misunderstanding caused you, @kdy1.

kdy1:
@qkniep It's ok. The solution seems quite clever.

PR #252, opened 2020-08-03
Add no-invalid-regexp rule.

qkniep:
This PR adds the [`no-invalid-regexp`](https://eslint.org/docs/rules/no-invalid-regexp) rule.
For this I added the [`js-regex`](https://github.com/qkniep/js-regex) crate, which I wrote for this purpose. It is basically a partial Rust port of [`regexpp`](https://github.com/mysticatea/regexpp) (only the validator part).


qkniep:
Sounds good. Will move the library to a module in `deno_lint` instead.

bartlomieju:
> Sounds good. Will move the library to a module in `deno_lint` instead.

Great, thank you!

PR #262, opened 2020-08-16
Need some help adding getting the correct span for the lint error

AkinAguda:
After the regex matches some irregular white space, I need to get the span where the white space was found and pass it into `add_diagnostic` @bartlomieju


src/rules/mod.rs
```diff
@@ -187,6 +189,7 @@ pub fn get_all_rules() -> Vec<Box<dyn LintRule>> {
     no_extra_non_null_assertion::NoExtraNonNullAssertion::new(),
     no_extra_semi::NoExtraSemi::new(),
     no_func_assign::NoFuncAssign::new(),
+    no_irregular_whitescpace::NoIrregularWhitespace::new(),
```
lucacasonato, line 192: `no_irregular_whitescpace` -> `no_irregular_whitespace`

AkinAguda: Oh thanks. Didn't notice that

src/rules/no_irregular_whitespace.rs
```diff
@@ -0,0 +1,62 @@
...
+use swc_ecmascript::ast::Module;
+
+pub struct NoIrregularWhitespace;
+
+fn test_for_whitespace(value: &Cow<str>) -> bool {
+  lazy_static! {
+    static ref ALL_IRREGULARS: Regex = Regex::new(r"[\f\v\u0085\ufeff\u00a0\u1680\u180e\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200a\u200b\u202f\u205f\u3000\u2028\u2029]").unwrap();
+    static ref LINE_BREAK_MATCHER: Regex = Regex::new(r"[^\r\n]+").unwrap();
+  }
+  ALL_IRREGULARS.is_match(value)
```
bartlomieju, line 14: This is good to test if there is some irregular whitespace 👍  after this check you can use [`Regex::find_iter`](https://docs.rs/regex/1.3.9/regex/struct.Regex.html#method.find_iter) to get `Matches` with all occurences and then use [`Match::range`](https://docs.rs/regex/1.3.9/regex/struct.Match.html#method.range) method to get indexes. 

After that you can create new span by adding start position of Regex's span. Check out [`Span::new`](https://docs.rs/swc_common/0.9.1/swc_common/struct.Span.html#method.new) method.

src/rules/no_irregular_whitespace.rs
```diff
@@ -0,0 +1,62 @@
...
+    for line_info in lines.into_iter() {
+      let source_file_and_index = context
+        .source_map
+        .lookup_line(BytePos(line_info.line_index as u32))
+        .unwrap();
+      let source_code = source_file_and_index
+        .sf
+        .get_line(line_info.line_index)
+        .unwrap();
+      if test_for_whitespace(&source_code) {
```
bartlomieju: Good start, but you might want to use `Visitor` pattern from other rules and check only [`Regex`](https://docs.rs/swc_ecma_ast/0.28.0/swc_ecma_ast/struct.Regex.html) nodes.

bartlomieju: That way you'll be able to use `Regex.span` to get position in the file and combine it with ranges from Rust's `Match`es structs

AkinAguda: Oh I see. Thanks a lot

src/rules/no_irregular_whitespace.rs
AkinAguda: Oh cool. I get it. Thanks

src/rules/no_irregular_whitespace.rs
```diff
@@ -0,0 +1,73 @@
...
+use swc_common::{hygiene::SyntaxContext, BytePos, Span};
+use swc_ecmascript::ast::Module;
+
+pub struct NoIrregularWhitespace;
+
+fn test_for_whitespace(value: &str) -> Option<Matches> {
+  lazy_static! {
+    static ref ALL_IRREGULARS: Regex = Regex::new(r"[\f\v\u0085\ufeff\u00a0\u1680\u180e\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200a\u200b\u202f\u205f\u3000\u2028\u2029]").unwrap();
+    static ref LINE_BREAK_MATCHER: Regex = Regex::new(r"[^\r\n]+").unwrap();
+  }
```
bartlomieju, line 13: Nitpick: move `lazy_static!` outside this function

src/rules/no_irregular_whitespace.rs
```diff
@@ -0,0 +1,73 @@
...
+  use super::*;
+  use crate::test_util::*;
+
+  #[test]
+  fn no_irregular_whitespace_valid() {
+    assert_lint_err_on_line::<NoIrregularWhitespace>(
+      "console.log('The last space in this string will make it　fail');",
+      1,
+      55,
+    );
```
bartlomieju, line 71: Please port remaining tests from ESLint

PR #263, opened 2020-08-18
feat: wasm-bindgen and stdweb compatbility

divy-work:
Uses wasm compatible `instant` instead of `std::instant`. Fall backs to `std::instant` when feature not specified.

Use cases for wasm compatibility - playground :smile:


caspervonb:
Worth noting that std::time::Instant is supported with wasm32-wasi.

divy-work:
Right, but not in `wasm32-unknown`

lucacasonato:
> Worth noting that std::time::Instant is supported with wasm32-wasi

Does deno_lint compile with `wasm32-wasi`? If so I think this patch is not necessary. You can polyfill the `wasi_time_get ` syscall on the web using `Date` (or `performance.now()`).

divy-work:
Got it.

eliassjogreen:
This would be nice for wasm32-unknown-unknown support for a project the denosaurs are working on, I don't really see a reason why not to support it? wasm-pack (nor wasm-bindgen I think) doesn't seem to support wasi which makes the build process way more complicated and just a mess. If wasm32 support is not planned could you provide an example of how I would polyfill it?

bartlomieju:
@eliassjogreen I'm not sure, CC @caspervonb

caspervonb:
> I don't really see a reason why not to support it

I do, we should not be supporting ABIs that don't exist. Some of the cargo ecosystem have abused the target with feature flags but it is still a bare metal target with no system calls.

> wasm-pack (nor wasm-bindgen I think) doesn't seem to support wasi which makes the build process way more complicated and just a mess. 

They do, just compile with cargo build --target wasm32-wasi then run wasm-bindgen after without wasm-pack (wasm-pack is not a great tool IMO as it hides all the compilation steps from you).

caspervonb:
Kom inn på Deno Discord om du sitter fast og trenger hjelp @eliassjogreen 🙃

Issue #162, opened 2020-06-12
Feature request: Add suggestion(s) to error messages


iykekings:
Error messages should also have suggestions for possible fixes of the error.

bartlomieju:
This would be a very useful feature to have. 

It can be implemented by adding a new field to `LintDiagnostic` called `suggestion` or `hint`, which could optionally be set when creating diagnostics in rules.

```rust
pub struct LintDiagnostic {
  pub range: Range,
  pub filename: String,
  pub message: String,
  pub code: String,
  pub suggestion: Option<String>,
}
```

`Option<String>` should be a valid choice because not all rules must provide suggestions (eg. in trivial rules it might not make sense to add suggestions.

iykekings:
@bartlomieju could there be a scenario where there are multiple solutions to one error?

bartlomieju:
@iykekings sorry I missed you comment. Yeah, there might be such scenario - in that case we can change `LintDiagnostic::hint` to be `Vec<String>`.

jcc10:
> (eg. in trivial rules it might not make sense to add suggestions.)

I would just like to add that trivial depends on who you are talking to. Some cases may be trivial but the conventional way is the wrong way.

The best example is ```const self = this```. This is a "hack" for when you are running a function in a different scope and the linter hates it. But it has been the convention for the past forever. The better(?) way is to use a arrow function, but even I, who found this out over a year ago, will still look up if I am remembering that fact correctly.

bartlomieju:
I'm going to close this issue now, most of the rules now include a `hint` suggesting what to change to make linter happy.

PR #267, opened 2020-08-19
Add no-unused-vars rule 

kdy1:
I may split swc_emca_transforms with features (or multiple crates) to reduce compile time...

### DIFF
--- a/src/rules/no_unused_vars.rs
+++ b/src/rules/no_unused_vars.rs
@@ -5,7 +5,7 @@ use swc_ecmascript::ast::{
   ArrowExpr, CatchClause, ClassMethod, ExportDecl, ExportNamedSpecifier, Expr,
   FnDecl, FnExpr, Ident, ImportDefaultSpecifier, ImportNamedSpecifier,
   ImportStarAsSpecifier, MemberExpr, NamedExport, Param, Pat, SetterProp,
-  VarDeclOrPat, VarDeclarator,
+  TsEntityName, TsTypeRef, VarDeclOrPat, VarDeclarator,
 };
 use swc_ecmascript::utils::find_ids;
 use swc_ecmascript::utils::ident::IdentLike;
@@ -36,10 +36,15 @@ impl LintRule for NoUnusedVars {
     let mut collector = Collector {
       used_vars: Default::default(),
       cur_defining: Default::default(),
+      used_types: Default::default(),
     };
     module.visit_with(module, &mut collector);
 
-    let mut visitor = NoUnusedVarVisitor::new(context, collector.used_vars);
+    let mut visitor = NoUnusedVarVisitor::new(
+      context,
+      collector.used_vars,
+      collector.used_types,
+    );
     module.visit_with(module, &mut visitor);
   }
 }
@@ -47,6 +52,7 @@ impl LintRule for NoUnusedVars {
 /// Collects information about variable usages.
 struct Collector {
   used_vars: HashSet<Id>,
+  used_types: HashSet<Id>,
   /// Currently defining functions or variables.
   ///
   ///
@@ -61,6 +67,18 @@ struct Collector {
 }
 
 impl Visit for Collector {
+  fn visit_ts_type_ref(&mut self, ty: &TsTypeRef, _: &dyn Node) {
+    fn get_id(r: &TsEntityName) -> Id {
+      match r {
+        TsEntityName::TsQualifiedName(q) => get_id(&q.left),
+        TsEntityName::Ident(i) => i.to_id(),
+      }
+    }
+
+    let id = get_id(&ty.type_name);
+    self.used_types.insert(id);
+  }
+
   fn visit_expr(&mut self, expr: &Expr, _: &dyn Node) {
     match expr {
       Expr::Ident(i) => {
@@ -157,11 +175,20 @@ impl Visit for Collector {
 struct NoUnusedVarVisitor {
   context: Arc<Context>,
   used_vars: HashSet<Id>,
+  used_types: HashSet<Id>,
 }
 
 impl NoUnusedVarVisitor {
-  fn new(context: Arc<Context>, used_vars: HashSet<Id>) -> Self {
-    Self { context, used_vars }
+  fn new(
+    context: Arc<Context>,
+    used_vars: HashSet<Id>,
+    used_types: HashSet<Id>,
+  ) -> Self {
+    Self {
+      context,
+      used_vars,
+      used_types,
+    }
   }
 }
 
--- a/src/rules/no_unused_vars.rs
+++ b/src/rules/no_unused_vars.rs
@@ -616,11 +616,11 @@ mod tests {
     assert_lint_err::<NoUnusedVars>("import x from \"y\";", 7);
     assert_lint_err::<NoUnusedVars>(
       "export function fn2({ x, y }) {\n console.log(x); \n};",
-      0,
+      25,
     );
     assert_lint_err::<NoUnusedVars>(
       "export function fn2( x, y ) {\n console.log(x); \n};",
-      0,
+      24,
     );
   }
 
--- a/src/rules/no_unused_vars.rs
+++ b/src/rules/no_unused_vars.rs
@@ -8,8 +8,8 @@ use swc_ecmascript::visit::Node;
 use swc_ecmascript::visit::Visit;
 use swc_ecmascript::{
   ast::{
-    ArrowExpr, CatchClause, ClassDecl, ClassMethod, Constructor, Decl,
-    ExportDecl, ExportNamedSpecifier, Expr, FnDecl, FnExpr, Ident,
+    ArrowExpr, CatchClause, ClassDecl, ClassMethod, ClassProp, Constructor,
+    Decl, ExportDecl, ExportNamedSpecifier, Expr, FnDecl, FnExpr, Ident,
     ImportDefaultSpecifier, ImportNamedSpecifier, ImportStarAsSpecifier,
     KeyValueProp, MemberExpr, MethodKind, Module, NamedExport, Param, Pat,
     Prop, SetterProp, TsEntityName, TsExprWithTypeArgs, TsModuleDecl,
@@ -39,6 +39,8 @@ impl LintRule for NoUnusedVars {
       used_types: Default::default(),
     };
     module.visit_with(module, &mut collector);
+
+    dbg!(&collector.used_vars, &collector.used_types);
 
     let mut visitor = NoUnusedVarVisitor::new(
       context,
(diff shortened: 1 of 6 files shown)


src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,136 @@
...
+  use crate::test_util::*;
+
+  #[test]
+  fn no_unused_vars_ok() {
+    assert_lint_ok::<NoUnusedVars>("var a = 1; console.log(a)");
+    assert_lint_ok::<NoUnusedVars>(
+      "var a = 1; function foo() { console.log(a) } ",
+    );
+    assert_lint_ok::<NoUnusedVars>(
+      "var a = 1; const arrow = () => a; console.log(arrow)",
```
kdy1, line 119: I added `console.log(arrow)` to prevent `arrow` being unused.

src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    // Hoisting. This code is wrong, but it's not related with unused-vars
+    assert_lint_ok::<NoUnusedVars>("console.log(a); var a = 1;");
+  }
+
+  #[test]
+  fn no_unused_vars_err() {
+    assert_lint_err::<NoUnusedVars>("var a = 0", 4);
+
+    // variable shadowing
```
kdy1, line 130: Is description like this supposed to be written as comments?

src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,136 @@
...
+
+impl NoUnusedVarVisitor {
+  fn new(context: Arc<Context>, used_vars: HashSet<Id>) -> Self {
+    Self { context, used_vars }
+  }
+}
+
+/// As we only care about variables, only variable declrations are checked.
+impl Visit for NoUnusedVarVisitor {
+  // TODO(kdy1): swc_ecmascript::visit::noop_visit_type!() after updating swc
```
kdy1, line 89: This trick can be applied to other visitors, too.

src/rules/no_unused_vars.rs
bartlomieju: Sure, sounds good 👍

src/rules/no_unused_vars.rs
bartlomieju: Sounds good 👍  `deno_lint` already uses `swc_ecmascript 0.3.2` so it can be used already?

src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,142 @@
...
+  used_vars: HashSet<Id>,
+}
+
+impl NoUnusedVarVisitor {
+  fn new(context: Arc<Context>, used_vars: HashSet<Id>) -> Self {
+    Self { context, used_vars }
+  }
+}
+
+/// As we only care about variables, only variable declrations are checked.
```
bartlomieju, line 93: Actually ESLint also check function params and any other declared bindings

bartlomieju (commented):
@kdy1 looks great, as a rule of thumb we try to port all tests from ESLint with default or strictest settings to ensure feature parity. I think at the moment unused function parameters would be ignored. WDYT?

src/rules/no_unused_vars.rs
kdy1: Oh.. I didn't know that. I'll add support for it.

kdy1:
I think it's the right way. I'll add support for parameters and copy tests from eslint

src/rules/no_unused_vars.rs
kdy1: I'll do this.
I think doing so for all passes, which is a good fit for separate pr, will reduces binary size a lot. Rustc cannot deduce that a visitor does not interact with type.

kdy1:
Side note: `noop_fold_type` moved from `swc_ecmascript::transforms` to `swc_ecmascript::visit` in `swc_ecmascript@0.4.0`  to reduce the binary size related to helpers in `swc_ecmascript::utils`

src/rules/no_unused_vars.rs
bartlomieju: Sounds good 👍  `deno_lint` is actually only 650kB when compiled into `deno` but any size reduction is more than welcome

bartlomieju:
`no-unused-vars` has many configs in ESLint but I think we should go with:
- `vars: all` - all variables must be used, including ones from global scope
- `varsIgnorePattern: ^_` - just like in Rust if variable starts with underscore it can be unused
- `args: all` - all params must be used, unless they start with `^_`
- `ignoreRestSibling: false`
- `caughtErrors: all` - catch binding variable also must start with `^_` if not used

kdy1:
@bartlomieju Should I follow error message of eslint?
unused-vars pass in eslint has multiple error kinds.

bartlomieju:
> @bartlomieju Should I follow error message of eslint?
> unused-vars pass in eslint has multiple error kinds.

@kdy1 a single error message is fine for now, we'll likely address it as part of #162

kdy1:
I'm almost done. I ported all tests from eslint which satisfies the default config above.


Currently some tests related to control flow ignored. e.g. assignment expression which assigns to itself

kdy1:
Oh. I forgot typescript related testing :)
How can I parse typescript file while linting?

bartlomieju:
> Oh. I forgot typescript related testing :)
> How can I parse typescript file while linting?

All tests are parsed as TypeScript files so it should just work ®

kdy1:
@bartlomieju I think typescript checker requires updating swc and implementing a new pass inside swc.

How should I do?
May I split commit into two pull requests? (one for javascript, one for typescript)
If not, when do you plan to update the swc?

bartlomieju:
@kdy1 whichever works best for you, splitting into two PRs, sounds good to me. I can upgrade swc today.

kdy1:
@bartlomieju Thanks! I will split a pr into two. And no hurry, I have to modify resolver (to guarantee uniqueness of type names) and it will take some time.

kdy1:
I forgot undrafting this.
r? @bartlomieju

kdy1:
# Current status
 - Depends on https://github.com/swc-project/swc/pull/994

 - All ecmascript tests are handled
 - All typescript valid cases are handled.
 - TypeSciprt invalid cases are not copied yet.

src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,2052 @@
...
+    // const text = 'text';
+    // export function Foo() {
+    //   return (
+    //     <div>
+    //       <input type=\"search\" size={30} placeholder={text} />
+    //     </div>
+    //   );
+    // }
+    //       ",
+    //     );
```
bartlomieju, line 1780: @kdy1 do you want to uncomment his before landing?

src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,2052 @@
...
+export const a: A<SomeOther> = {
+  foo: 'bar',
+};
+
+      ",
+    );
+  }
+
+  #[test]
+  #[ignore = "typescript property analysis is not implemented yet"]
```
bartlomieju, line 1535: Is this ignore still relevant?

src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,2052 @@
...
+    assert_lint_err::<NoUnusedVars>(
+      "var a = function(){ return function() { a(); } };",
+      4,
+    );
+    assert_lint_err::<NoUnusedVars>("const a = () => { a(); };", 6);
+    assert_lint_err::<NoUnusedVars>("const a = () => () => { a(); };", 6);
+  }
+
+  #[test]
+  #[ignore = "pure method analysis is not implemented yet"]
```
bartlomieju, line 935: ditto

src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,2052 @@
...
+    assert_lint_err::<NoUnusedVars>("function foo(a) { a += a + 1 } foo();", 0);
+    assert_lint_err::<NoUnusedVars>("function foo(a) { a++ } foo();", 0);
+    assert_lint_err::<NoUnusedVars>("var a = 3; a = a * 5 + 6;", 0);
+    assert_lint_err::<NoUnusedVars>("var a = 2, b = 4; a = a * 2 + b;", 0);
+
+    assert_lint_err::<NoUnusedVars>("const a = 1; a += 1;", 6);
+  }
+
+  #[test]
+  #[ignore = "control flow analysis is not implemented yet"]
```
bartlomieju, line 878: ditto

src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,2052 @@
...
+    assert_lint_err::<NoUnusedVars>("export default (a) => {};", 16);
+    assert_lint_err::<NoUnusedVars>(
+      "export default (a, b) => { console.log(a); };",
+      19,
+    );
+    assert_lint_err::<NoUnusedVars>("try{}catch(err){};", 11);
+  }
+
+  #[test]
+  #[ignore = "control flow analysis is not implemented yet"]
```
bartlomieju, line 862: ditto

src/rules/no_unused_vars.rs
```diff
@@ -0,0 +1,2052 @@
...
+    assert_lint_err::<NoUnusedVars>("var _a; var b;", 12);
+    assert_lint_err::<NoUnusedVars>("function foo(a, _b) { } foo()", 13);
+    assert_lint_err::<NoUnusedVars>(
+      "function foo(a, _b, c) { return a; } foo();",
+      20,
+    );
+  }
+
+  #[test]
+  #[ignore = "control flow analysis is not implemented yet"]
```
bartlomieju, line 804: ditto

src/rules/no_unused_vars.rs
kdy1: I'm not sure because It would require lots of work. (Currently, jsx nodes are not handled by the resolver.)

src/rules/no_unused_vars.rs
kdy1: Yes, it's not implemented yet. It requires more delicate scope analysis I think

src/rules/no_unused_vars.rs
bartlomieju: Sounds good

src/rules/no_unused_vars.rs
kdy1: It can be done in future when some more scope / control flow analysis is implemented, but not now.

Also, I think it's ok to exclude this from the initial version.

bartlomieju (approved):
LGTM, thank you @kdy1 this is massive!

ry:
Nice work @kdy1 !

PR #276, opened 2020-08-24
chore: remove unnecessary allocation

magurotuna:
This is a so trivial patch; unnecessary string allocation has been removed.

With regard to replacing `contains()` with `iter().any()`, I referred to [the std library document](https://doc.rust-lang.org/std/primitive.slice.html#method.contains) .

> If you do not have an `&T`, but just an `&U` such that `T: Borrow<U>` (e.g. `String: Borrow<str>`), you can use `iter().any`


bartlomieju (approved):
Sure, LGTM, thanks @magurotuna

PR #278, opened 2020-08-24
[WIP] (rule) no-misleading-character-class

humancalico:


humancalico:
**Update**
This PR is currently stalled as the [parser](https://github.com/mysticatea/regexpp/blob/master/src/parser.ts) from https://github.com/mysticatea/regexpp needs to be ported first. If anyone wants to take a stab at it that would be great.

ry:
@humancalico Closing because stale. Feel free to rebase this work and open a new PR.

PR #279, opened 2020-08-25
Scope improvement

kdy1:
I will 

 1. Migrate lint rules from `scopes` to `flat_scope`
 2. Remove `scopes`
 3. Rename `flat_scop` to scopes

### Verification 

I'll not fully implement these rules, but I'll verify that it is implementable with my design.

- [x] no-class-assign
- [x] no-const-assign
- [x] no-ex-assign
- [x] no-func-assign
- [x] no-global-assign

Simillar to no-undef

- [x] no-import-assign
- [x] no-undef

If a variable is used but no binding exists, it's undef.

- [x] no-redeclare

It can be done by getting vars from the id of a declaration, and comparing paths.


kdy1:
r? @bartlomieju 

Please feel free to change the title.

src/scopes.rs
```diff
@@ -1,20 +1,52 @@
...
+use swc_ecmascript::utils::ident::IdentLike;
+use swc_ecmascript::utils::Id;
 use swc_ecmascript::visit::Node;
 use swc_ecmascript::visit::Visit;
+use swc_ecmascript::visit::VisitWith;
+
+#[derive(Debug)]
+pub struct Scope {
+  vars: HashMap<Id, Var>,
+  // TODO(kdy1): Maybe change this to swc_atoms::JsWord
```
bartlomieju, line 19: I removed `swc_atoms` dependency a few days ago, I think we can get away using `String`s

src/scopes.rs
```diff
@@ -30,8 +62,8 @@ pub enum BindingKind {
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub enum ScopeKind {
-  Program,
-  Module,
+  // Module,
```
bartlomieju, line 65: Remove? I guess `Module` is just "root" scope that is handled implicitly?

src/scopes.rs
```diff
@@ -41,1046 +73,210 @@ pub enum ScopeKind {
...
-    let function_catch_scope = function_scope.get_child_scope(-1).unwrap();
-    assert!(function_catch_scope.get_binding("e").is_some());
+  fn visit_switch_stmt(&mut self, n: &SwitchStmt, _: &dyn Node) {
+    n.discriminant.visit_with(n, self);
 
-    let catch_scope = module_scope.get_child_scope(-1).unwrap();
-    assert!(catch_scope.get_binding("message").is_some());
+    self.visit_with_path(ScopeKind::Switch, &n.cases);
   }
 }
```
bartlomieju, line 282: Sad to see tests go, but we can add them later

bartlomieju (approved):
@kdy1 it's really amazing how short is your implementation! I wasn't aware that SWC already has infra to support this feature

src/scopes.rs
kdy1: I will do so. Is `0.2` ok?

src/scopes.rs
kdy1: I removed it mainly because there was no usage related to it. I think adding it is not a bad option.

src/scopes.rs
kdy1: Oh.. I missed the assertion. I'll see if it's possible to restore.

src/scopes.rs
```diff
@@ -776,311 +334,21 @@ try {
...
+    let scope = test_scope(source_code);
+    assert_eq!(var(&scope, "a").kind(), BindingKind::Const);
+    assert_eq!(var(&scope, "a").path(), &[]);
+
+    assert_eq!(var(&scope, "b").kind(), BindingKind::Param);
+    assert_eq!(scope.ids_with_symbol(&"c".into()).unwrap().len(), 2);
+    assert_eq!(
+      var(&scope, "d").path(),
+      &[ScopeKind::Function, ScopeKind::Block]
+    );
```
bartlomieju, line 346: Thanks for adding the test case 👍

bartlomieju (approved):
LGTM :)

PR #289, opened 2020-08-28
Chore: sort rules alphabetically

magurotuna:
`get_recommended_rules` and `get_all_rules` give us the provided rules, but currently their orders are not sorted alphabetically.
It would be better for them to be sorted, because Deno is going to use `get_recommended_rules` by https://github.com/denoland/deno/pull/7222 and then the rules returned by it will be displayed in the same order when running `deno lint --unstable --rules`.
So in this PR, I put them in sorted order.

Besides, I added tests to make sure the rules are sorted alphabetically, which will help us detect unordered rules in future patches.


magurotuna:
I have also checked the rules in README and found there are missing rules and two rules have different codes from the original ESLint's rules. I have fixed them as well.

bartlomieju (approved):
Nice, thanks @magurotuna!

PR #292, opened 2020-08-30
Chore: fix and unify visibility

magurotuna:
Currently, there are some problems about visibility.

### 1. Items that are not required to be `pub` have public visibility

I think that from the crate consumers' point of view, the items that should be public are:

- `diagnostic::LintDiagnostic`
- `diagnostic::Location`
- `linter::Context`
- `linter::IgnoreDirective`
- `linter::Linter`
- `linter::LinterBuilder`
- `rules::LintRule`
- `rules::get_all_rules`
- `rules::get_recommended_rules`
- `rules::<rule-name>::<rule-struct>`
- `swc_util::SwcDiagnosticBuffer`

Therefore I have made the other items private.

### 2. Different visitors have different visibility for their constructors (some are `pub`, the others are private).

The visitors don't have to be public to crate consumers because they are the implementation detail of the rules and used by `lint_module` internally. So I have made all the items about `visitor` private.

### DIFF
--- a/src/js_regex/validator.rs
+++ b/src/js_regex/validator.rs
@@ -92,7 +92,8 @@ fn combine_surrogate_pair(lead: i64, trail: i64) -> i64 {
 }
 
 #[derive(Clone, Copy, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
-pub enum EcmaVersion {
+#[allow(dead_code)]
+pub(crate) enum EcmaVersion {
   ES5,
   ES2015,
   ES2016,
@@ -104,7 +105,7 @@ pub enum EcmaVersion {
 }
 
 #[derive(Debug)]
-pub struct EcmaRegexValidator {
+pub(crate) struct EcmaRegexValidator {
   reader: Reader,
   strict: bool,
   ecma_version: EcmaVersion,
@@ -137,7 +138,7 @@ impl DerefMut for EcmaRegexValidator {
 }
 
 impl EcmaRegexValidator {
-  pub fn new(ecma_version: EcmaVersion) -> Self {
+  pub(crate) fn new(ecma_version: EcmaVersion) -> Self {
     EcmaRegexValidator {
       reader: Reader::new(),
       strict: false,
@@ -158,7 +159,7 @@ impl EcmaRegexValidator {
   }
 
   /// Validates flags of a EcmaScript regular expression.
-  pub fn validate_flags(&self, flags: &str) -> Result<(), String> {
+  pub(crate) fn validate_flags(&self, flags: &str) -> Result<(), String> {
     let mut existing_flags = HashSet::<char>::new();
 
     for flag in flags.chars() {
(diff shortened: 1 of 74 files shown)


magurotuna:
Deno is using its own `get_default_ts_config` function rather than `deno_lint::swc_util::get_default_ts_config`.
Is deno_lint's one being exported only for `examples/dlint`?

bartlomieju:
> 1. Items that are not required to be pub have public visibility

Sounds good

> 2. Different visitors have different visibility for their constructors (some are pub, the others are private).

Also sounds good

> Deno is using its own get_default_ts_config function rather than deno_lint::swc_util::get_default_ts_config.
Is deno_lint's one being exported only for examples/dlint?

Yes, please keep it as is, it will be soon refactored after `AstParser` is moved to SWC.

I'm for removing unnecessary `pub` modifiers, but we don't need to change files like `scopes.rs` - they are not exported from library so they can keep `pub` modifiers on the methods - they'll only be available in the crate code itself.

src/swc_util.rs
```diff
@@ -35,7 +35,7 @@ use swc_ecmascript::{
 };
 
 #[allow(unused)]
-pub fn get_default_es_config() -> Syntax {
+pub(crate) fn get_default_es_config() -> Syntax {
```
bartlomieju, line 38: Revert

src/swc_util.rs
```diff
@@ -51,8 +51,7 @@ pub fn get_default_es_config() -> Syntax {
   Syntax::Es(config)
 }
 
-#[allow(unused)]
-pub fn get_default_ts_config() -> Syntax {
+pub(crate) fn get_default_ts_config() -> Syntax {
```
bartlomieju, line 54: ditto

src/scopes.rs
```diff
@@ -22,34 +22,36 @@ pub struct Scope {
 
 impl Scope {
   // Get all declarations with a symbol.
-  pub fn ids_with_symbol(&self, sym: &JsWord) -> Option<&Vec<Id>> {
+  #[allow(dead_code)]
```
bartlomieju, line 25: Please revert changes to this file

src/js_regex/validator.rs
```diff
@@ -92,7 +92,8 @@ fn combine_surrogate_pair(lead: i64, trail: i64) -> i64 {
 }
 
 #[derive(Clone, Copy, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
-pub enum EcmaVersion {
+#[allow(dead_code)]
+pub(crate) enum EcmaVersion {
```
bartlomieju, line 96: revert

bartlomieju (changes requested):

magurotuna:
@bartlomieju 
Thanks a lot for the review and answering my question!
I fixed the part I was overdoing. Could you please check it?

bartlomieju (approved):
Cool! Thanks @magurotuna

PR #297, opened 2020-08-31
Add no-inner-declarations rule

bartlomieju:
Before landing we must verify that it yields the same results in `std/` as ESLint.

CC @magurotuna would you be interested in integrating?


magurotuna:
@bartlomieju Yes, I'll check it.

src/rules/no_inner_declarations.rs
```diff
@@ -0,0 +1,286 @@
...
+      "no-inner-declarations",
+      &format!("Move {} declaration to {} root", kind, root),
+    );
+  }
+}
+
+impl Visit for NoInnerDeclarationsVisitor {
+  noop_visit_type!();
+
+  fn visit_function(&mut self, fn_: &ast::Function, parent: &dyn Node) {
```
magurotuna, line 171: Seems that implementing `visit_arrow_expr` here is also necessary to check if it's inside function

bartlomieju: Ah yeah, good catch 👍

bartlomieju: Done

magurotuna:
Great!
As I said at https://github.com/denoland/deno/issues/7285#issuecomment-683843792, now we get the same result as ESLint.

magurotuna (approved):
LGTM :+1:

PR #298, opened 2020-08-31
feat: use position range in diagnostic

### LINKED ISSUE #290: `start` and `end` for location and zero base
before:

```json
{
 "diagnostics": [
    {
     "location": {
        "filename": "/path/to/file/mod.ts",
         "line": 1,
         "col": 0,
      },
     "message": "`debugger` statement is not allowed",
     "code": "no-debugger",
     "line_src": "debugger;",
     "snippet_length": 9,
    },
  ],
 "errors": [],
}
```

after:

```json
{
 "diagnostics": [
    {
	  "file": "/path/to/file/mod.ts",
      "location": {
	  	"start": {
          "line": 0,
           "col": 0
        },
		"end": {
          "line": 0,
          "col": 8
        }
      },
      "message": "`debugger` statement is not allowed",
      "code": "no-debugger",
      "line_src": "debugger;",
      "snippet_length": 9,
    },
  ],
  "errors": [],
}
```

lucacasonato:
Closes #290

### DIFF
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -33,6 +33,12 @@ fn red(s: String) -> impl fmt::Display {
   style(&s, style_spec)
 }
 
+fn bold(s: String) -> impl fmt::Display {
+  let mut style_spec = ColorSpec::new();
+  style_spec.set_bold(true);
+  style(&s, style_spec)
+}
+
 fn cyan(s: String) -> impl fmt::Display {
   let mut style_spec = ColorSpec::new();
   style_spec.set_fg(Some(Ansi256(14)));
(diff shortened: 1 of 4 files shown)


bartlomieju:
Nice, just to double check, does it work properly if `range.start == range.end` (meaning `col` diff == 0)?

lucacasonato:
> Nice, just to double check, does it work properly if `range.start == range.end` (meaning `col` diff == 0)?

Yeah I think so  - `snippet_length` would be 0, so no `^` would be printed. There would just be an empty line. I do make the assumption that start line + col is always before, end line + col, but I think that is reasonable.

bartlomieju (approved):
LGTM, very nice @lucacasonato

PR #304, opened 2020-09-08
Add no-undef

kdy1:
I confused `no-global-assign` with `no-undef`.
I thought `no-global` was about prohibiting assignment to global variables.
So I implemented it, but the name of such pass was `no-undef`.


src/rules/no_undef.rs
```diff
@@ -0,0 +1,345 @@
...
+      | "Promise"
+      | "Date"
+      | "NaN"
+      | "isNaN"
+      | "hasOwnProperty"
+      | "eval"
+      | "toString"
+      | "WeakRef"
+      | "Map"
+      | "FinalizationRegistry" => return,
```
bartlomieju, line 142: @kdy1 could you make it a static slice at top level and add `Deno` as well?

src/rules/no_undef.rs
```diff
@@ -0,0 +1,345 @@
...
+    assert_lint_err::<NoUndef>("require(\"a\");", 0);
+
+    // TODO(kdy1): parse as jsx
+    // assert_lint_err::<NoUndef>("var React; React.render(<img attr={a} />);", 0);
+  }
+
+  #[test]
+  fn err_3() {
+    // TODO(kdy1): parse as jsx
+    // assert_lint_err::<NoUndef>("var React, App; React.render(<App attr={a} />);", 0);
```
bartlomieju, line 330: Test case should be parseable as `tsx` already

bartlomieju (commented):
Looking good, I'll check that it works as expected on `deno` codebase before landing.

src/rules/mod.rs
```diff
@@ -64,6 +64,7 @@ pub mod no_sparse_arrays;
 pub mod no_this_alias;
 pub mod no_this_before_super;
 pub mod no_throw_literal;
+pub mod no_undef;
```
bartlomieju, line 67: Add `no_undef` to "recommended" and "all" set

kdy1: Fixed it.

src/rules/no_undef.rs
kdy1: Yes, I'll do it soon.

bartlomieju:
Some false positives:
```
    await new Promise((resolve: () => void, _) => {
      setTimeout(resolve, 100);
    });

// 
(no-undef) resolve is not defined
      setTimeout(resolve, 100);
                 ^^^^^^^
    at /Users/biwanczuk/dev/deno/cli/tests/unit/net_test.ts:331:17
```

```
const checkErr = (e: Error): void => {
      if (e.message === "Listener has been closed") {
        assertEquals(acceptErrCount, 1);
      } else if (e.message === "Another accept task is ongoing") {
        acceptErrCount++;
      } else {
        throw new Error("Unexpected error message");
      }
    };

// 
(no-undef) e is not defined
    const checkErr = (e: Error): void => {
                      ^^^^^^^^
    at /Users/biwanczuk/dev/deno/cli/tests/unit/net_test.ts:136:22

(no-undef) e is not defined
      if (e.message === "Listener has been closed") {
          ^
    at /Users/biwanczuk/dev/deno/cli/tests/unit/net_test.ts:137:10
```

Other builtins that should be listed:
- `Symbol`
- `TextEncoder`
- `TextDecoder`
- `undefined`
- `Deno`
- `Uint8Array` 
- `Error`

src/rules/no_undef.rs
kdy1: How can I parse it as tsx?

bartlomieju: Sorry, I was mistaken... You'd need to add additional helper to pass syntax in tests - by default it uses `TypeScript` syntax. 

https://github.com/denoland/deno_lint/blob/313ef64e548ec971bb6b5ab279aa363315c8269b/src/test_util.rs#L8-L14

If it won't break other tests, feel free to change default syntax to TSX.

kdy1:
False positives were cause by the bug in `ts_resolver`, which is now fixed.

bartlomieju:
@kdy1 some false positives I found

Class constructor param:
```
(no-undef) options is not defined
  constructor(options: ContextOptions) {
              ^^^^^^^^^^^^^^^^^^^^^^^
    at /Users/biwanczuk/dev/deno/std/wasi/snapshot_preview1.ts:305:14

(no-undef) options is not defined
    this.args = options.args ? options.args : [];
                ^^^^^^^
    at /Users/biwanczuk/dev/deno/std/wasi/snapshot_preview1.ts:306:16
```

Case using enum variant:
```
(no-undef) OpCode is not defined
        case OpCode.Close: {
             ^^^^^^
    at /Users/biwanczuk/dev/deno/std/ws/mod.ts:262:13
```

Class name inside class method:
```
(no-undef) Buffer is not defined
  value: Buffer,
         ^^^^^^
    at /Users/biwanczuk/dev/deno/std/node/buffer.ts:592:9
```

Catch binding:
```
(no-undef) err is not defined
  } catch (err) {
           ^^^
    at /Users/biwanczuk/dev/deno/std/ws/example_client.ts:56:11
```

bartlomieju:
@kdy1 the only missing globals I found are `Headers` and `TransformStream`

kdy1:
Got it, thanks!

kdy1:
Published a new version of `swc_ecma_visit`, `swc_ecma_transforms`, `swc_ecmascript`.

Patch of `swc_ecma_visit` is about reducing compile time while debugging,  and patch of `swc_ecma_transforms` is about fixing resolver.

I copied some failing test cases from deno std_lint and verified locally that new version of resolver make the lint work properly.

bartlomieju:
@kdy1 thanks, I will start upgrades soon

bartlomieju (approved):
LGTM 🚀

PR #307, opened 2020-09-09
Add no-redeclare

kdy1:
Resolver (aka span hygiene) makes many things easy :)

Some parts of the lint is not implemented because there are no list of globals yet. (Related: #304)

--- 

Note: Parameter issues seems like a bug of swc.


bartlomieju (approved):
LGTM 🚀

PR #319, opened 2020-09-14
no-undef: Ignore arguments

### LINKED ISSUE #317: Deno lint doesn't recognize function arguments object
If you have a function and want to use the arguments object, deno lint will show `(no-undef) arguments is not defined`. To pass lint I switched to using `(...args: any[])` in my call signature and `// deno-lint-ignore no-explicit-any` to ignore my use of any.

kdy1:
This is very naive patch but it will be enough for most users as main purpose of the lint is catching typo.
Closes #317

### DIFF
--- a/src/rules/no_undef.rs
+++ b/src/rules/no_undef.rs
@@ -158,6 +158,12 @@ impl NoUndefVisitor {
       return;
     }
 
+    // Implicitly defined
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

Issue #318, opened 2020-09-14
bug: incorrect assessment of unreachable code


kitsonk:
The following code:

```ts
function normalize(type: string): string | undefined {
  switch (type) {
    case "urlencoded":
      return "application/x-www-form-urlencoded";
    case "multipart":
      return "multipart/*";
  }

  if (type[0] === "+") {
    return `*/*${type}`;
  }

  return type.includes("/") ? type : lookup(type);
}
```

Results in the following incorrect lint violation:

```
(no-unreachable) This statement is unreachable
  if (type[0] === "+") {
  ^^^^^^^^^^^^^^^^^^^^^^
    return `*/*${type}`;
^^^^^^^^^^^^^^^^^^^^^^^^
  }
^^^
    at /example.ts:9:2
```

bartlomieju:
CC @kdy1

Xernisfy:
An unconditional return statement inside of a class constructor also leads to a "unreachable" linting error:

```ts
class Class {
  constructor() {
    return this;
  }
}

console.log("unreachable???");
```
```
(no-unreachable) This statement is unreachable
console.log("unreachable???");
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    at \\?\C:\Users\akn\OneDrive - EPG\Synced\TMP\Demo3DWebClient\local\deno-lint.ts:7:0

Found 1 problem
Checked 1 file
```

A conditional return however does not produce an error:
```ts
class Class {
  constructor() {
    if (Deno) return this;
  }
}

console.log("unreachable???");
```

magurotuna:
@Xernisfy 
Thanks for the clear report. Seems like that should be a bug. Could you open a new issue for that?

PR #324, opened 2020-09-15
no-unreachable: false positive after switch

bartlomieju:
Ref https://github.com/denoland/deno_lint/issues/318

### DIFF
--- a/src/rules/no_unreachable.rs
+++ b/src/rules/no_unreachable.rs
@@ -208,6 +208,28 @@ mod tests {
   }
 
   #[test]
+  fn ok_10() {
+    assert_lint_ok::<NoUnreachable>(
+      r#"
+function normalize(type: string): string | undefined {
+  switch (type) {
+    case "urlencoded":
+      return "application/x-www-form-urlencoded";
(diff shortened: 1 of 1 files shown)


src/rules/no_unreachable.rs
```diff
@@ -207,6 +207,28 @@ mod tests {
...
+  fn ok_10() {
+    assert_lint_ok::<NoUnreachable>(
+      r#"
+function normalize(type: string): string | undefined {
+  switch (type) {
+    case "urlencoded":
+      return "application/x-www-form-urlencoded";
+    case "multipart":
+      return "multipart/*";
+  }
```
bartlomieju, line 220: @kdy1 I think in situation where `switch` has no `default` it is non-exhaustive and shouldn't mark the block as "forced done".

kdy1: @bartlomieju  You are correct. Making it not forced-done will fix it.

bartlomieju:
Closing in favor of https://github.com/denoland/deno_lint/pull/325

Issue #327, opened 2020-09-20
no-unsafe-finally doesn't catch nested violations


bartlomieju:
```
try {
  /* */
} finally {
  try {
    /* */
  } finally {
    throw "oh no";
  }
}
```

From https://rdambrosio016.github.io/rust/2020/09/18/pure-ast-based-linting-sucks.html

PR #329, opened 2020-09-21
fix(defualt-param-last): handle nested violations

magurotuna:
Just like #327, `default-param-last` doesn't handle violations in nested functions, so I've fixed it.

Moreover, it doesn't check arrow functions. Fixed it as well.


bartlomieju (approved):
LGTM, thanks @magurotuna

Issue #330, opened 2020-09-21
Rules that should be refined further


magurotuna:
Now that all the recommended rules have been implemented, it's good time to look back at the rules again.
I have looked through how the rules are implemented, and have noticed that some rules need more refinement. Current issues I've found are:

- unable to catch violations in nested blocks, just like #327 
- some cases that should be reported are ignored (e.g. normal funcions are considered, but _arrow_ functions are not)
- insufficient test cases (I think porting the ESLint's tests would be good as it would guarantee the compatibility with ESLint.)

### rules that need refinement

Note that this list has been made from my quick look, thus it may contain rules that actually don't need any fix.

- [x] default-param-last #329 
- [ ] explicit-function-return-type
- [ ] for-direction
- [ ] getter-return
- [ ] no-array-constructor
- [ ] no-async-promise-executor
- [ ] no-case-declarations
- [ ] no-class-assign
- [ ] no-compare-neg-zero
- [ ] no-cond-assign
- [ ] no-const-assign
- [ ] no-constant-condition
- [ ] no-dupe-args
- [ ] no-dupe-keys
- [ ] no-duplicate-case
- [ ] no-empty
- [ ] no-eval
- [ ] no-ex-assign
- [ ] no-func-assign
- [ ] no-global-assign
- [ ] no-inferrable-type
- [ ] no-invalid-regexp
- [ ] no-new-symbol
- [ ] no-non-null-asserted-optional-chain
- [ ] no-obj-calls
- [ ] no-prototype-builtins
- [ ] no-self-assign
- [ ] no-setter-return
- [ ] no-shadow-restricted-names
- [ ] no-sparse-arrays
- [ ] no-this-alias
- [ ] no-throw-literal
- [x] no-unsafe-finally #328 
- [ ] no-unsafe-negation
- [ ] no-var
- [ ] no-with
- [ ] prefer-as-const
- [ ] single-var-declarator
- [ ] use-isnan
- [ ] valid-typeof

bartlomieju:
@magurotuna have you found bugs in all the listed rules? 

IIRC most of implemented rules have ESLint tests ported, but I'm more than happy to add more tests if they are lacking.

magurotuna:
@bartlomieju 

> have you found bugs in all the listed rules?

not necessarily, but I feel like lots of the listed rules are unable to catch nested violations.
For example, with `no-with` rule, if we run a test like

```rust
    assert_lint_err_n::<NoWith>(
[...]

PR #332, opened 2020-09-21
[WIP] Refine explicit-function-return-type

magurotuna:
WORK IN PROGRESS

This fix is a part of #330, but this rule is not "recommended" rules, so I put it off; I'll work on recommended rules first.

Unfortunately, this rule seems more difficult to get compatible with the original one than I thought when I started to work - in fact, the original implementation is achieved using [a lot of helper functions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts).
It indicates that we need to take care of many situations to implement this rule.

### DIFF
--- a/src/rules/explicit_function_return_type.rs
+++ b/src/rules/explicit_function_return_type.rs
@@ -1,9 +1,9 @@
 // Copyright 2020 the Deno authors. All rights reserved. MIT license.
 use super::Context;
 use super::LintRule;
-use swc_ecmascript::visit::noop_visit_type;
-use swc_ecmascript::visit::Node;
-use swc_ecmascript::visit::Visit;
+use swc_common::Span;
+use swc_ecmascript::ast::{ArrowExpr, ClassMethod, Function, MethodKind};
+use swc_ecmascript::visit::{self, noop_visit_type, Node, Visit, VisitWith};
 
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


Skillz4Killz:
Can we consider removing this rule entirely from the main deno lint rules please which are the default set of rules on `deno lint`? Leave this as something that can be added as a plugin. I've brought up the concerns with this rule before and how this rule is really a pain in the arse more than it is helpful and it also can cause infinite loop bugs because it hides errors that TS would catch. This rule encourages you to cast the return type thereby telling TS you know better. But 99% of the time, its the machine that knows better. As Ry said Deno should be easy and simple to use, so IMO the default `deno lint` should not include this which makes the dev experience worse. This should be added on by the end-user themself throught the plugin system. @bartlomieju @magurotuna 

image

nayeemrmn:
> and it also can cause infinite loop bugs because it hides errors that TS would catch

@Skillz4Killz Nope, I remember clearly that you were wrong about that.

Skillz4Killz:
> > and it also can cause infinite loop bugs because it hides errors that TS would catch
> 
> @Skillz4Killz Nope, I remember clearly that you were wrong about that.

I didn't want to continue the convo at the time since no one else felt it was desired to change the entire codebase in deno when it was already written but no I was not wrong nayeem. It does indeed hide the bug.

bartlomieju:
> Can we consider removing this rule entirely from the main deno lint rules please which are the default set of rules on deno lint? 

This rule is not switched on in `deno lint`. Tested on `1.4.1`

Skillz4Killz:
> > Can we consider removing this rule entirely from the main deno lint rules please which are the default set of rules on deno lint?
> 
> This rule is not switched on in `deno lint`. Tested on `1.4.1`

BEAUTIFUL! That's perfect!

ry:
@magurotuna closing because old/stale.

PR #333, opened 2020-09-22
fix(for-direction): handle nested for loops

magurotuna:
This is a part of #330 

`for-direction` currently doesn't deal with nested for-loops. I have fixed it.


bartlomieju (approved):
LGTM, thanks @magurotuna

PR #334, opened 2020-09-22
fix(getter-return): handle nested stuff

magurotuna:
~~WIP~~

This is a part of #330 

This PR allows `getter-return` to handle nested stuff and adds some more tests.


magurotuna:
@bartlomieju Done it. PTAL:)

bartlomieju (approved):
LGTM, thank you @magurotuna

PR #335, opened 2020-09-25
fix(no-array-constructor): handle nested stuff

magurotuna:
This is a part of #330 

This PR allows `no-array-constructor` to handle nested stuff.


bartlomieju (approved):
LGTM!

PR #338, opened 2020-09-26
fix(no-async-promise-executor): handle nested or parenthesized stuff

magurotuna:
This is a part of #330

This PR allows `no-async-promise-executor` to handle nested or parenthesized stuff.


bartlomieju (approved):
LGTM, thank you @magurotuna!

PR #341, opened 2020-09-28
Fix no-unreachable

### LINKED ISSUE #340: Error with no-unreachable, but totally reachable
```javascript
_getString(n: number) {
    let ret = "";
    let p: BufferListItem | null = (this.head as BufferListItem);
    let c = 0;
    p = p.next as BufferListItem;
    do {
      const str = p.data;
      if (n > str.length) {
        ret += str;
        n -= str.length;
      } else {
        if (n === str.length) {
          ret += str;
          ++c;
          if (p.next) {
            this.head = p.next;
          } else {
            this.head = this.tail = null;
          }
        } else {
          ret += str.slice(0, n);
          this.head = p;
          p.data = str.slice(n);
        }
        break;
      }
      // deno-lint-ignore no-unreachable
      ++c;
      // deno-lint-ignore no-unreachable
      p = p.next;
    } while (p);
    // deno-lint-ignore no-unreachable
    this.length -= c;
    // deno-lint-ignore no-unreachable
    return ret;
  }
```

https://github.com/Soremwar/deno/blob/13c74a19baff1d64c5511f9a9ba73d2dd1c8a74a/std/node/_stream/buffer_list.ts#L145

After the `else` statement ends, all bets seem to be off for deno_lint

kdy1:
Closes #340.

### DIFF
--- a/src/control_flow/mod.rs
+++ b/src/control_flow/mod.rs
@@ -386,7 +386,7 @@ impl Visit for Analyzer<'_> {
             if another.is_some() {
               self.mark_as_done(n.span.lo, Done::Break);
             } else {
-              self.scope.done = Some(Done::Break)
+              self.scope.done = None
             }
           }
           // TODO: Check for continue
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, @kdy1 thanks for quick fix!

PR #342, opened 2020-09-28
add prefer-const rule

### LINKED ISSUE #337: Add prefer-const rule
This is one of most useful rules, we should definitely add it.

https://eslint.org/docs/rules/prefer-const

magurotuna:
Closes #337 

This PR adds [prefer-const](https://eslint.org/docs/rules/prefer-const).

I feel like it's ad-hoc, or a little dirty implementation, and has room for being refactored. But all the tests from ESLint are passed except for one case.

### DIFF
--- a/src/rules/prefer_const.rs
+++ b/src/rules/prefer_const.rs
@@ -483,19 +483,19 @@ mod tests {
   fn hoge() {
     assert_lint_err::<PreferConst>(
       r#"let a, b; ({a = 0, b} = obj); b = 0; foo(a, b);"#,
-      0,
+      4,
     );
     assert_lint_err::<PreferConst>(
       r#"let predicate; [, {foo:returnType, predicate}] = foo();"#,
-      0,
+      4,
     );
     assert_lint_err::<PreferConst>(
       r#"let predicate; [, {foo:returnType, predicate}, ...bar ] = foo();"#,
-      0,
+      4,
     );
     assert_lint_err::<PreferConst>(
       r#"let predicate; [, {foo:returnType, ...predicate} ] = foo();"#,
-      0,
+      4,
     );
   }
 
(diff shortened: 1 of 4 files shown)


src/rules/prefer_const.rs
```diff
@@ -0,0 +1,854 @@
...
+            return;
+        }
+        id = setInterval(() => {}, 250);
+    }
+    foo();
+  "#,
+      r#"/*exported a*/ let a; function init() { a = foo(); }"#,
+      // TODO(magurotuna): this is ported from ESLint, but I have no idea why this is valid,
+      // so comment it out for now.
+      // r#"/*exported a*/ let a = 1"#,
```
magurotuna, line 636: As I commented, I have no idea why this belongs to "valid" tests :confused: 
https://github.com/eslint/eslint/blob/v7.10.0/tests/lib/rules/prefer-const.js#L70

bartlomieju: @magurotuna this is probably some "magical" ESLint directive that makes it valid - we don't have such thing so you can safely remove this test :)

bartlomieju:
@kdy1 your review would be much appreciated

kdy1 (approved):
LGTM, I forgot about lazy-initialization while writing a comment, but it seems like you handled it correctly.

src/rules/prefer_const.rs
magurotuna: oh, thanks, I didn't know that. I'll remove it.

magurotuna:
I've realized there are a couple of typos, I will fix them as well.

bartlomieju:
@magurotuna LGTM as well. Let me know when you think the PR is ready to be landed

magurotuna:
@bartlomieju 
Done it! I think it's ready, PTAL :)

bartlomieju (approved):
LGTM, thank you @magurotuna, this rule is very useful!

PR #345, opened 2020-09-29
fix(prefer-const): handle `+=` and `-=` correctly

### LINKED ISSUE #344: Bug: False positive on prefer-const
```js
let x = 0;
x += 1;
```

This should be okay, but we're getting an error

```
thread 'rules::prefer_const::tests::hoge' panicked at 'Unexpected diagnostics found:
[
    LintDiagnostic {
        range: Range {
            start: Position {
                line: 1,
                col: 4,
            },
            end: Position {
                line: 1,
                col: 5,
            },
        },
        message: "\'x\' is never reassigned. Use \'const\' instead",
        code: "prefer-const",
        hint: None,
    },
]

source:
let x = 0; x += 1;
```

I'll fix it soon.

magurotuna:
Fixes #344


bartlomieju (approved):
LGTM 👍

PR #347, opened 2020-09-29
fix(prefer-const): treat param of catch clause like thet of function

### LINKED ISSUE #346: Bug: False positive on prefer-const (catch clause)
```js
try {
  foo();
} catch (e) {
  console.error(e);
}
```

This should be okay but we're getting an error like

```
---- rules::prefer_const::tests::prefer_const_valid stdout ----
thread 'rules::prefer_const::tests::prefer_const_valid' panicked at 'Unexpected diagnostics found:
[
    LintDiagnostic {
        range: Range {
            start: Position {
                line: 1,
                col: 22,
            },
            end: Position {
                line: 1,
                col: 23,
            },
        },
        message: "\'e\' is never reassigned. Use \'const\' instead",
        code: "prefer-const",
        hint: None,
    },
]

source:
try { foo(); } catch (e) {}
```

magurotuna:
Fixes #346

### DIFF
--- a/src/rules/prefer_const.rs
+++ b/src/rules/prefer_const.rs
@@ -534,7 +534,7 @@ impl Visit for PreferConstVisitor {
     self.enter_scope();
 
     if let Some(param) = &catch_clause.param {
-      self.extract_decl_idents(param, true, false);
+      self.extract_param_idents(param);
     }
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, nice fix

Issue #159, opened 2020-06-12
Add descriptions and examples to each lint rule


bartlomieju:
All lint rules should be updated to contains some kind of method that returns description as well as some examples of valid and invalid. This feature will be used to generate documentation.

- [x] adjacent-overload-signatures
- [x] ban-ts-comment
- [x] ban-types
- [x] ban-untagged-ignore
- [x] ban-untagged-todo
- [x] constructor-super
- [x] default-param-last
- [x] eqeqeq
- [x] explicit-function-return-type
- [x] explicit-module-boundary-types
- [x] for-direction
- [x] getter-return
- [x] no-array-constructor
- [x] no-async-promise-executor
- [x] no-await-in-loop
- [x] no-case-declarations
- [x] no-class-assign
- [x] no-compare-neg-zero
- [x] no-cond-assign
- [x] no-const-assign
- [x] no-constant-condition
- [x] no-control-regex
- [x] no-debugger
- [x] no-delete-var
- [x] no-dupe-args
- [x] no-dupe-class-members
- [x] no-dupe-else-if
- [x] no-dupe-keys
- [x] no-duplicate-case
- [x] no-empty
- [x] no-empty-character-class
- [x] no-empty-interface
- [x] no-empty-pattern
- [x] no-eval
- [x] no-ex-assign
- [x] no-explicit-any
- [x] no-extra-boolean-cast
- [x] no-extra-non-null-assertion
- [x] no-extra-semi
- [x] no-fallthrough
- [x] no-func-assign
- [x] no-global-assign
- [x] no-import-assign
- [x] no-inferrable-types
- [x] no-inner-declarations
- [x] no-invalid-regexp
- [x] no-irregular-whitespace
- [x] no-misused-new
- [x] no-mixed-spaces-and-tabs
- [x] no-namespace
- [x] no-new-symbol
- [x] no-non-null-asserted-optional-chain
- [x] no-non-null-assertion
- [x] no-obj-calls
- [x] no-octal
- [x] no-prototype-builtins
- [x] no-redeclare
- [x] no-regex-spaces
- [x] no-self-assign
- [x] no-setter-return
- [x] no-shadow-restricted-names
- [x] no-sparse-arrays
- [x] no-this-alias
- [x] no-this-before-super
- [x] no-throw-literal
- [x] no-undef
- [x] no-unreachable
- [x] no-unsafe-finally
- [x] no-unsafe-negation
- [x] no-unused-labels
- [x] no-unused-vars
[...]

PR #349, opened 2020-10-01
Implement `doc` Method for `no-empty` Rule

DomParfitt:
Implements the `doc` method for the `no-empty` rule as per #159.

@bartlomieju can you let me know if you're happy with the general text content, format, etc.? :slightly_smiling_face:


bartlomieju (approved):
LGTM, thanks @DomParfitt this is great!

PR #351, opened 2020-10-01
Add `docs` Method for `valid-typeof` Rule

DomParfitt:
Adds the `docs` method to the `valid-typeof` rule as per #159.


bartlomieju (approved):
🚀

PR #352, opened 2020-10-01
fix(no-class-assign): increase tests and fix to pass them

magurotuna:
This is a part of #330 

Adds more tests (mostly from ESLint) and fixes to pass the tests.

### DIFF
--- a/src/scopes.rs
+++ b/src/scopes.rs
@@ -2,8 +2,8 @@ use std::collections::HashMap;
 use swc_atoms::JsWord;
 use swc_common::DUMMY_SP;
 use swc_ecmascript::ast::{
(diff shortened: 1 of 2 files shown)


src/scopes.rs
```diff
@@ -158,6 +158,23 @@ impl Visit for Analyzer<'_> {
     n.decls.iter().for_each(|v| {
       v.init.visit_with(n, self);
 
+      // If the class name and the variable name are the same like `let Foo = class Foo {}`,
+      // this binding should be treated as `BindingKind::Class`.
```
bartlomieju, line 162: Thanks for this comment

bartlomieju (approved):
LGTM

PR #354, opened 2020-10-01
fix(no-compare-neg-zero): handle nested stuff

magurotuna:
This is a part of #330 

Adds a little more tests and fixes the implementation to make it handle nested stuff.


bartlomieju (approved):
🚀

PR #355, opened 2020-10-01
fix(control_flow): handle class constructors

### LINKED ISSUE #353: bug: unconditional return in constructor produces unreachable code after class
An unconditional return statement inside of a class constructor also leads to a "unreachable" linting error:

```ts
class Class {
  constructor() {
    return this;
  }
}

console.log("unreachable???");
```
```
console.log("unreachable???");
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    at \deno-lint.ts:7:0

Found 1 problem
Checked 1 file
```

A conditional return however does not produce an error:
```ts
class Class {
  constructor() {
    if (Deno) return this;
  }
}

console.log("unreachable???");
```

_Originally posted in https://github.com/denoland/deno_lint/issues/318#issuecomment-702154495_

magurotuna:
Fixes #353

### DIFF
--- a/src/rules/no_unreachable.rs
+++ b/src/rules/no_unreachable.rs
@@ -618,18 +618,6 @@ class Class {
 console.log("unreachable???");
       "#,
     );
-
-    assert_lint_ok::<NoUnreachable>(
-      r#"
-class Class {
-  constructor() {
-    return this;
-  }
-}
-
-console.log("unreachable???");
-      "#,
-    );
   }
 
   #[test]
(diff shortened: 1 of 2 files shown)


src/rules/no_unreachable.rs
```diff
@@ -603,4 +603,35 @@ function normalize(type: string): string | undefined {
       "#,
     );
   }
+
+  // https://github.com/denoland/deno_lint/issues/353
+  #[test]
+  fn issue_353_1() {
```
bartlomieju, line 609: 👍

bartlomieju (approved):
Nice fix

Issue #358, opened 2020-10-02
False positives in "prefer-const"


bartlomieju:
```
(prefer-const) 'allowBlockScalars' is never reassigned. Use 'const' instead
  let allowBlockScalars: boolean,
      ^^^^^^^^^^^^^^^^^
    at /Users/biwanczuk/dev/deno/std/encoding/_yaml/loader/loader.ts:1442:6

(prefer-const) 'modulePaths' is never reassigned. Use 'const' instead
let modulePaths: string[] = [];
    ^^^^^^^^^^^
    at /Users/biwanczuk/dev/deno/std/node/module.ts:628:4
```

CC @magurotuna 

I temporarly ignored them in `std/` to upgrade the linter, but we should fix those problems.

magurotuna:
I'll look into it.

magurotuna:
Seems like the second one, `std/node/module.ts`, has something to do with hoisting... I didn't care about that. I may need some time to fix it.

bartlomieju:
> Seems like the second one, `std/node/module.ts`, has something to do with hoisting... I didn't care about that. I may need some time to fix it.

That's okay, I will just disable `prefer-const` rule for this release

magurotuna:
```
 262   │ (prefer-const) 'openStringSyntax' is never reassigned. Use 'const' instead
 263   │   --> /home/yusuktan/Repos/github.com/magurotuna/deno/std/encoding/toml.ts:74:8
 264   │    |
 265   │ 74 |     let openStringSyntax = "";
 266   │    |         ^^^^^^^^^^^^^^^^
 267   │    |
```

-> hoisting

```
 346   │ (prefer-const) 'modulePaths' is never reassigned. Use 'const' instead
 347   │    --> /home/yusuktan/Repos/github.com/magurotuna/deno/std/node/module.ts:628:4
 348   │     |
 349   │ 628 | let modulePaths: string[] = [];
 350   │     |     ^^^^^^^^^^^
 351   │     |
```

-> hoisting

```
 605   │ (prefer-const) 'allowBlockScalars' is never reassigned. Use 'const' instead
 606   │     --> /home/yusuktan/Repos/github.com/magurotuna/deno/std/encoding/_yaml/loader/loader.ts:1442:6
 607   │      |
 608   │ 1442 |   let allowBlockScalars: boolean,
 609   │      |       ^^^^^^^^^^^^^^^^^
 610   │      |
```

-> maybe nested assignment expression

-----------------------

[...]

PR #361, opened 2020-10-03
fix(prefer-const): handle nested assignments correctly

magurotuna:
This PR fixes (partially) #358 

Previously, nested `AssignmentExpression` were reported as errors if the variables within them were declared as `let` without initialization and assigned only once. For example:

```js
let a; // <------ declared as `let` without initialization
const b = (a = foo === bar); // <------- `a` is within  the nested `AssignmentExpression` and assigned only once
```

In this case we should not report `a` as an error, because it couldn't be declared as `const`.
Therefore I've changed the implementation a bit so _nested_ `AssignmentExpression` and _not-nested_ one are handled 
differently.

### DIFF
--- a/src/rules/prefer_const.rs
+++ b/src/rules/prefer_const.rs
@@ -6,11 +6,12 @@ use std::sync::Arc;
 use swc_atoms::JsWord;
 use swc_common::Span;
 use swc_ecmascript::ast::{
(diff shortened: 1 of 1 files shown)


magurotuna:
I've also confirmed that the false positive on `std/encoding/_yaml/loader/loader.ts:1442:6` is fixed by this patch.

bartlomieju (approved):
LGTM, nice fix

PR #362, opened 2020-10-04
fix(prefer-const): refactor & handle hoisting

### LINKED ISSUE #358: False positives in "prefer-const"

magurotuna:
~~WIP~~

Fixes #358 

This PR looks quite big, but the policy of this change is just simple.
It introduces a new visitor called `VariableCollector`, which is in charge of collecting variable declarations per scope. And the job of `PreferConstVisitor` has changed to visiting any assignments and updating the variable statuses (initialized, reassigned, etc.). This change is required to handle hoisting correctly.
(Previously `PreferConstVisitor` did both collecting declarations and visiting assignments by itself)

I've confirmed that the false positives that were reported against Deno's std are all fixed by this change.

### DIFF
--- a/src/rules/prefer_const.rs
+++ b/src/rules/prefer_const.rs
@@ -182,20 +182,16 @@ impl Visit for VariableCollector {
   fn visit_function(&mut self, function: &Function, _: &dyn Node) {
     self.with_child_scope(function.span, |a| {
       for param in &function.params {
-        param.visit_with(function, a);
+        param.visit_children_with(a);
+        let idents: Vec<Ident> = find_ids(&param.pat);
+        for ident in idents {
+          a.insert_var(&ident, true, false, true);
+        }
       }
       if let Some(body) = &function.body {
         body.visit_children_with(a);
       }
     });
-  }
-
-  fn visit_param(&mut self, param: &Param, _: &dyn Node) {
-    param.visit_children_with(self);
-    let idents: Vec<Ident> = find_ids(&param.pat);
-    for ident in idents {
-      self.insert_var(&ident, true, false, true);
-    }
   }
 
   fn visit_block_stmt(&mut self, block_stmt: &BlockStmt, _: &dyn Node) {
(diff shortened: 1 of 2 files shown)


magurotuna:
@bartlomieju I'm doing small refactoring, but I think it's ready to be reviewed. PTAL 😃

magurotuna:
Actually I don't need to do rafactor.

bartlomieju (approved):
@magurotuna thanks for fixing this problem, LGTM. 

I'm a bit worried that you needed to add all that machinery, I thought scope analysis could be used in this case, but I guess not.

bartlomieju:
@kdy1 could you take a look and suggest if this rule could leverage scope analysis from `scopes.rs`?

kdy1:
@bartlomieju  I think it can, but I expect many rework and I'm not sure if it worths the work required to do.

magurotuna:
Yes, I thought `scopes.rs` would be a great help. However, current `scopes.rs` seems to have limited functionality - if we query it about variable `foo`, then it answers all `foo`s in the module. To implement `prefer-const`, I need the functionality that if we query about a span of some `foo`, it gives us the declaration of that `foo`. 

I'm not sure, but `VariableCollector` in my implementation might cover `Scope` in `scopes.rs` in functionality.

kdy1:
I agree it's quite limited. It's designed for some specific rules and I didn't add many scope-related features to it.
Adding features to it would not be hard, though.

PR #364, opened 2020-10-05
Add sort-imports rule

AkinAguda:


kitsonk:
As I mentioned (https://github.com/denoland/deno/issues/7150#issuecomment-687270286), I am personally not in favour of rules for things that `deno lint` automatically does.  Like we don't have a rule for semi-colon insertion, quoting style, indentation, or trailing commas, etc.

nayeemrmn:
What is this sorting exactly? If it's not sorting the _symbols_ within a single import statement, but the import statements themselves (by specifier), that's something `deno fmt` won't handle because it changes side-effect ordering. It would be perfect for `deno lint`. cc @dsherret

dsherret:
I could add sorted import declarations to deno fmt... I was just recommending not to because of side effects. I could turn it on for deno though if it's wanted and leave it as off by default in dprint-plugin-typescript.

I agree with @kitsonk and don't think it's something that should be added to deno lint as it's a formatting concern.

AkinAguda:
Oh I see. Just asking, if deno lint also serves as a linter for javascript all together, wouldn't it be great if it had all the rules that eslint supports? Like [sort imports](https://eslint.org/docs/rules/sort-imports) @kitsonk

kitsonk:
`deno lint` is intended to be opionated, just because a rule exists doesn't specifically make it suitable for Deno.

bartlomieju:
@AkinAguda talked with me about this rule prior to `dprint` adding sorting capabilities. 

I think this rule is still useful - let's put it in the set of `all` rules (not `recommended` which is used in `deno lint`), but it must only check sort order of members, not import statements.

AkinAguda:
Okay. I'll change the scope of this and open a PR at a later time. Thanks everyone!

PR #365, opened 2020-10-05
refactor: use &mut Context instead of Arc<Context>

bartlomieju:


magurotuna:
@bartlomieju The motivation of the refactor is that lint on one module should be done in a single thread?

bartlomieju:
@magurotuna I've used `Arc<Context>` in the beginning to get around borrow checker problem, but it's really not necessary. Using `&mut Context` does means that linting of a module will be done on a single thread. Before landing this I want to see if there's any gain by running rules on multiple threads - it feels to me that synchronization costs will be greater than profits of running rules on multiple threads.

magurotuna:
@bartlomieju 
I see.

> Before landing this I want to see if there's any gain by running rules on multiple threads - it feels to me that synchronization costs will be greater than profits of running rules on multiple threads.

That should be interesting examination. My guess is that it could be best to read files in parallel, and then run linting on each file in respective threads. That said, applying all rules to each module (file) is done in a single thread. (and I think this is how deno_lint currently works with Deno CLI.)

If you prove `&mut Context` is better, then I will have to replace `Arc<Mutex<RawScope>` with `Rc<RefCell<RawScope>>` :)

https://github.com/denoland/deno_lint/blob/9d930d113f93aa2339a467ca05f335f08624ea29/src/rules/prefer_const.rs#L76

bartlomieju:
> @bartlomieju 
> 
> I see.
> 
> 
> 
> > Before landing this I want to see if there's any gain by running rules on multiple threads - it feels to me that synchronization costs will be greater than profits of running rules on multiple threads.
> 
> 
> 
> That should be interesting examination. My guess is that it could be best to read files in parallel, and then run linting on each file in respective threads. That said, applying all rules to each module (file) is done in a single thread. (and I think this is how deno_lint currently works with Deno CLI.)
> 

Correct, files are loaded using threadpool and then each file is limited on a single thread.


> 
> If you prove `&mut Context` is better, then I will have to replace `Arc<Mutex<RawScope>` with `Rc<RefCell<RawScope>>` :)
> 

Running benchmarks on CI shows no difference between using `Arc` and `&mut`. But don't worry about it, I'll land your PR first.

> 
> https://github.com/denoland/deno_lint/blob/9d930d113f93aa2339a467ca05f335f08624ea29/src/rules/prefer_const.rs#L76

bartlomieju:
@magurotuna I've changed your `Scope` to use `Rc<RefCell<RawScope>>` instead

src/rules/ban_untagged_ignore.rs
```diff
@@ -17,10 +15,11 @@ impl LintRule for BanUntaggedIgnore {
 
   fn lint_module(
     &self,
-    context: Arc<Context>,
+    context: &mut Context,
     _module: &swc_ecmascript::ast::Module,
   ) {
-    for ignore_directive in &context.ignore_directives {
+    let ignore_directives = context.ignore_directives.clone();
+    for ignore_directive in &ignore_directives {
```
bartlomieju, line 22: Not ideal, maybe use `Rc` for `ignore_directives`?

src/rules/ban_ts_ignore.rs
```diff
@@ -36,19 +36,18 @@ impl LintRule for BanTsIgnore {
...
-        self.lint_comment(&context, comment);
-      }
-    });
-    context.trailing_comments.values().for_each(|comments| {
-      for comment in comments {
-        self.lint_comment(&context, comment);
-      }
-    });
+    let leading = context.leading_comments.clone();
+    let trailing = context.trailing_comments.clone();
```
bartlomieju, line 43: Not ideal, maybe use `Rc` for comments

PR #369, opened 2020-10-06
chore: enable prefer-const again

magurotuna:
I forgot to turn on `prefer-const` in #362 . I think it's ready to be landed 👍


bartlomieju (approved):
LGTM

PR #371, opened 2020-10-07
Adds `docs` Method for `no-unsafe-finally` Rule

DomParfitt:
Adds the `docs` method to the `no-unsafe-finally` rule as per #159.


bartlomieju (approved):
LGTM, thanks @DomParfitt nice docs!

PR #372, opened 2020-10-07
Added `docs` Method for `no-compare-neg-zero` Rule

DomParfitt:
Adds `docs` method for the `no-compare-neg-zero` rule as per #159.


bartlomieju (approved):
LGTM

PR #373, opened 2020-10-07
Add `docs` Method for `no-dupe-keys` Rule

DomParfitt:
Adds the `docs` method for the `no-dupe-keys` rule as per #159.


bartlomieju (approved):
LGTM

PR #382, opened 2020-10-11
fix(getter-return): use ControlFlow to check if getters return something

### LINKED ISSUE #348: getter-return and infinite loops
The following code produces "Getter requires a return deno-lint(getter-return)" despite the return statement.
I wonder if it's possible for the linter to recognize that the only way out of the loop and out of the function is through the return?

I would have just ignored the linter and not opened an issue except that the linter clearly recognizes that the last line when uncommented is redundant since uncommenting it silences deno-lint(getter-return), but produces a deno-lint(no-unreachable) error instead.

```
get root(): this {
        // deno-lint-ignore no-this-alias
        let primary = this;
        while (true) {
            if (primary.parent !== undefined) {
                primary = primary.parent;
            } else {
                return primary;
            }
        }
        //return this;
    }
```

deno 1.4.0
v8 8.7.75
typescript 4.0.2
VS Code extension 2.3.1

I'm fine with a "Won't Fix" on this BTW.

magurotuna:
Currently, `getter-return` rule only checks `if` and `switch` statements, so it doesn't handle more complicated cases such as including `while` loops.
This PR allows `getter-return` to use `ControlFlow` to check if getters return something, which will make it possible to handle complicated cases.

Fixes #348

### DIFF
--- a/src/control_flow/mod.rs
+++ b/src/control_flow/mod.rs
@@ -1,5 +1,5 @@
 use std::{
-  collections::{HashMap, HashSet},
+  collections::{BTreeMap, HashSet},
   mem::take,
 };
 use swc_common::{BytePos, Spanned, DUMMY_SP};
@@ -11,7 +11,7 @@ use swc_ecmascript::{
 
 #[derive(Debug)]
 pub struct ControlFlow {
-  meta: HashMap<BytePos, Metadata>,
+  meta: BTreeMap<BytePos, Metadata>,
 }
 
 impl ControlFlow {
@@ -60,9 +60,10 @@ impl Metadata {
   }
 }
 
+#[derive(Debug)]
 struct Analyzer<'a> {
   scope: Scope<'a>,
-  info: HashMap<BytePos, Metadata>,
+  info: BTreeMap<BytePos, Metadata>,
 }
 
 #[derive(Debug)]
@@ -87,12 +88,16 @@ struct Scope<'a> {
   found_break: Option<Option<Id>>,
   found_continue: bool,
 }
+
 #[derive(Debug, Copy, Clone)]
 enum Done {
   /// Return, Throw, or infinite loop
   Forced,
   // Break or continue
   Break,
+  // Pass through a block, like a function's block statement which ends without returning a value
+  // or throwing an exception
+  Pass,
 }
 
 impl<'a> Scope<'a> {
@@ -157,12 +162,18 @@ impl Analyzer<'_> {
 
     if let Some(done) = done {
       match kind {
-        BlockKind::Function => {}
+        BlockKind::Function => {
+          match done {
+            Done::Forced | Done::Pass => self.mark_as_done(lo, done),
+            _ => unreachable!(),
+          }
+          self.scope.done = prev_done;
+        }
         BlockKind::Block => {
           if let Done::Forced = done {
             self.mark_as_done(lo, done);
           } else if self.scope.done.is_none() {
-            self.scope.done = Some(Done::Break)
+            self.scope.done = Some(Done::Break);
           }
         }
         BlockKind::Case => {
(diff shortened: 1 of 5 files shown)


bartlomieju:
CC @kdy1 could you review control flow bit?

src/control_flow/mod.rs
```diff
@@ -86,12 +101,17 @@ struct Scope<'a> {
...
 enum Done {
   /// Return, Throw, or infinite loop
   Forced,
-  // Break or continue
+  /// Break or continue
   Break,
+  /// Pass through a block, like a function's block statement which ends without returning a value
+  /// or throwing an exception. Note that a node marked as `Done::Pass` won't prevent further execution, unlike
+  /// `Done::Forced` and `Done::Break`.
+  Pass,
```
kdy1, line 114: It's a nitpicking, but this name seems bit confusing. I used `Option<Done>` because it's clear that it will continue execution if `done` is `None`.

kdy1: I misread the comment. Please ignore the comment about `Option<Done>`.

But I think `Done::Pass` is confusing anyway. 
How about `Fallthrough`? It seems like a commonly used alternative.

src/control_flow/mod.rs
```diff
@@ -252,6 +287,8 @@ impl Visit for Analyzer<'_> {
 
     if let Some(done) = self.scope.done {
       self.mark_as_done(s.span.lo, done);
+    } else {
+      self.mark_as_done(s.span.lo, Done::Pass);
```
kdy1, line 291: Maybe you can rename

 - `Done` => `End`
 - `Pass` => `Continue`
 - `mark_as_done` => `mark_as_end`

I'm not sure which one is better.
cc @bartlomieju What do you think about this?

src/control_flow/mod.rs
```diff
@@ -351,14 +389,17 @@ impl Visit for Analyzer<'_> {
 
       if a.scope.found_break.is_some() {
         case_done = Some(Done::Break);
-      } else if let Some(Done::Forced) = a.scope.done {
+      } else if a.scope.done == Some(Done::Forced) {
```
kdy1, line 392: I'm curious what does this change do.

src/control_flow/mod.rs
```diff
@@ -428,9 +475,7 @@ impl Visit for Analyzer<'_> {
       false
     };
 
-    if unreachable {
-      self.info.entry(n.span().lo).or_default().unreachable = true;
-    }
+    self.info.entry(n.span().lo).or_default().unreachable = unreachable;
```
kdy1, line 478: This will allocate an entry in hashmap even if it's reachable. Is this intended?

(I'm not talking about performance as it is a linter)

kdy1 (commented):
Nice work!

src/control_flow/mod.rs
kdy1: Maybe 

 - `Done` => `End`
 - `Pass` => `Continue`
 - `mark_as_done` => `mark_as_end`

would be better

src/control_flow/mod.rs
bartlomieju: Suggested names seem more descriptive and self-explanatory to me 👍

magurotuna: > Done => End
> Pass => Continue
> mark_as_done => mark_as_end

Looks nice to me, too. I'll rename them.

src/control_flow/mod.rs
kdy1: Ah. I see. Seems reasonable to me.

src/control_flow/mod.rs
magurotuna: Yes, it's intended. Without this change, there would be some statements that are not stored in the hashmap, which is not consistent with the comment appended to the `meta` method (the comment says that all statements can be queried).

https://github.com/denoland/deno_lint/blob/19bc382db7572d9b011f37eb90f1daa7df9279e0/src/control_flow/mod.rs#L26-L32

src/control_flow/mod.rs
```diff
@@ -569,3 +650,543 @@ impl Visit for Analyzer<'_> {
...
+"#;
+    let flow = analyze_flow(src);
+    assert_flow!(flow, 1, false, Some(End::Continue)); // switch stmt
+    assert_flow!(flow, 18, false, Some(End::Break)); // `case 1`
+    assert_flow!(flow, 30, false, Some(End::Break)); // break stmt
+    assert_flow!(flow, 39, false, Some(End::Forced)); // `default`
+    assert_flow!(flow, 48, false, Some(End::Forced)); // BlockStmt of `default`
+    assert_flow!(flow, 54, false, Some(End::Forced)); // return stmt
+    assert_flow!(flow, 70, false, Some(End::Forced)); // throw stmt
+  }
```
bartlomieju, line 1191: Great tests!

bartlomieju (approved):
LGTM, thank you @magurotuna, great upgrade!

PR #383, opened 2020-10-11
refactor: remove Arc/Rc

magurotuna:
- `leading_comments`
- `trailing_comments`
- `ignore_directives`
- `scope`
- `control_flow`

The above fields of `Context` struct are declared with `Arc` or `Rc` now, but it seems unnecessary so I removed them.
If `Arc` and `Rc` are removed, it will cost us a lot to call `.clone`. To address it, I refactored some implementations so that unnecessary cloning won't happen.

### DIFF
--- a/src/linter.rs
+++ b/src/linter.rs
@@ -23,17 +23,15 @@ lazy_static! {
     regex::Regex::new(r",\s*|\s").unwrap();
 }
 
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):
LGTM, great PR 🚀

PR #389, opened 2020-10-13
 fix(no-cond-assign): handle nested stuff

magurotuna:
This is a part of #330

Allows `no-cond-assign` to be able to handle nested `if`, `while`, `do-while`, `for`, and conditional operators.


bartlomieju (approved):
LGTM, thanks @magurotuna

Issue #386, opened 2020-10-12
Refactor diagnostic formatting in "dlint" example


bartlomieju:
Currently "dlint" example uses diagnostic formatting similar to that of Rust.

This logic is not super robust and it turns out that there's a crate that handles this logic (used in rustc itself) - https://docs.rs/annotate-snippets/0.9.0/annotate_snippets/snippet/index.html

We should try to switch out to using this crate.

CC @lucacasonato

PR #391, opened 2020-10-13
example(dlint): use annotate-snippets for diagnostic formatting

bartlomieju:
This commit updates "dlint" example to use "annotate-snippets" crate
for diagnostic formatting. It allows to remove custom formatting implementation
(which was supposed to look the same) as well as colors implementation.

Closes https://github.com/denoland/deno_lint/issues/386

```
error[no-explicit-any]: `any` type is not allowed
   --> ./benchmarks/oak/router.ts:6:27
    |
  1 | /**
  2 |  * Adapted directly from @koa/router at
  3 |  * https://github.com/koajs/router/ which is licensed as:
  4 |  *
...
899 |
900 |     return this as Router<any, any>;
    |                           ^^^
    |
error[no-explicit-any]: `any` type is not allowed
   --> ./benchmarks/oak/router.ts:6:32
    |
  1 | /**
  2 |  * Adapted directly from @koa/router at
  3 |  * https://github.com/koajs/router/ which is licensed as:
  4 |  *
...
899 |
900 |     return this as Router<any, any>;
    |                                ^^^
    |
Found 36 problems
```

Screenshot 2020-10-13 at 20 30 17

Current problems:
- first 4 lines of source file are displayed for each error
- line number next to file is wrong - not sure what's the problem here - it's computed from `SourceAnnotation` which properly underlines offending token

Given that a lot of lines have been removed I think we should pursue this solution.

### DIFF
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -1,4 +1,6 @@
 // Copyright 2020 the Deno authors. All rights reserved. MIT license.
+use annotate_snippets::display_list;
+use annotate_snippets::snippet;
 use clap::App;
 use clap::AppSettings;
 use clap::Arg;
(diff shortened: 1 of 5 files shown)


bartlomieju:
Screenshot 2020-10-15 at 13 31 52

examples/dlint/main.rs
```diff
@@ -35,7 +36,41 @@ fn create_cli_app<'a, 'b>() -> App<'a, 'b> {
     )
 }
 
-fn format_diagnostic(diagnostic: &LintDiagnostic, source: &str) -> String {
+fn get_slice_source_and_range<'a>(
+  source: &'a str,
+  range: &Range,
+) -> (&'a str, (usize, usize)) {
+  let line_start_indexes = std::iter::once(0)
+    .chain(source.match_indices('\n').map(|l| l.0 + 1))
```
RDambrosio016: Just a heads up, this is incorrect because ECMAScript defines `\n`, `\r`, `\u{2028}`, and `\u{2029}` all as valid linebreak characters (although we havent fixed it in rslint yet either 😄)

bartlomieju: @RDambrosio016 - thanks for pointing this out. I'll open an issue about it. All in all `annotate-snippets` is pretty cool, although I found the need to manually take slices of source code and adjusting ranges a bit cumbersome... Otherwise it results in output as seen in https://github.com/denoland/deno_lint/pull/391#issue-502664928

RDambrosio016: Thats kind of odd, it sounds like the ranges are a bit broken, no?

bartlomieju: It does sound like it - I checked `annotate-snippets` issues but couldn't find anything about it, I guess that's just how the crate works ¯\_(ツ)_/¯ anyway, taking a slice and adjusting range is no big deal so that's fine.

Stupremee:
I fixed both of your current problems in our local copy of `annotate-snippet` in `rslint_errors`, so you don't have these errors if you switch to `rslint_errors`.

bartlomieju:
@Stupremee which problems are you referring to? I actually managed to fix these problems:
```
- first 4 lines of source file are displayed for each error
- line number next to file is wrong - not sure what's the problem here - it's computed from SourceAnnotation which properly underlines offending token
```
That was done by adding little utility function. That said I will look into `rslint_errors` soon.

Stupremee:
Yea I fixed these two problems directly in annotate-snippets.

PR #392, opened 2020-10-14
refactor: switch to `impl Into<String>` to avoid unnecessary allocation

magurotuna:
Currently `Context::add_diagnostic` method receives code and message with `&str`.
It results in unnecessary allocation because the callers of `add_diagnostic` often pass a message like `&format!("foo {}", bar)`. 
It'd be better to pass the built `String` value (not reference) to`add_diagnostic` for efficiency.

So, I switched their types to `impl Into<String>`, which will allow us to build and allocate new `String` only if that's required.

### DIFF
--- a/src/linter.rs
+++ b/src/linter.rs
@@ -39,8 +39,8 @@ impl Context {
   pub(crate) fn add_diagnostic(
     &mut self,
     span: Span,
-    code: &str,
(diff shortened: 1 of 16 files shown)


bartlomieju (approved):
LGTM

PR #394, opened 2020-10-15
fix(no-empty): handle nested empty blocks in switch

magurotuna:
This is a part of #330 

Allows `no-empty` to be able to handle nested empty blocks in switch's discriminant, and did small refactoring.


bartlomieju (approved):
Nice fix, thanks @magurotuna

PR #396, opened 2020-10-15
doc: Add markup docs for rules

cknight:
Dipping my toe into #159, this add docs for `adjacent-overload-signatures` and `ban-ts-comment`.  

As an aside, `ban-ts-comment` as a name is confusing.  This lint rule does not ban any TS directives (as is the default intent of the eslint equivalent), but instead only allows them with comments/descriptions (which is confusing considering the name).  I've updated the rule diagnostic text to reflect this, but if we aren't tied to eslint's names something like `no-ts-directive-without-comment` or some such might be more descriptive.

### DIFF
--- a/src/rules/adjacent_overload_signatures.rs
+++ b/src/rules/adjacent_overload_signatures.rs
@@ -31,6 +31,78 @@ impl LintRule for AdjacentOverloadSignatures {
     let mut visitor = AdjacentOverloadSignaturesVisitor::new(context);
     visitor.visit_module(module, module);
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Requires overload signatures to be adjacent to each other.
+
+Overloaded signatures which are not next to each other can lead to code which is hard to read and maintain.
+
+### Valid:
+(bar is declared after foo)
+```typescript
+type FooType = {
+  foo(s: string): void;
+  foo(n: number): void;
+  foo(sn: string | number): void;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, thanks @cknight!

>This lint rule does not ban any TS directives (as is the default intent of the eslint equivalent), but instead only allows them with comments/descriptions (which is confusing considering the name).

Then we should probably better align our implementation to ESlint's - I'm open to suggestions.

cknight:
@bartlomieju It is more or less aligned, just opinionated.  Vanilla `ban-ts-comment` out of the box in eslint will ban the `ts-ignore` and `ts-nocheck` directives all-together, while requiring a comment for `ts-expect-error`.  Users can further configure this if they wish.  Our implementation, unlike the eslint default, allows `@ts-ignore` and `@ts-nocheck` as long as they have comments, where eslint would ban them all-together.  eslint also enforces a minimum length of 3 for any comments.  Our code doesn't appear to do this.

p.s. I see there's a desire to add `hint` as well, now present in the LintDiagnostic struct (#162).  Bearing in mind I don't know any Rust, how would I go about adding this (assuming adding the hint text is still a task)?

bartlomieju:
@cknight I think we could just add hint as you suggested - the field on `LintDiagnostic` is there, it's just not exposed in the API. Let me add it real quick in a new PR.

PR #398, opened 2020-10-16
doc: add hints

cknight:
A small PR just to check this is what is expected.


bartlomieju (approved):
LGTM, thanks @cknight, I'll update example later to show those hints.

PR #399, opened 2020-10-16
refactor: Remove ban-ts-ignore

cknight:
The curernt deno lint rules include both `ban-ts-comment` and `ban-ts-ignore`.  The former controls behaviour of most TS directives (ts-expect-error, ts-nocheck and ts-ignore), while the latter specifically covers just ts-ignore.  `ban-ts-ignore` was removed from typescript-eslint in May as `ban-ts-comment` covered the same functionality but for all directives (include ts-ignore).  See https://github.com/typescript-eslint/typescript-eslint/releases/tag/v3.0.0 for more details.


magurotuna (approved):
good catch!

bartlomieju (approved):
LGTM, thanks @cknight, indeed this rule is superfluous

PR #400, opened 2020-10-16
test: add `assert_lint_err` macro to assert messages and hints 

### LINKED ISSUE #395: error messages and hints should be asserted in tests
Current tests for the rules are checking if error spans and the number of errors match what we expect. I'd like error messages and hints to be asserted as well, because there are some rules that have more than one possible messages.

For example, `getter-return` has two possible error messages in [ESLint](https://github.com/eslint/eslint/blob/v7.11.0/lib/rules/getter-return.js#L58-L61) (and `deno_lint` imitates it):

1. `Expected to return a value in {{name}}.`
2. `Expected {{name}} to always return a value.`

So, the design of new assertion I'm thinking now looks like:

```rust
assert_lint_err!(
  GetterReturn,
  [
    {
      code: "const foo = { get getter() {} };",
      col: 14,
      message: "Expected to return a value in getter.",
    },
    {
      code: r#"
class Foo {
  get foo() {
    class Bar {
      get bar() {}
    };
    return 42;
  }
}
      "#,
      line: 5,
      col: 6,
      message: "Expected to return a value in bar.",
      hint: "This is a hint",
    },
  ],
);
```

- `code`, `col` and `message` are required
- `line` can be omitted if an error is at line 1
- `hint` can be omitted if there's no hint

@bartlomieju Any thoughts on this?

magurotuna:
Closes #395 

There's a difference in the interface of macro from what I proposed in #395, because I have to deal with multiple errors against one source code.

```rust
assert_lint_err_macro! {
  NoUndef,
  "({b: a} = {});": [
    {
      col: 5,
      message: "a is not defined",
    },
  ],
  "[obj.a, obj.b] = [0, 1];": [
    {
      col: 1,
      message: "obj is not defined",
    },
    {
      col: 8,
      message: "obj is not defined",
    },
  ],
  "const c = 0; const a = {...b, c};": [
    {
      col: 27,
      message: "b is not defined",
    },
  ],
};
```

In addition to error macro, I also added `assert_lint_ok` macro.
After this PR gets merged, I'll replace `assert_lint_err` and `assert_lint_ok` function with the macro in tests other than `no_unref.rs`.

### DIFF
--- a/src/diagnostic.rs
+++ b/src/diagnostic.rs
@@ -1,12 +1,28 @@
 // Copyright 2020 the Deno authors. All rights reserved. MIT license.
 #[cfg(feature = "json")]
 use serde::Serialize;
+use std::fmt;
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Copy)]
 #[cfg_attr(feature = "json", derive(Serialize))]
 pub struct Position {
   pub line: usize,
   pub col: usize,
+}
+
+impl fmt::Display for Position {
+  fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+    write!(f, "(line: {}, col: {})", self.line, self.col)
+  }
+}
+
+impl Into<Position> for (usize, usize) {
+  fn into(self) -> Position {
+    Position {
+      line: self.0,
+      col: self.1,
+    }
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
LGTM @magurotuna, nice utility which should make testing better

PR #402, opened 2020-10-16
doc: Additional docs for a few rules

cknight:
This PR:
* Adds clarity in comments to the differences in `ban-ts-comment` between typescript-eslint and deno lint
* Adds documentation to `ban-types`, `ban-untagged-ignore` and `ban-untagged-todo`

### DIFF
--- a/src/rules/ban_ts_comment.rs
+++ b/src/rules/ban_ts_comment.rs
@@ -8,6 +8,18 @@ use swc_common::Span;
 
 pub struct BanTsComment;
 
+/*
+ * This rule differs from typescript-eslint.  In typescript-eslint the following
+ * defaults apply:
+ * ts-expect-error: allowed with comment
+ * ts-ignore: not allowed
+ * ts-nocheck: not allowed
+ *
+ * This rules defaults:
+ * ts-expect-error: allowed with comment
+ * ts-ignore: allowed with comment
+ * ts-nocheck: allowed with comment
+ */
 impl BanTsComment {
   fn report(&self, context: &mut Context, span: Span) {
     context.add_diagnostic_with_hint(
--- a/src/rules/ban_ts_comment.rs
+++ b/src/rules/ban_ts_comment.rs
@@ -13,7 +13,7 @@ impl BanTsComment {
     context.add_diagnostic(
       span,
       "ban-ts-comment",
-      "ts directives are not allowed",
+      "ts directives are not allowed without comment",
     );
   }
 }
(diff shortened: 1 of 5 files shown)


src/rules/ban_untagged_todo.rs
```diff
@@ -55,6 +56,28 @@ impl LintRule for BanUntaggedTodo {
       self.report(context, span);
     }
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Requires TODOs to be annotated with either a user tag (@user) or an issue reference (#issue).
+
+TODOs without reference to a user or an issue become stale with no easy way to get more information
```
magurotuna, line 63:
```suggestion
TODOs without reference to a user or an issue become stale with no easy way to get more information.
```

nitpick :smiley:

src/rules/ban_untagged_ignore.rs
```diff
@@ -37,13 +37,32 @@ impl LintRule for BanUntaggedIgnore {
...
+        "Ignore directive requires lint rule name(s)",
+        "Add one or more lint rule names.  E.g. // deno-lint-ignore adjacent-overload-signatures",
       )
     }
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Requires `deno-lint-ignore` to be annotated with one or more rule names.
+
+Ignoring all rules can mask unexpected or future problems.  Therefore you need to explicitly specify which rule(s) are to be ignored.
```
magurotuna, line 52:
```suggestion
Ignoring all rules can mask unexpected or future problems. Therefore you need to explicitly specify which rule(s) are to be ignored.
```

redundant space :eyes:

src/rules/ban_ts_comment.rs
```diff
@@ -8,6 +8,18 @@ use swc_common::Span;
...
+ * ts-expect-error: allowed with comment
+ * ts-ignore: not allowed
+ * ts-nocheck: not allowed
+ *
+ * This rules defaults:
+ * ts-expect-error: allowed with comment
+ * ts-ignore: allowed with comment
+ * ts-nocheck: allowed with comment
+ */
 impl BanTsComment {
```
magurotuna, line 23:
```suggestion
/// This rule differs from typescript-eslint. In typescript-eslint the following
/// defaults apply:
/// - ts-expect-error: allowed with comment
/// - ts-ignore: not allowed
/// - ts-nocheck: not allowed
///
/// This rules defaults:
/// - ts-expect-error: allowed with comment
/// - ts-ignore: allowed with comment
/// - ts-nocheck: allowed with comment
pub struct BanTsComment;

impl BanTsComment {
```

If we append the comment on the `BanTsComment` struct in three slashes style, we can get documentation like this. I think it would be a little more useful for other developers 👍 

image

src/rules/ban_types.rs
```diff
@@ -31,6 +31,44 @@ impl LintRule for BanTypes {
     let mut visitor = BanTypesVisitor::new(context);
     visitor.visit_module(module, module);
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Bans the use of primitive wrapper objects (e.g. `String` the object is a 
+      wrapper of `string` the primitive) in addition to the non-explicit `Function`
+      type and the misunderstood `Object` type. 
+
+There are very few situations where primitive wrapper objects are desired and
```
magurotuna, line 40:
```suggestion
  fn docs(&self) -> &'static str {
    r#"Bans the use of primitive wrapper objects (e.g. `String` the object is a 
wrapper of `string` the primitive) in addition to the non-explicit `Function`
type and the misunderstood `Object` type. 

There are very few situations where primitive wrapper objects are desired and
```

I'm not sure, but I feel like leading spaces would result in being rendered a little badly.

src/rules/ban_ts_comment.rs
cknight: Thanks for the tip!

src/rules/ban_types.rs
cknight: Markdown should automatically handle extra spaces gracefully (by stripping them).  Regardless, your suggestion is neater and my extra spaces unintentional.

magurotuna (approved):
LGTM 👍

bartlomieju (approved):
Nice, thank you @cknight

PR #404, opened 2020-10-17
fix(no-constant-condition): handle nested stuff

magurotuna:
This is a part of #330 

What I did in this PR:

- allowing it to handle nested `IfStmt` and `CondExpr`
- removing `swc_ecmascript::ast::` by importing on the top of `no_constant_condition.rs`


bartlomieju (approved):
LGTM

PR #406, opened 2020-10-17
fix(no-dupe-keys): handle nested objects & getter and setter with the same name

magurotuna:
This is a part of #330 

What I did In this PR:

- adding more tests (mostly from ESLint)
- allowing it to handle nested objects e.g. `const obj = { a: { b: 1, b: 2 } };`
- adding proper handling for getter and setter with the same name e.g. `const obj = { get a() {}, set a() {} };`

### DIFF
--- a/src/rules/no_dupe_keys.rs
+++ b/src/rules/no_dupe_keys.rs
@@ -153,29 +153,33 @@ mod tests {
       r#"var foo = { bar: "baz", set bar() {} };"#,
       10,
     );
-    assert_lint_err::<NoDupeKeys>(r#"var x = { a: b, ['a']: b };"#, 0);
-    assert_lint_err::<NoDupeKeys>(r#"var x = { '': 1, '': 2 };"#, 0);
-    assert_lint_err::<NoDupeKeys>(r#"var x = { '': 1, [``]: 2 };"#, 0);
-    assert_lint_err::<NoDupeKeys>(r#"var x = { 012: 1, 10: 2 };"#, 0);
-    assert_lint_err::<NoDupeKeys>(r#"var x = { 0b1: 1, 1: 2 };"#, 0);
-    assert_lint_err::<NoDupeKeys>(r#"var x = { 0o1: 1, 1: 2 };"#, 0);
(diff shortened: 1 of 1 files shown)


magurotuna:
@bartlomieju ready for review. PTAL

src/rules/no_dupe_keys.rs
```diff
@@ -102,60 +212,111 @@ mod tests {
...
+    assert_lint_err::<NoDupeKeys>(r#"var x = { a: b, ['a']: b };"#, 8);
+    assert_lint_err::<NoDupeKeys>(r#"var x = { '': 1, '': 2 };"#, 8);
+    assert_lint_err::<NoDupeKeys>(r#"var x = { '': 1, [``]: 2 };"#, 8);
+    assert_lint_err::<NoDupeKeys>(r#"var x = { 012: 1, 10: 2 };"#, 8);
+    assert_lint_err::<NoDupeKeys>(r#"var x = { 0b1: 1, 1: 2 };"#, 8);
+    assert_lint_err::<NoDupeKeys>(r#"var x = { 0o1: 1, 1: 2 };"#, 8);
+    // TODO(magurotuna): this leads to panic due to swc error
+    // It seems like tsc v4.0.2 cannot handle this either
+    // playground: https://www.typescriptlang.org/play?target=99&ts=4.0.2#code/MYewdgzgLgBCBGArGBeGBvAUDGBGMAXDACwBMANJgL4DcQA
+    // assert_lint_err::<NoDupeKeys>(r#"var x = { 1n: 1, 1: 2 };"#, 8);
```
bartlomieju, line 286: @magurotuna could you report that in SWC repo?

bartlomieju (approved):
LGTM, thanks for the fix @magurotuna

src/rules/no_dupe_keys.rs
magurotuna: @bartlomieju I thought I would do so, but as I commented, the TypeScript compiler can't parse this either. So I felt like it's okay for now. But I'll report it anyway 😄

magurotuna: https://github.com/swc-project/swc/issues/1173

Filed an issue about it

PR #409, opened 2020-10-17
fix: constructor-super, doc: add lint docs

### LINKED ISSUE #408: Poor UX/incorrect handling of constructor-super rule
Given the following code:
```
class B extends null {
    constructor() { super(); }
}
```
`constructor-super` will flag with `Constructors of classes which inherit from a non constructor must not call super()`.  This seems non-nonsensical to me as the real issue is that constructors aren't allowed here, rather than anything to do with super().

Taking out the `super()` will then yield a second flag of this rule with `Classes which inherit from a non constructor must not define a constructor`.

In both situations (constructor with and without call to super()) I would expect the rule to just flag with `Classes which inherit from a non constructor must not define a constructor`.  In the current setup you end up removing the `super()` only to be hit with the second flag of this issue to further remove the constructor, which is confusing.

cknight:
This PR:
* Resolves #408 
* Contributes to #159 with docs and hints for `constructor-super`, `default-param-last` and `eqeqeq`

### DIFF
--- a/src/rules/constructor_super.rs
+++ b/src/rules/constructor_super.rs
@@ -34,10 +34,10 @@ impl LintRule for ConstructorSuper {
   }
 
   fn docs(&self) -> &'static str {
-    r#"Verifies the correct usage of constructors and calls to super().
+    r#"Verifies the correct usage of constructors and calls to `super()`.
 
-Defined constructors of derived classes (e.g. class A extends B) must always call
-super().  Classes which extend non-constructors (e.g. class A extends null) must
+Defined constructors of derived classes (e.g. `class A extends B`) must always call
+`super()`.  Classes which extend non-constructors (e.g. `class A extends null`) must
 not have a constructor.
 
 ### Valid:
--- a/src/rules/constructor_super.rs
+++ b/src/rules/constructor_super.rs
@@ -32,6 +32,51 @@ impl LintRule for ConstructorSuper {
     let mut visitor = ConstructorSuperVisitor::new(context);
     visitor.visit_module(module, module);
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Verifies the correct usage of constructors and calls to super().
+
+Defined constructors of derived classes (e.g. class A extends B) must always call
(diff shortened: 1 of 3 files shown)


src/rules/constructor_super.rs
```diff
@@ -32,6 +32,51 @@ impl LintRule for ConstructorSuper {
     let mut visitor = ConstructorSuperVisitor::new(context);
     visitor.visit_module(module, module);
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Verifies the correct usage of constructors and calls to super().
+
+Defined constructors of derived classes (e.g. class A extends B) must always call
+super().  Classes which extend non-constructors (e.g. class A extends null) must
+not have a constructor.
```
magurotuna, line 41:
```suggestion
  fn docs(&self) -> &'static str {
    r#"Verifies the correct usage of constructors and calls to `super()`.

Defined constructors of derived classes (e.g. `class A extends B`) must always call
`super()`.  Classes which extend non-constructors (e.g. `class A extends null`) must
not have a constructor.
```

I think it would be nice to get inline codes to be wrapped in backticks

src/rules/eqeqeq.rs
```diff
@@ -47,9 +69,17 @@ impl<'c> Visit for EqeqeqVisitor<'c> {
       } else {
         "expected '!==' and instead saw '!='."
       };
-      self
-        .context
-        .add_diagnostic(bin_expr.span, "eqeqeq", message)
+      let hint = if bin_expr.op == BinaryOp::EqEq {
+        "Use '==='"
+      } else {
+        "Use '!=="
```
magurotuna, line 75:
```suggestion
        "Use '!=='"
```

I didn't notice it at a first glance since it's a really tiny missing 😃

cknight: Nice spot!

magurotuna (approved):
Great 👍

bartlomieju (approved):
LGTM, awesome @cknight!

PR #414, opened 2020-10-18
doc: add lint docs

cknight:
This PR contributes to #159, with docs on the following rules:
* explicit-function-return-types
* explicit-module-boundary-types
* for-direction
* getter-return


bartlomieju (approved):
LGTM, thank you @cknight

PR #415, opened 2020-10-19
feat: proof of concept for plugin system

littledivy:
Ref #159 

- [x] add deno_core as dev dependency
- [ ] accept some flag (--plugins) that collects path to JS files containing JS visitors
- [x] create JsRuntime and execute visitor code in global scope (this is hacky, but way simpler to implement than using modules)
- [x] obtain JSON AST from Rust
 - [ ] load provided plugin files
-  [ ] for each loaded plugin file run its visitor and collect diagnostics (let's say that we expect plugin files to have default export which is a subclass of Visitor)
 - [x]   return diagnostics to Rust for further processing


src/linter.rs
```diff
@@ -36,7 +36,7 @@ pub struct Context {
 }
 
 impl Context {
-  pub(crate) fn add_diagnostic(
```
littledivy, line 39: This was earlier private to the crate itself - had to make it public for external plugins

examples/dlint/plugin.rs
```diff
@@ -0,0 +1,55 @@
...
+  fn new() -> Box<Self> {
+    Box::new(WarnRawGitImport)
+  }
+
+  fn code(&self) -> &'static str {
+    "no-raw-git-import"
+  }
+
+  fn lint_module(
+    &self,
```
bartlomieju, line 20: It's a regular rule, just outside of `deno_lint` lib

bartlomieju:
@littledivy thanks for opening the PR. My idea for JS plugin support is as follows:
- add `deno_core` as dev dependency
- accept some flag (`--plugins`) that collects path to JS files containing JS visitors
- create `JsRuntime` and execute visitor code in global scope (this is hacky, but way simpler to implement than using modules)
- obtain JSON AST from Rust
- load provided plugin files
- for each loaded plugin file run its visitor and collect diagnostics
- return diagnostics to Rust for further processing

examples/dlint/plugin.rs
littledivy: Yup - the goal is to port than to JS

bartlomieju:
@littledivy I've fixed the panic you were experiencing

littledivy:
Plugin <-> Rust diagnostic works although a lot of things are lacking. Little of them listen below -

* Implement diagnostics formatter in JS (will be used with addDiagnostic)
* Merge inbuilt diagnostics with plugin diagnostics 
* Plugin rule-code? or just the plugin file/url name? 

image

bartlomieju:
@littledivy I've rebased and cleaned your PR a little bit

examples/dlint/test_plugin.js
```diff
@@ -0,0 +1,26 @@
...
+            col: 1,
+            bytePos: 0
+          },
+          end: {
+            line: 1,
+            col: e.span.end,
+            bytePos: 0,
+          }
+        }
+      });
```
bartlomieju, line 19: I suggest this method calls into Rust, it will be easier to get proper location there

magurotuna:
Now, if we have `foo.ts` with the following content:

```typescript
import { foo } from "bar";

let forbiddenIdentifier = 42;
const hello;
```

and running `cargo run --example dlint -- run foo.ts --plugin ./examples/dlint/plugins/test_plugin1.js ./examples/dlint/plugins/test_plugin2.js`, then we can see the result:

image

magurotuna:
Next, I'm going to add proper handling for filtering diagnostics. That said, if a file which `dlint` runs on has some ignore directives like:

```typescript
import { foo } from "bar";

// deno-lint-ignore forbidden-ident-name
let forbiddenIdentifier = 42;
const hello;
```

then we should make `dlint` treat the identifier `forbiddenIdentifier` as allowed.

bartlomieju:
> Next, I'm going to add proper handling for filtering diagnostics. That said, if a file which `dlint` runs on has some ignore directives like:
> 
> ```ts
> import { foo } from "bar";
> 
> // deno-lint-ignore forbidden-ident-name
> let forbiddenIdentifier = 42;
> const hello;
> ```
> 
> then we should make `dlint` treat the identifier `forbiddenIdentifier` as allowed.

@magurotuna this is done as a last step of filtering diagnotics:
https://github.com/denoland/deno_lint/blob/4cb2659c8e81a225cc0673d97d91953d8c6b0a1b/src/linter.rs#L263-L288

If you somehow push diagnostics from JS plugins before filtering them it all should still work

magurotuna:
@bartlomieju 
Thanks for the info. Then, it looks like all I have to do is to make diagnostics from plugins added to the `Context` that ordinary rules are using. I'll give it a try that way.

magurotuna:
```typescript
import { foo } from "bar";

// deno-lint-ignore @deno-lint-plugin/forbidden-ident-name
let forbiddenIdentifier = 42;
const hello;
```

Running `cargo run --example dlint -- run foo.ts --plugin ./examples/dlint/plugins/test_plugin1.js ./examples/dlint/plugins/test_plugin2.js` on the above script, the output result is:

image

So now, filtering diagnostics from plugins is working!

examples/dlint/plugins/test_plugin1.js
```diff
@@ -0,0 +1,13 @@
+export default class Plugin extends Visitor {
```
bartlomieju, line 1: Suggestion: add another class called `LintRule` that has static `ruleCode` method. Then add `registerRule` global method that allows to register multiple rules from a single plugin, eg.
```
registerRule(SomeRule.ruleCode(), SomeRule);
registerRule(SomeRule2.ruleCode(), SomeRule2);
```

src/linter.rs
```diff
@@ -23,9 +23,11 @@ use swc_ecmascript::parser::Syntax;
 
 pub use swc_common::SourceFile;
 
+#[derive(Default)]
```
bartlomieju, line 26: Why is `Default` needed?

examples/dlint/js.rs
```diff
@@ -0,0 +1,249 @@
...
+      .unwrap_or_else(HashSet::new);
+
+    diagnostics.into_iter().for_each(|d| {
+      if let Some(hint) = d.hint {
+        context.add_diagnostic_with_hint(d.span, d.code, d.message, hint);
+      } else {
+        context.add_diagnostic(d.span, d.code, d.message);
+      }
+    });
+    context.set_plugin_codes(codes);
```
bartlomieju, line 222: I don't like this API - we should get list of all registered rules before running them, take a look at my suggestion for `registerRule` API

examples/dlint/js.rs
```diff
@@ -0,0 +1,249 @@
...
+      "Deno.core.jsonOpSync('add_code', { code: plugin.ruleCode() });\n";
+  }
+
+  dummy_source
+}
+
+impl Plugins for JsRuleRunner {
+  fn run(&mut self, context: &mut Context, program: Program) {
+    self.runtime.register_op(
+      "get_program",
```
bartlomieju, line 177: I know I added this API, but it's debatable, maybe instead we could call global function like `runPlugins()` that would take JSON serialized AST as an argument?

src/scopes.rs
```diff
@@ -15,7 +15,7 @@ use swc_ecmascript::visit::Node;
 use swc_ecmascript::visit::Visit;
 use swc_ecmascript::visit::VisitWith;
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
bartlomieju, line 18: Why `Default`?

examples/dlint/js.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    let mut runner = Self {
+      runtime: JsRuntime::new(RuntimeOptions {
+        module_loader: Some(Rc::new(FsModuleLoader)),
+        ..Default::default()
+      }),
+      dummy_source: create_dummy_source(plugin_paths),
+    };
+    runner.init();
```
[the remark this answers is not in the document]
magurotuna: Removed `init` method.

examples/dlint/js.rs
```diff
@@ -0,0 +1,249 @@
...
+          let code_from_js: Code = serde_json::from_value(args).unwrap();
+          let mut stored =
+            state.try_take::<Codes>().unwrap_or_else(HashSet::new);
+          // TODO(magurotuna): To differenciate builtin and plugin rules, adds prefix to plugins.
+          // This should be discussed further before it's stabilized.
+          stored.insert(format!("@deno-lint-plugin/{}", code_from_js.code));
+          state.put::<Codes>(stored);
+
+          Ok(serde_json::json!({}))
+        },
```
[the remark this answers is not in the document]
magurotuna: Done

examples/dlint/js.rs
```diff
@@ -0,0 +1,249 @@
...
+            .iter_mut()
+            .for_each(|d| d.code = format!("@deno-lint-plugin/{}", d.code));
+
+          let mut stored =
+            state.try_take::<Diagnostics>().unwrap_or_else(Vec::new);
+          stored.extend(diagnostics_from_js);
+          state.put::<Diagnostics>(stored);
+
+          Ok(serde_json::json!({}))
+        },
```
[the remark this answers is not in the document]
magurotuna: Done

src/linter.rs
magurotuna: This `default` is for creating `Context` like this:

```rust
    let mut context = Context {
      file_name,
      source_map: self.ast_parser.source_map.clone(),
      leading_comments: leading,
      trailing_comments: trailing,
      ignore_directives: RefCell::new(ignore_directives),
      scope,
      control_flow,
      top_level_ctxt,
      ..Default::default() // <==============
    };
```

But it is not necessarily needed if `plugin_codes` field is not necessary in `Context`. So I'll deal with it after the other review comments are addressed.

examples/dlint/plugins/test_plugin1.js
magurotuna: I might not understand your suggestion so I'd like to clarify it.

What you mean is something like as follows?

User's code:

```javascript
// some-rule.js
export default class SomeRule extends Visitor {
  static ruleCode() {
    return "some-rule";
  }

  // rule implementation...
}

// another-rule.js
export default class AnotherRule extends Visitor {
  static ruleCode() {
    return "another-rule";
  }

  // rule implementation...
}
```

then dlint executes the following JS:

```javascript
import Rule1 from "some-rule.js";
import Rule2 from "another-rule.js";

const rules = [];

function registerRule(ruleClass) {
  rules.push({
    code: ruleClass.ruleCode(),
    // https://stackoverflow.com/questions/10314338/get-name-of-object-or-class
    name: ruleClass.constructor.name,
  });
}

function runPlugins(programAst) {
  for (const rule of rules) {
    const result = new rule().collectDiagnostics(programAst);
    Deno.core.jsonOpSync("add_diagnostics", result);
  }
}

registerRule(Rule1);
registerRule(Rule2);
Deno.core.jsonOpSync("add_code", rules);
```

And then dlint calls `runPlugins` in the above JS with JSON serialized AST passed.

-----------

I have no idea if it's doable to call a JS function from Rust, but your suggestion sounds like it's doable.
I'm not familiar with deno_core so I got a little confused.

bartlomieju: @magurotuna yes! That looks very much what I had in mind

bartlomieju:
```
rules.push({
    code: ruleClass.ruleCode(),
    // https://stackoverflow.com/questions/10314338/get-name-of-object-or-class
    name: ruleClass.constructor.name,
  });
```
^-- I don't think `name` is necessary here - you just want to store rule code and its constructor

magurotuna: Thanks for the clarification!

One more question:

> That way diagnostic could have codes written as `pluginName/ruleCode`

What do you intend by `pluginName` here? The name of class name or other things?

bartlomieju: > What do you intend by pluginName here? The name of class name or other things?

Not clear right now, I guess plugin (collection of files) could register itself with any name

magurotuna: Got it. Then I'll leave the current style (`@deno-lint-plugin/some-rule-code`) as it is for now. We'll discuss how to handle a collection of files as a plugin later.

The idea that has come to me was as follows.

User's each plugin file:

```javascript
// some-rule.js (not default export)
export class SomeRule extends Visitor {
  static ruleCode() {
    return "some-rule";
  }

  // rule implementation...
}

// another-rule.js (not default export)
export class AnotherRule extends Visitor {
  static ruleCode() {
    return "another-rule";
  }

  // rule implementation...
}
```

User's plugin root file:

```javascript
// my-plugin.js
// plugin's root file name is used as part of rule code in diagnostics.
// For instance, if `AnotherRule` detects an error, it outputs a diagnostic
// whose code is `my-plugin/another-rule`.

export * from "./some-rule.js";
export * from "./another-rule.js";
```

The script that dlint executes via deno_core:

```javascript
import * as plugin1 from "./my-plugin.js";

const rules = [];

function registerRule(ruleClass) {
  rules.push(ruleClass);
  Deno.core.jsonOpSync("add_code", { code: ruleClass.ruleCode()} );
}

function runPlugins(programAst) {
  for (const rule of rules) {
    const result = new rule().collectDiagnostics(programAst);
    Deno.core.jsonOpSync("add_diagnostics", result);
  }
}

for (const rule in plugin1) {
  registerRule(rule);
}
```

magurotuna:
I've done with refactoring.

- creates `registerRule` function in JS
- renames `Plugin` to `LintRule`
- makes `ruleCode` method static
- creates `runPlugins` global function that is called from Rust
- removes `Default `derives

@bartlomieju PTAL when you have time.

examples/dlint/js.rs
```diff
@@ -0,0 +1,267 @@
...
+      let module = deno_core::ModuleSource {
+        code: content,
+        module_url_specified: module_specifier.to_string(),
+        module_url_found: module_specifier.to_string(),
+      };
+      Ok(module)
+    }
+    .boxed_local()
+  }
+}
```
bartlomieju, line 147: @magurotuna can you open a PR in main repo to add this loader to `deno_core`?

magurotuna: @bartlomieju Sure.

src/linter.rs
```diff
@@ -354,24 +369,39 @@ impl Linter {
...
+
+    let d = self.filter_diagnostics(&mut context);
     let end = Instant::now();
     debug!("Linter::lint_module took {:#?}", end - start);
 
     d
   }
 }
+
+pub trait Plugins {
```
bartlomieju, line 401: `s/Plugins/Plugin`

src/linter.rs
```diff
@@ -157,6 +165,11 @@ impl LinterBuilder {
     self.rules = rules;
     self
   }
+
+  pub fn plugins(mut self, plugins: Option<Box<dyn Plugins>>) -> Self {
```
bartlomieju, line 169: Maybe take a vector of plugins instead? `Vec<Box<dyn Plugin>>`

examples/dlint/js.rs
```diff
@@ -0,0 +1,267 @@
...
+
+impl Plugins for JsRuleRunner {
+  // TODO(magurotuna): this method sometimes panics, so maybe should return `Result`?
+  fn run(&mut self, context: &mut Context, program: Program) {
```
bartlomieju, line 151: Yes, it should return `Result`

examples/dlint/js.rs
```diff
@@ -0,0 +1,267 @@
...
+    // TODO(magurotuna): `futures::executor::block_on` doesn't seem ideal, but works for now
```
bartlomieju, line 152: it's fine to call `block_on` here

examples/dlint/js.rs
```diff
@@ -0,0 +1,267 @@
...
+      .execute("visitor.js", include_str!("visitor.js"))
+      .unwrap();
+    runtime.register_op(
+      "add_diagnostics",
+      deno_core::json_op_sync(op_add_diagnostics),
+    );
+    runtime
+      .register_op("add_rule_code", deno_core::json_op_sync(op_add_rule_code));
+
+    // TODO(magurotuna): `futures::executor::block_on` doesn't seem ideal, but works for now
```
bartlomieju, line 99: It's alright

examples/dlint/lint.ts
```diff
@@ -0,0 +1,3 @@
+import * as xD from "xD";
```
littledivy, line 1: This was for initial testing - can remove this now :)

src/linter.rs
magurotuna: done

src/linter.rs
magurotuna: done :)

Cargo.toml
```diff
@@ -25,12 +25,12 @@ swc_ecmascript = { version = "=0.14.1", features = ["parser", "transforms", "uti
 regex = "1.4.1"
 once_cell = "1.5.2"
 derive_more = { version = "0.99.11", features = ["display"] }
+deno_core = "0.67.0"
```
bartlomieju, line 28: Yikes, @magurotuna this is not good, we can't have `deno_core` as a regular dep, it must be `dev-dependency`

examples/dlint/js.rs
magurotuna: Okay, I've added `deno_core` to _dependencies_ (not dev_dependencies) to use `AnyError`.

examples/dlint/lint.ts
magurotuna: thanks, I've removed 👍

src/linter.rs
```diff
@@ -398,10 +400,10 @@ impl Linter {
...
 }
 
-pub trait Plugins {
+pub trait Plugin {
   fn run(
     &mut self,
     context: &mut Context,
     program: swc_ecmascript::ast::Program,
-  );
+  ) -> Result<(), AnyError>;
```
bartlomieju, line 408: Use [`anyhow::Error`](https://docs.rs/anyhow/1.0.32/anyhow/struct.Error.html) instead

Cargo.toml
magurotuna: ok I switched it to anyhow instead.

src/linter.rs
magurotuna: `deno_core` uses anyhow@v1.0.32, while `deno_lint` uses v1.0.34. Is it better to use v1.0.32 in `deno_lint` too?

bartlomieju: @magurotuna sure, sounds good

magurotuna: ah I was mistaken, both `deno_core` and `deno_lint` specify anyhow's version in caret style (by default), and `1.0.32` and `1.0.34` are compatible in semver - so there's no problem with current Cargo.toml

bartlomieju (approved):
LGTM 🚀 

@littledivy @magurotuna thanks for working on this awesome feature! I have a few ideas how it could be improved further. Let's see how we could integrate it into `deno lint` at some point!

PR #428, opened 2020-10-21
doc: put invalid examples first and add new docs

cknight:
This PR:
* Re-orders most examples to put the invalid examples above the valid ones.  This makes more semantic sense ("here's what's wrong followed by here's how it should look"), follows the pattern of eslint and sets a consistent pattern in the deno lint rules.
* Contributes to #159 with documentation now available on the following rules:
  * no-class-assign
  * no-const-assign
  * no-constant-condition


bartlomieju (approved):
Great, thank you @cknight!

PR #430, opened 2020-10-22

cknight:
This PR contributes to #159 adding the following docs:
* no-control-regex
* no-debugger
* no-delete-var
* no-dupe-args
* no-dupe-class-members
* no-dupe-else-if
* no-duplicate-case

### DIFF
--- a/src/rules/no_debugger.rs
+++ b/src/rules/no_debugger.rs
@@ -29,6 +29,31 @@ impl LintRule for NoDebugger {
     let mut visitor = NoDebuggerVisitor::new(context);
     visitor.visit_module(module, module);
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows the use of the `debugger` statement
+
+`debugger` is a statement which is meant for stopping the javascript execution
+environment and start the debugger at the statement.  Modern debuggers and tooling
+no longer need this statement and leaving it in can cause the execution of your
+code to stop in production.
+    
+### Invalid:
+```typescript
+function isLongString(x:string) {
+  debugger;
+  return x.length > 100;
+}
+```
+
+### Valid:
+```typescript
+function isLongString(x:string) {
+  return x.length > 100;  // set breakpoint here instead
+}
+```
+"#
+  }
 }
 struct NoDebuggerVisitor<'c> {
   context: &'c mut Context,
@@ -48,10 +73,11 @@ impl<'c> Visit for NoDebuggerVisitor<'c> {
     debugger_stmt: &DebuggerStmt,
     _parent: &dyn Node,
   ) {
-    self.context.add_diagnostic(
+    self.context.add_diagnostic_with_hint(
       debugger_stmt.span,
       "no-debugger",
       "`debugger` statement is not allowed",
+      "Remove the `debugger` statement",
     );
   }
 }
--- a/src/rules/no_debugger.rs
+++ b/src/rules/no_debugger.rs
@@ -48,7 +48,7 @@ function isLongString(x: string) {
 
 ### Valid:
 ```typescript
-function isLongString(x:string) {
+function isLongString(x: string) {
   return x.length > 100;  // set breakpoint here instead
 }
 ```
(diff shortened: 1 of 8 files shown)


src/rules/no_debugger.rs
```diff
@@ -29,6 +29,31 @@ impl LintRule for NoDebugger {
...
+    r#"Disallows the use of the `debugger` statement
+
+`debugger` is a statement which is meant for stopping the javascript execution
+environment and start the debugger at the statement.  Modern debuggers and tooling
+no longer need this statement and leaving it in can cause the execution of your
+code to stop in production.
+    
+### Invalid:
+```typescript
+function isLongString(x:string) {
```
magurotuna, line 43:
```suggestion
function isLongString(x: string) {
```

It would be nice to make the example formatted :)
(if we run `deno fmt --check`, it shows us:

```
1 | -function isLongString(x:string) {
1 | +function isLongString(x: string) {

error: Found 1 not formatted file in 1 file
```

src/rules/no_debugger.rs
```diff
@@ -29,6 +29,31 @@ impl LintRule for NoDebugger {
...
+  debugger;
+  return x.length > 100;
+}
+```
+
+### Valid:
+```typescript
+function isLongString(x:string) {
```
magurotuna, line 51:
```suggestion
function isLongString(x: string) {
```

ditto

src/rules/no_delete_var.rs
```diff
@@ -31,6 +31,33 @@ impl LintRule for NoDeleteVar {
...
+delete a; // would return false
+delete b; // would return false
+delete c; // would return false
+```
+
+### Valid:
+```typescript
+var obj = {
+  a: 1
+}
```
magurotuna, line 56:
```suggestion
var obj = {
  a: 1,
};
```

`deno fmt` recommends semicolon and trailing comma.

src/rules/no_duplicate_case.rs
```diff
@@ -30,6 +30,44 @@ impl LintRule for NoDuplicateCase {
     let mut visitor = NoDuplicateCaseVisitor::new(context);
     visitor.visit_module(module, module);
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows using the same case clause in a switch statement more than once
+
+When you reuse a case test expression in a `switch` statement, the duplicate case is
+never be reached meaning this is almost always a bug.
```
magurotuna, line 38:
```suggestion
When you reuse a case test expression in a `switch` statement, the duplicate case will
never be reached meaning this is almost always a bug.
```

I guess you wanted to write like this?

cknight:
Thanks for all the suggestions @magurotuna!

src/rules/no_duplicate_case.rs
```diff
@@ -30,6 +30,44 @@ impl LintRule for NoDuplicateCase {
...
+switch (someText) {
+    case "a":
+        break;
+    case "b":
+        break;
+    case "a":  // duplicate test expression
+        break;
+    default:
+        break;
+}
```
magurotuna, line 52:
```suggestion
const someText = "a";
switch (someText) {
  case "a":
    break;
  case "b":
    break;
  case "a": // duplicate test expression
    break;
  default:
    break;
}
```

Unless there's particular reason, using 2 spaces indent would be recommended 😃

src/rules/no_duplicate_case.rs
```diff
@@ -30,6 +30,44 @@ impl LintRule for NoDuplicateCase {
...
+switch (someText) {
+    case "a":
+        break;
+    case "b":
+        break;
+    case "c":
+        break;
+    default:
+        break;
+}
```
magurotuna, line 67:
```suggestion
const someText = "a";
switch (someText) {
  case "a":
    break;
  case "b":
    break;
  case "c":
    break;
  default:
    break;
}
```

ditto

magurotuna (approved):
Great fix & nice work! LGTM

bartlomieju (approved):
LGTM, thanks @cknight!

Thank you for review @magurotuna

PR #433, opened 2020-10-24
refactor: use `once_cell` instead of `lazy_static`

### LINKED ISSUE #432: Replace `lazy_static!` with `once_cell::unsync::Lazy`
Currently we're using `lazy_static` in many places for various purposes e.g. creating `regex`es or `HashMap`s.
It works fine, but the syntax of `lazy_static` looks a little bit strange to me; `static ref` doesn't appear outside of the `lazy_static` macro.
So I'd like to replace `lazy_static` with `once_cell`.  `once_cell` provides almost the same functionality as `lazy_static` with no macro.  `once_cell` actually works more than `lazy_static` in some cases (you can see details at https://docs.rs/once_cell/1.4.1/once_cell/)
Furthermore, [the APIs of `once_cell` is being proposed for the Rust std](https://github.com/rust-lang/rfcs/pull/2788), which is also why I want to replace.

There may be a concern about increasing dependency. In fact, `once_cell` is already included in the dependency graph (`swc_common` is using `once_cell`), so we don't need to be concerned.
The output of `cargo tree` looks like:

```
deno_lint v0.2.5 (/home/yusuktan/Repos/github.com/magurotuna/deno_lint)
├── lazy_static v1.4.0
(snip)
├── swc_common v0.10.4
(snip)
│   ├── log v0.4.11 (*)
│   ├── once_cell v1.4.1  // <--------------------- `once_cell` is already included
│   ├── owning_ref v0.4.1
│   │   └── stable_deref_trait v1.2.0
│   ├── scoped-tls v1.0.0
│   ├── serde v1.0.115 (*)
│   ├── swc_visit v0.2.1
│   │   ├── either v1.6.0
│   │   └── swc_visit_macros v0.2.1
│   │       ├── Inflector v0.11.4
│   │       │   ├── lazy_static v1.4.0
│   │       │   └── regex v1.3.9 (*)
│   │       ├── pmutil v0.5.3 (*)
│   │       ├── proc-macro2 v1.0.21 (*)
│   │       ├── quote v1.0.7 (*)
│   │       ├── swc_macros_common v0.3.1 (*)
│   │       └── syn v1.0.40 (*)
│   └── unicode-width v0.1.8
(snip)
```

And the following examples shows how a code would change:

(Before)

```rust
lazy_static! {
  static ref UNDERSCORE_CHAR_RE: Regex =
    Regex::new(r"([^_])_([a-z])").unwrap();
}
```

(After)

```rust
use once_cell::unsync::Lazy;

static UNDERSCORE_CHAR_RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"([^_])_([a-z])").unwrap());
```

magurotuna:
This PR replaces `lazy_static` with `once_cell`. The motivation for this is described in #432.

There is one difference from the proposal; actually I use not `once_cell::unsync::Lazy` but `once_cell::sync::Lazy` because it didn't compile when I used the unsync version. It looks like a requirement from `swc_common`.

```
error[E0277]: `std::cell::Cell<std::option::Option<fn() -> regex::Regex>>` cannot be shared between threads safely
  --> src/linter.rs:22:1
   |
22 | / static IGNORE_COMMENT_CODE_RE: once_cell::unsync::Lazy<Regex> =
23 | |   once_cell::unsync::Lazy::new(|| Regex::new(r",\s*|\s").unwrap());
   | |___________________________________________________________________^ `std::cell::Cell<std::option::Option<fn() -> regex::Regex>>` cannot be shared between threads safely
   |
   = help: within `swc_common::sync::Lazy<regex::Regex>`, the trait `std::marker::Sync` is not implemented for `std::cell::Cell<std::option::Option<fn() -> regex::Regex>>`
   = note: required because it appears within the type `swc_common::sync::Lazy<regex::Regex>`
   = note: shared static variables must have a type that implements `Sync`
```

Closes #432


bartlomieju (approved):
Cool, LGTM

Issue #431, opened 2020-10-24
Tracking issue for switching to `assert_lint_err!` macro


magurotuna:
This is a tracking issue for switching from `assert_lint_err` functions to `assert_lint_err!` macro.

I introduced the new test utility macro `assert_lint_err!` which allows us to make sure lint error messages and hints are equal to what we expect.
Now we would like to replace the functions with the macro.

The current statuses of the rules:
(✅ means it is included in the recommended rules)

- [x] adjacent-overload-signatures ✅ (#410)
- [x] ban-ts-comment ✅ (#410)
- [x] ban-types ✅ (#435)
- [x] ban-untagged-ignore ✅ (#436)
- [x] ban-untagged-todo (#537)
- [x] camelcase ✅ (#410)
- [x] constructor-super ✅ (#434)
- [x] default-param-last (#684)
- [x] eqeqeq (#684)
- [x] explicit-function-return-type (#684)
- [x] explicit-module-boundary-types (#684)
- [x] for-direction ✅ (#437)
- [x] getter-return ✅ (#446)
- [x] no-array-constructor ✅ (#442)
- [x] no-async-promise-executor ✅ (#442)
- [x] no-await-in-loop (#442)
- [x] no-case-declarations ✅ (#447)
- [x] no-class-assign ✅ (#448)
- [x] no-compare-neg-zero ✅ (#507)
- [x] no-cond-assign ✅ (#508)
- [x] no-const-assign (#684)
- [x] no-constant-condition ✅ (#511)
- [x] no-control-regex ✅ (#512)
- [x] no-debugger ✅ (#515)
- [x] no-delete-var ✅ (#455)
- [x] no-dupe-args ✅ (#516)
- [x] no-dupe-class-members ✅ (#517)
- [x] no-dupe-else-if ✅ (#518)
- [x] no-dupe-keys ✅ (#519)
- [x] no-duplicate-case ✅ (#520)
- [x] no-empty ✅ (#474)
- [x] no-empty-character-class ✅ (#442)
- [x] no-empty-interface ✅ (#526)
- [x] no-empty-pattern ✅ (#442)
- [x] no-eval (#490)
- [x] no-ex-assign ✅ (#442)
- [x] no-explicit-any ✅ (#442)
- [x] no-extra-boolean-cast ✅ (#530)
- [x] no-extra-non-null-assertion ✅ (#531)
- [x] no-extra-semi ✅ (#535)
- [x] no-fallthrough ✅ (#539)
[...]

PR #442, opened 2020-10-25
refactor: use assert_lint_err! macro

bartlomieju:
Refactored to use `assert_lint_err!` in following rules:
- no-empty-character-class
- no-empty-pattern
- no-ex-assign
- no-explicit-any
- no-import-assign
- require-yield
- prefer-namespace-keyword
- no-with
- no-async-promise-executor
- no-array-constructor
- no-await-in-loop

Ref #431


bartlomieju:
@magurotuna PTAL

magurotuna (approved):
LGTM 👍

PR #444, opened 2020-10-25
feat(example): add all-rules subcommand

magurotuna:
Sometimes I feel like having the list of all rules. For example, in case where a check box list is wanted just like #431.
So I added a new subcommand `all-rules` to dlint example. This is how it works:

### pretty format
image

### json format
image

### pretty format (with particular rule name)
image

### json format (with particular rule name)
image

### DIFF
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -8,10 +8,9 @@ use clap::SubCommand;
 use deno_lint::diagnostic::LintDiagnostic;
 use deno_lint::diagnostic::Range;
 use deno_lint::linter::LinterBuilder;
-use deno_lint::rules::get_recommended_rules;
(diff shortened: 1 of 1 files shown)


magurotuna:
Most likely this PR will cause conflict with #441. I will resolve the conflict after #441 is merged 👍

magurotuna:
Update: added ✔️ mark to rules included in the recommended rule set.

image

At the moment the marks are also shown when we run `rules` (not `all-rules`) subcommand:

image

It doesn't make sense, I'm wondering if it should be fixed 🤔

bartlomieju:
@magurotuna that's fine, maybe instead of `all-rules` use `dlint rules --all`?

magurotuna:
@bartlomieju that sounds good. I'll fix

magurotuna:
Done

bartlomieju (approved):

Issue #366, opened 2020-10-06
Support linting scripts


bartlomieju:
Currently `deno_lint` works only on ES modules. 

In `deno` repo all runtime code is written as scripts, it'd be great if we could use `deno_lint` to lint those files as well. 

This will require change to `LintRule` trait - instead of `lint_module()` there should be `lint_program()` method, as well as new method on `Linter` itself to treat loaded file as script instead of module. 

Not sure about integration with `deno lint` subcommand yet. Would be great if we could infer if file is script/module (alternatively we could use some directive to inform linter about it: `// deno-lint script` / `// deno-lint module`) 

CC @lucacasonato

nayeemrmn:
> Not sure about integration with `deno lint` subcommand yet. Would be great if we could infer if file is script/module (alternatively we could use some directive to inform linter about it: `// deno-lint script` / `// deno-lint module`)

Only modules are supported in user space so the subcommand should continue to assume modules. Plus there's no way to infer between scripts and modules until https://github.com/tc39/proposal-modules-pragma passes... but that proposal puts scripts as the default. Not very good for us.

bartlomieju:
@nayeemrmn thanks for pointing this out. Then I think `// deno-lint-pragma script` would be the way to go, we won't have to document it for `deno lint`, but use it only internally in tests.

bartlomieju:
> @nayeemrmn thanks for pointing this out. Then I think `// deno-lint-pragma script` would be the way to go, we won't have to document it for `deno lint`, but use it only internally in tests.

Just realized that this approach might not work - we first need to parse the file to check for such comment...

bartlomieju:
Waiting on https://github.com/swc-project/swc/pull/1189

nayeemrmn:
@bartlomieju We lint `cli/rt` with `dlint` now. I don't think we did anything though... it seems the scripts are valid modules as they are written.

Either way, can this be closed? There's no other motivation to support scripts.

lucacasonato:
Indeed resolved through https://github.com/denoland/deno_lint/pull/484

bartlomieju:
FYI discovering whether a file is a script/module is now handled implicitly by SWC when using `parse_program()` API.

PR #445, opened 2020-10-25
refactor: Initial support for linting scripts

bartlomieju:
Change "LintRule" trait to have "lint_program" instead of "lint_module".

Towards #366

### DIFF
--- a/src/linter.rs
+++ b/src/linter.rs
@@ -16,8 +16,16 @@ use swc_common::comments::SingleThreadedComments;
 use swc_common::BytePos;
 use swc_common::SourceMap;
 use swc_common::Span;
+use swc_common::Spanned;
 use swc_common::{comments::Comment, SyntaxContext};
 use swc_ecmascript::parser::Syntax;
+
+/// Describes if file should be treated as a script
+/// or ES module.
+pub enum FileType {
+  Module,
+  Script,
+}
 
 pub struct Context {
   pub file_name: String,
@@ -192,6 +200,7 @@ impl Linter {
     &mut self,
     file_name: String,
     source_code: String,
+    file_type: FileType,
   ) -> Result<Vec<LintDiagnostic>, SwcDiagnosticBuffer> {
     assert!(
       !self.has_linted,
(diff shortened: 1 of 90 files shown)


bartlomieju:
@magurotuna PTAL

RDambrosio016:
I dont really see why a user should have to care about if their code uses scripts or modules, imo you should assume scripts for .js and modules for .mjs

bartlomieju:
> I dont really see why a user should have to care about if their code uses scripts or modules, imo you should assume scripts for .js and modules for .mjs

@RDambrosio016 this is against how CLI treats files - both `.js` and `.mjs` are treated as modules. And currently this functionality is needed only for linting internal runtime codebase (which are written as scripts).

src/swc_util.rs
```diff
@@ -161,6 +161,47 @@ impl AstParser {
...
+      let mut diagnostic_builder = err.into_diagnostic(&self.handler);
+      diagnostic_builder.emit();
+      SwcDiagnosticBuffer::from_swc_error(buffered_err, self)
+    });
+
+    let parse_result = parse_result.map(|module| {
+      GLOBALS.set(&self.globals, || {
+        module.fold_with(&mut ts_resolver(self.top_level_mark))
+      })
+    });
```
magurotuna, line 200:
```suggestion
    let parse_result = parse_result.map(|script| {
      GLOBALS.set(&self.globals, || {
        script.fold_with(&mut ts_resolver(self.top_level_mark))
      })
    });
```

this would be more descriptive 👍

src/swc_util.rs
```diff
@@ -161,6 +161,47 @@ impl AstParser {
...
+
+    let comments = SingleThreadedComments::default();
+    let lexer = Lexer::new(
+      syntax,
+      JscTarget::Es2019,
+      StringInput::from(&*swc_source_file),
+      Some(&comments),
+    );
+
+    let mut parser = Parser::new_from(lexer);
```
magurotuna, line 188: This snippet is the same as `parse_module`, right? If so, it would be more readable if we extracted it into a common function or something IMO.

src/linter.rs
```diff
@@ -285,32 +313,33 @@ impl Linter {
...
+            )
+          })
+          .collect::<Vec<IgnoreDirective>>();
+
+        if directives.is_empty() {
+          None
+        } else {
+          Some(directives[0].clone())
+        }
+      });
```
magurotuna, line 342:
```suggestion
      comments.with_leading(program.span().lo(), |c| {
        c.iter().find_map(|comment| {
          parse_ignore_comment(
            &self.ignore_file_directives,
            &*self.ast_parser.source_map,
            comment,
            true,
          )
        })
      });
```

Unrelated to the essence of this PR, but we could make it simple like this with better performance 😃

src/swc_util.rs
bartlomieju: Yeah, not sure yet, let's keep it that way for now

magurotuna (approved):
LGTM 👍

PR #446, opened 2020-10-25
refactor(getter-return): switch to `assert_lint_err!` macro

magurotuna:
This is a part of #431

Switches from the `assert_lint_err` functions to the `assert_lint_err!` macro in `getter-return`. Additionally, fixes some more things.


bartlomieju (approved):
LGTM

PR #450, opened 2020-10-25
doc: Add linting hints and docs

cknight:
This PR contributes to #159 by adding the following docs:
* camelcase
* no-empty-character-class
* no-empty-interface

It also adds linting hints and code consistency to no-empty as well.


src/rules/camelcase.rs
```diff
@@ -36,6 +36,36 @@ impl LintRule for Camelcase {
...
+
+  fn docs(&self) -> &'static str {
+    r#"Enforces the use of camelCase in variable names
+
+Consistency in a code base is key for readability and maintainability.  This rule
+enforces variable definitions and assignments to be in camelCase.  Of note:
+* `_` is allowed at the start or end of a variable
+* All uppercase variable names (e.g. constants) may have `_` in their name
+* Function calls are exempt
+* This rule also applies to variables imported via ES6 module imports
```
magurotuna, line 48: It looks like this doesn't fully describe the `camelcase` rule to me.
It's a little hard to give a accurate description, but if I do my best to do it, it would be like...

```suggestion
Consistency in a code base is key for readability and maintainability.  This rule
enforces variable declarations and object property names to be in camelCase if they
are controllable on your own.  Of note:
* `_` is allowed at the start or end of a variable
* All uppercase variable names (e.g. constants) may have `_` in their name
* This rule also applies to variables imported / exported via ES module
```

src/rules/camelcase.rs
```diff
@@ -36,6 +36,36 @@ impl LintRule for Camelcase {
...
+    
+### Invalid:
+```typescript
+let first_name = "Ichigo";
+function do_something(){}
+import { not_camelCased } from "external-module";
```
magurotuna, line 54:
```suggestion
let first_name = "Ichigo";
const obj = { last_name: "Hoshimiya" };
const { last_name } = obj;

function do_something() {}
function foo({ snake_case = "default value" }) {}

class snake_case_class {}

import { not_camelCased } from "external-module.js";
export * as not_camelCased from 'mod.ts';
```

This rule is quite complicated, so it would be nice to provide plenty of invalid examples.

src/rules/camelcase.rs
```diff
@@ -36,6 +36,36 @@ impl LintRule for Camelcase {
...
+
+### Valid:
+```typescript
+let firstName = "Ichigo";
+const FIRST_NAME = "Ichigo";
+const __myPrivateVariable = "Hoshimiya";
+const myPrivateVariable_ = "Hoshimiya";
+function doSomething(){} // function declarations must be camelCase but...
+do_something();  // ...snake_case function calls are allowed
+import { not_camelCased as camelCased } from "external-module";
```
magurotuna, line 65:
```suggestion
let firstName = "Ichigo";
const FIRST_NAME = "Ichigo";
const __myPrivateVariable = "Hoshimiya";
const myPrivateVariable_ = "Hoshimiya";
obj.last_name = "Hoshimiya";
const { last_name: lastName } = obj;

function doSomething() {} // function declarations must be camelCase but...
do_something();  // ...snake_case function calls are allowed
function foo({ snake_case: camelCase = "default value" }) {}

class PascalCaseClass {}

import { not_camelCased as camelCased } from "external-module.js";
export * as camelCased from 'mod.ts';
```

ditto

magurotuna (changes requested):
LGTM except for a few suggestions 👍

src/rules/camelcase.rs
cknight: Thanks for the suggestion.  I think I need more help understanding this rule, as you say, it's really complex!  What do you mean by "if they are controllable on your own"?  Also, was `Function calls are exempt` removed because it was incorrect, too minor a detail, or too simplistic/badly worded?

src/rules/camelcase.rs
cknight: Happy to add more examples.  Would your destructuring example be valid if `obj` came from an external library outside your control?

src/rules/camelcase.rs
cknight: Why is `obj.last_name` valid?

cknight: I'm thinking it's not valid and the intention was to show valid destructuring of an object you don't directly control?  If so, the inclusion of the `obj.last_name` is a little confusing here.  See if my latest commit makes any more sense?

src/rules/camelcase.rs
magurotuna: yeah it's really complex. 
As for "if they are controllable on your own", if I try to put it another way, it would be "if you can rename the variable name to camelCase at that point". 
Consider function call like `do_something();`, in this case we definitely cannot rename `do_something` to camelcase at this point. But say declaring function expression like `const f = function do_something() { };`, here we can rename `do_something` to `doSomething`. This is exactly what I want to describe.

So, `Function calls are exempt` is definitely correct, but it sounds a little too specific. The core of `camelcase` rule is that it only checks declarations, and this also covers the fact that function calls won't be checked.

src/rules/camelcase.rs
magurotuna: > No, because you would need to alias the destructured variable.

exactly 👍

src/rules/camelcase.rs
magurotuna: yeah, ideally we should treat `obj.last_name` as invalid, but the problem here is that it's pretty difficult to check if  `last_name` property is created at this point or it was already created before. 

#### case 1

```ts
const obj: { [key: string]: number } = {};
obj.foo_bar = 42; // <------------- HERE
```

In this case, `foo_bar` property is created at line 2. Ideally `camelcase` rule reports an error of this line.

#### case 2

```ts
const obj = { foo_bar: 42 };
obj.foo_bar = 9999;
```

On the other hand, here `foo_bar` property is created at line 1, and what happens at line 2 is just reassignment of `foo_bar`. In this case `camelcase` rule should not report line 2, because it's impossible to rewrite it to `obj.fooBar = 9999;`.

We might be able to distinguish case1 and case2 if we had some analyzer for that, but we cannot at the moment. Thus we're treating `obj.foo_bar` as valid.

Yet indeed `obj.last_name` looks confusing to users. I'd be happy if you could add some comment to `obj.last_name` example that describes why this is treated as valid.

src/rules/camelcase.rs
```diff
@@ -60,9 +69,16 @@ let firstName = "Ichigo";
 const FIRST_NAME = "Ichigo";
 const __myPrivateVariable = "Hoshimiya";
 const myPrivateVariable_ = "Hoshimiya";
+const { last_name: lastName } = obj;  // where `obj` is defined externally
```
magurotuna, line 72: Whether `obj` is defined externally or internally doesn't change the linter behavior for this line, so I think this comment could be removed 👍

src/rules/camelcase.rs
```diff
@@ -41,17 +41,26 @@ impl LintRule for Camelcase {
     r#"Enforces the use of camelCase in variable names
 
 Consistency in a code base is key for readability and maintainability.  This rule
-enforces variable definitions and assignments to be in camelCase.  Of note:
+enforces variable declarations and object property names which you create to be
```
magurotuna, line 44: > which you create 

nice & simple explanation 👍

src/rules/camelcase.rs
magurotuna: or, `obj.last_name` example could just be removed as the background why we think of it as valid is hard to put into simple description. But I'd prefer to leave it in the valid examples because this is a difference in behavior between deno_lint and ESLint.

cknight: Thanks for the explanation @magurotuna .  I'm tempted to leave it out for the following reason.  People will generally come to the documentation when their code is flagged as violating this rule.  Therefore, they will first look at the invalid examples to understand why their code is wrong.  Providing a valid, but confusing, example will hamper their understanding. It feels like it would take a fair bit of text to cover this somewhat corner case.  Even with your examples above, in case 2 the declaration of `obj` would need to have happened in an imported module, otherwise deno lint would flag it.  If anyone feels strongly on including the necessary text to explain this valid situation then I can knock up an explanation, but my own opinion is that it would do more harm than good.

src/rules/camelcase.rs
cknight: Agreed.

src/rules/camelcase.rs
magurotuna: @cknight 
That also makes sense to me - I agree to not hamper users understanding. I missed users point of view. So let's land it as is 👍

magurotuna (approved):
LGTM, awesome work!

Issue #466, opened 2020-10-28
no-eval false negatives


cknight:
es-lint flags the following three examples with the no-eval rule which deno lint does not:
```javascript
(0, eval)("var a = 0");

var foo = eval;
foo("var a = 0");

this.eval("var a = 0");
```

ross-weir:
I'll try fix this one 👍

PR #468, opened 2020-10-28
doc: add lint hints and docs

cknight:
This PR contributes to #159 with the following rule additions:
* no-empty-pattern
* no-eval
* no-ex-assign

It also adds some commented out tests related to issue #466 (no-eval false negatives)


magurotuna (approved):
LGTM

Issue #464, opened 2020-10-28
no-global-assign false positives


bartlomieju:
```
onmessage = function (e): void {
  if (typeof self.Deno === "undefined") {
    throw new Error("Deno namespace not available in worker");
  }

  postMessage(e.data);
};

```

```
error[no-global-assign]: Assignment to global is not allowed
 --> /Users/biwanczuk/dev/deno/cli/tests/subdir/deno_worker.ts:1:1
  |
1 | onmessage = function (e): void {
  | ^^^^^^^^^
  |
error[no-global-assign]: Assignment to global is not allowed
 --> /Users/biwanczuk/dev/deno/cli/tests/subdir/event_worker.js:1:1
  |
1 | onmessage = function (e) {
  | ^^^^^^^^^
  |
error[no-global-assign]: Assignment to global is not allowed
 --> /Users/biwanczuk/dev/deno/cli/tests/subdir/event_worker_scope.js:4:1
  |
4 | onmessage = function (e) {
  | ^^^^^^^^^
  |
```

It's perfectly valid to assign value to `onmessage` (and similarly `onerror`)

bartlomieju:
Error on my side, when porting globals in #309 I did not take into account if the variable is overridable.

magurotuna:
looks like this should be reopened

PR #470, opened 2020-10-28
fix: constructor-super for declarations

bartlomieju:
Fixes https://github.com/denoland/deno_lint/issues/464


magurotuna:
seems like you specified the wrong issue?

Issue #457, opened 2020-10-28
Proposal: refactor LintRule trait, message, hint


magurotuna:
I propose we do refactoring `LintRule` trait into:

```rust
pub trait LintRule {
  type Message;
  type Hint;

  const CODE: &'static str;
  const TAGS: &'static [&'static str];
  const DOCS: &'static str;

  fn new() -> Box<Self>
  where
    Self: 'static + Sized;

  fn code(&self) -> &'static str {
    Self::CODE
  }

  fn tags(&self) -> &'static [&'static str] {
    Self::TAGS
  }

  fn docs(&self) -> &'static str {
    Self::DOCS
  }

  fn lint_program(
    &self,
    context: &mut Context,
    program: &swc_ecmascript::ast::Program,
  );
}
```

After this refactoring, trait implementation will look like:

```rust
use derive_more::Display; // comes from `derive_more` crate

#[derive(Display)]
pub enum FooRuleMessage {
  #[display(fmt = "ooops!")]
  Unexpected,
}

#[derive(Display)]
pub enum FooRuleHint {
  #[display(fmt = "some hint")]
  Bar,
  #[display(fmt = "This is dummy hint with two strings! {} {}", _0, _1)]
  Dummy(String, String),
}

impl LintRule for FooRule {
  type Message = FooRuleMessage;
  type Hint = FooRuleHint;

  const CODE: &'static str = "foo_rule";
  const TAGS: &'static [&'static str] = &["recommended"];
  const DOCS: &'static str = "some documents";

  fn new() -> Box<Self> {
    Box::new(FooRule)
  }

  fn lint_program(
    &self,
    context: &mut Context,
    program: &swc_ecmascript::ast::Program,
  ) {
    // snip
  }
}

```

## Motivation 

There are two reasons why I want to refactor. This refactor will allow us to:

1. avoid duplicate "code" definitions
2. get statically typed messages & hints that are associated with `LintRule` type

### 1. avoid duplicate "code" definitions

We already have `code` method on `LintRule` trait but it is not used when error messages are generated. I mean, to take `eqeqeq` rule for example, a diagnostic is added like this:

```rust
      self.context.add_diagnostic_with_hint(
        bin_expr.span,
        "eqeqeq",
        message,
        hint,
      )
```

[...]

PR #472, opened 2020-10-29
refactor: introduce typed message & hint

magurotuna:
As discussed in #457, I gave it a shot to introduce typed message & hint with `derive_more` crate for the following three rules:

- eqeqeq
- no-delete-var
- getter-return

This looks semantically better to me. What do you think? @bartlomieju


src/linter.rs
```diff
@@ -43,30 +43,35 @@ impl Context {
...
     self.diagnostics.push(diagnostic);
   }
 
   fn create_diagnostic(
     &self,
     span: Span,
-    code: impl Into<String>,
-    message: impl Into<String>,
+    code: String,
+    message: String,
```
magurotuna, line 74: This refactor is inspired by https://www.possiblerust.com/pattern/non-generic-inner-functions, in which the author says it is better to decrease generic functons as few as possible for faster compile time and less binary size.

src/rules/getter_return.rs
```diff
@@ -386,134 +401,129 @@ const obj = {
...
       GetterReturn,
 
       // object getter
       "const foo = { get getter() {} };": [
         {
           col: 14,
-          message: msg1("getter"),
-          hint: "Return a value from the getter function",
+          message: GetterReturnMessage::Expected("getter".to_string()),
+          hint: GetterReturnHint::Return,
```
magurotuna, line 412: `.to_string()` may look noisy here. I tried to remove it by some macro, but I couldn't. Probably it requres proc macro.

src/linter.rs
bartlomieju: I'm pretty sure this won't matter much as most of the time is spent compiling swc and its macros anyway :)

magurotuna: Sounds right, to be honest I just wanted to practice it, so I don't have a strong will to do it. I will revert :)

src/rules/getter_return.rs
bartlomieju: The solution looks great, but those `to_string()` calls are noisy. @magurotuna maybe we could use `&str` instead with appropriate lifetime?

magurotuna: sure, I'll give it a shot.

magurotuna: hmmm... I'm doing my best but using `&str` in GetterRetrunMessage seems extremely difficult for me :/ (How I tried is here: 7759d4b810737ef3086f036e0f154fc0e307eb73)
I'll try another solution using a declarative macro.

src/rules/no_delete_var.rs
```diff
@@ -101,7 +112,13 @@ mod tests {
   fn no_delete_var_invalid() {
     assert_lint_err! {
       NoDeleteVar,
-      r#"var someVar = "someVar"; delete someVar;"#: [{ col: 25, message: MESSAGE, hint: HINT }],
+      r#"var someVar = "someVar"; delete someVar;"#: [
+        {
+          col: 25,
+          message: variant!(NoDeleteVarMessage, Unexpected),
+          hint: variant!(NoDeleteVarHint, Remove),
```
bartlomieju, line 119: This seems superfluous - `NoDeleteVarHint::Remove` is shorter 😃

src/rules/getter_return.rs
```diff
@@ -97,35 +114,33 @@ impl<'c> GetterReturnVisitor<'c> {
...
     self.errors.insert(
       span,
-      format!(
-        "Expected to return a value in '{}'.",
+      GetterReturnMessage::Expected(
         self
           .getter_name
           .clone()
-          .expect("the name of getter is not set")
+          .expect("the name of getter is not set"),
```
bartlomieju, line 131: We could potentially write this one as:
```rust
variant!(GetterReturnMessage, Expected, self.getter_name.unwrap())
```

src/rules/no_delete_var.rs
magurotuna: I was wondering which to choose, `variant!(FooEnum, Bar)` or `FooEnum::Bar` if the variant has no value. I don't have strong opinion about this, so let's go with your idea :) I'll rewrite soon.

src/rules/getter_return.rs
magurotuna: Currently `variant!` macro is in `test_util.rs`, which is compiled only when in tests due to `#[cfg(test)]`. 
Do you think `variant!` should be able to be used outside tests?

bartlomieju: Let's leave it in `test_util.rs` for a bit and see if it will be useful in other rules 👍

magurotuna: OK I see.

src/rules/no_delete_var.rs
magurotuna: Done

bartlomieju (approved):
LGTM, I agree this is cleaner. Let's land it and see impact on `deno` binary size in a few days.

PR #481, opened 2020-10-30
fix control flow analysis on for statements

### LINKED ISSUE #477: no-unreachable false positive
```
    const partsA = [];
    const partsB = [];
    for (let i = 0; i < Math.max(partsA.length, partsB.length); i++) {
      const partA = partsA[i];
      const partB = partsB[i];
      if (partA === undefined) return -1;
      if (partB === undefined) return 1;
      if (partA === partB) continue;
      const priorityA = partA.startsWith(":") ? partA.endsWith("*") ? 0 : 1 : 2;
      const priorityB = partB.startsWith(":") ? partB.endsWith("*") ? 0 : 1 : 2;
      return Math.max(Math.min(priorityB - priorityA, 1), -1);
    }
    return 0;
```

Here the `return 0;` is marked as unreachable, although it is reachable if partsA.length == 0 & partsB.length == 0.

magurotuna:
Fixes #477

### DIFF
--- a/src/control_flow.rs
+++ b/src/control_flow.rs
@@ -901,6 +901,42 @@ function foo() {
   }
 
   #[test]
+  fn for_2() {
+    // infinite loop
+    let src = r#"
+function foo() {
+  for (let i = 0; true; i++) {
+    return 1;
+  }
+  bar();
+}
+    "#;
+    let flow = analyze_flow(src);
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, great fix @magurotuna

PR #484, opened 2020-10-31
refactor: use parse_program() for parsing

bartlomieju:
This commit removes "FileType" added in #445. 

This is done by removing "AstParser::parse_script" and "AstParser::parse_module"
in favor of "AstParser::parse_program" which uses new "parse_program"
API provided by SWC.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,11 +38,7 @@ mod lint_tests {
       .build();
 
     linter
-      .lint(
-        "lint_test.ts".to_string(),
-        source.to_string(),
-        FileType::Module,
-      )
+      .lint("lint_test.ts".to_string(), source.to_string())
       .expect("Failed to lint")
   }
 
@@ -80,6 +76,7 @@ mod lint_tests {
       false,
     );
 
+    eprintln!("diagnostics {:#?}", diagnostics);
     assert_eq!(diagnostics.len(), 0);
   }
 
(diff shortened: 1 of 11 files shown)


bartlomieju:
@magurotuna please review

src/rules/no_inner_declarations.rs
```diff
@@ -6,6 +6,7 @@ use swc_ecmascript::ast;
 use swc_ecmascript::visit::noop_visit_type;
 use swc_ecmascript::visit::Node;
 use swc_ecmascript::visit::Visit;
+// use swc_ecmascript::visit::VisitAll;
```
magurotuna, line 9: Maybe you forgot to delete this, or intended to replace `Visit` with `VisitAll`?

src/lib.rs
```diff
@@ -80,6 +76,7 @@ mod lint_tests {
       false,
     );
 
+    eprintln!("diagnostics {:#?}", diagnostics);
```
magurotuna, line 79: Is this `println!` just for debugging?

magurotuna (commented):
Just a few nitpicks, otherwise LGTM 👍

src/rules/no_inner_declarations.rs
bartlomieju: Forgot to delete

src/lib.rs
bartlomieju: Yep, removed

magurotuna (approved):
LGTM

PR #485, opened 2020-11-01
fix(no-fallthrough): allow no body for statement fallthrough

### LINKED ISSUE #467: no-fallthrough misses some statements
```js
      case "symbol":
      case "function":
      default:
        throw new DOMException("Uncloneable value in stream", "DataCloneError");
```

```
   --> /Users/biwanczuk/dev/deno/cli/rt/01_web_util.js:110:7
    |
110 |       case "function":
    |       ^^^^^^^^^^^^^^^^
    |
```

`case "symbol"` should be marked as well

ross-weir:
Fixes #467 

This seems unideal but I couldn't come up with a better way of handling it.

For snippets like this:
```js
switch(foo) {
  case 0:

  default:
    foo()
```

There seems to be no body detected for `case 0`, to get around this count the new lines. If there's no body and there's more than 1 new line in the span of `case 0:\n\n default:` then the body must be made up of only new lines.

(Thanks for the help bartlomieju)

### DIFF
--- a/src/rules/no_fallthrough.rs
+++ b/src/rules/no_fallthrough.rs
@@ -99,12 +99,12 @@ impl<'c> Visit for NoFallthroughVisitor<'c> {
           hi: cases[case_idx + 1].span.lo(),
           ctxt: case.span.ctxt,
         };
-        let snippet = self.context.source_map.span_to_snippet(span).unwrap();
+        let span_lines = self.context.source_map.span_to_lines(span).unwrap();
         // When the case body contains only new lines `case.cons` will be empty.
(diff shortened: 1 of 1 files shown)


RDambrosio016:
This is technically wrong because JS accepts \r, \n, \r\n, \u{2028} and \u{2029} all as valid newlines

ross-weir:
> This is technically wrong because JS accepts \r, \n, \r\n, \u{2028} and \u{2029} all as valid newlines

Thanks for the info! I'll see if I can use `span_to_lines` or something instead of counting by `\n`

bartlomieju (approved):
LGTM, nice fix @ross-weir

PR #486, opened 2020-11-01
fix(no-undef): handle class declarations hoisting

### LINKED ISSUE #463: no-undef false positive
```
...

  function createHttpClient(options) {
    return new HttpClient(opCreateHttpClient(options));
  }

  function opCreateHttpClient(args) {
    return core.jsonOpSync("op_create_http_client", args);
  }

  class HttpClient {
    constructor(rid) {
      this.rid = rid;
    }
    close() {
      core.close(this.rid);
    }
  }


...
```

```
error[no-undef]: HttpClient is not defined
   --> /Users/biwanczuk/dev/deno/op_crates/fetch/26_fetch.js:877:16
    |
877 |     return new HttpClient(opCreateHttpClient(options));
    |                ^^^^^^^^^^
    |
```

Note: the file in question is a script, but it shouldn't really matter here.

magurotuna:
Fixes #463 

This PR fixes #463 by adding a new scope analyzer named `DeclFinder`. Although there's still a problem with variables that are declared with `var`, it works fine for Deno's runtime JS scripts.

### DIFF
--- a/src/rules/no_undef.rs
+++ b/src/rules/no_undef.rs
@@ -1,9 +1,11 @@
 // Copyright 2020 the Deno authors. All rights reserved. MIT license.
+// TODO(magurotuna): remove
+#![allow(unused)]
 use super::Context;
 use super::LintRule;
 use crate::globals::GLOBALS;
-use swc_atoms::js_word;
-use swc_common::SyntaxContext;
+use swc_atoms::{js_word, JsWord};
+use swc_common::{Span, Spanned, SyntaxContext};
 use swc_ecmascript::{
   ast::*,
   utils::ident::IdentLike,
@@ -12,7 +14,9 @@ use swc_ecmascript::{
 };
 use swc_ecmascript::{utils::find_ids, utils::Id};
 
-use std::collections::HashSet;
+use std::cell::RefCell;
+use std::collections::{BTreeMap, BTreeSet, HashSet};
+use std::rc::Rc;
 
 pub struct NoUndef;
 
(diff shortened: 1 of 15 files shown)


magurotuna:
Current status: I've completed fixing the false positive described in #463. Now, running `deno_lint` for Deno's runtime JS code, we can see two discrete errors about `no-undef`:

```
  17   │ error[no-undef]: bootstrap is not defined
  18   │  --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/90_deno_ns.js:3:1
  19   │   |
  20   │ 3 | __bootstrap.denoNs = {
  21   │   | ^^^^^^^^^^^
  22   │   |
  23   │ error[no-undef]: bootstrap is not defined
  24   │  --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/90_deno_ns.js:4:9
  25   │   |
  26   │ 4 |   test: __bootstrap.testing.test,
  27   │   |         ^^^^^^^^^^^
  28   │   |
  29   │ error[no-undef]: bootstrap is not defined
  30   │  --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/90_deno_ns.js:5:12
  31   │   |
  32   │ 5 |   metrics: __bootstrap.metrics.metrics,
  33   │   |            ^^^^^^^^^^^
  34   │   |
  35   │ error[no-undef]: bootstrap is not defined
  36   │  --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/90_deno_ns.js:6:12
  37   │   |
  38   │ 6 |   Process: __bootstrap.process.Process,
  39   │   |            ^^^^^^^^^^^
  40   │   |

.....(the rest is omitted)
```

`90_deno_ns.js` has many errors about `__bootstrap`.

---------------------

```
 843   │ error[no-undef]: currentPart is not defined
 844   │     --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/02_console.js:1161:5
 845   │      |
 846   │ 1161 |     currentPart = "";
 847   │      |     ^^^^^^^^^^^
 848   │      |
 849   │ error[no-undef]: currentPart is not defined
 850   │     --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/02_console.js:1172:25
 851   │      |
 852   │ 1172 |           const value = currentPart.trim();
 853   │      |                         ^^^^^^^^^^^
 854   │      |
 855   │ error[no-undef]: currentPart is not defined
 856   │     --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/02_console.js:1177:11
 857   │      |
 858   │ 1177 |           currentPart = "";
 859   │      |           ^^^^^^^^^^^
 860   │      |
 861   │ error[no-undef]: currentPart is not defined
 862   │     --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/02_console.js:1182:22
 863   │      |
 864   │ 1182 |         currentKey = currentPart.trim();
 865   │      |                      ^^^^^^^^^^^
 866   │      |
 867   │ error[no-undef]: currentPart is not defined
 868   │     --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/02_console.js:1183:9
 869   │      |
 870   │ 1183 |         currentPart = "";
 871   │      |         ^^^^^^^^^^^
 872   │      |
 873   │ error[no-undef]: currentPart is not defined
 874   │     --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/02_console.js:1187:7
 875   │      |
 876   │ 1187 |       currentPart += c;
 877   │      |       ^^^^^^^^^^^
 878   │      |
 879   │ error[no-undef]: currentPart is not defined
 880   │     --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/02_console.js:1190:21
 881   │      |
 882   │ 1190 |       const value = currentPart.trim();
 883   │      |                     ^^^^^^^^^^^
 884   │      |
 885   │ error[no-undef]: currentPart is not defined
 886   │     --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/rt/02_console.js:1195:7
 887   │      |
 888   │ 1195 |       currentPart = "";
 889   │      |       ^^^^^^^^^^^
 890   │      |
```

Looking at `02_console.js`, we can find `currentPart` actually not defined. So this is "true positive".

magurotuna:
However current implementation doesn't handle variables declared with `var` properly. I have to fix it.

src/rules/no_undef.rs
```diff
@@ -26,139 +23,809 @@ impl LintRule for NoUndef {
...
+    //       {
+    //         var target = 0;
+    //       }
+    //     }
+    //     "#;
+    //   let idents = get_idents(src, "target");
+    //   let finder = decl_finder(src);
+    //   let target_used = &idents[0];
+    //   assert!(finder.decl_exists(&target_used));
+    // }
```
magurotuna, line 807: This example shows that `DeclFinder` (which is `no-undef` specific analyzer) cannot deal with `var`. If we uncomment this test, test will fail.
`var target = 0;` is not block-scoped but function-scoped, so `target++` should be treated as valid.

CC @bartlomieju @kdy1

kdy1: This seems like a bug of resolver in swc. I'll check if it's the case.

magurotuna: Oh really? I think there's a problem with my implementation of `DeclFinder`. In `DeclFinder`, it collects variable declarations in a naive way. It doen't use swc's `SyntaxContext` or something like that.

magurotuna: To be honest I have no idea what `SyntaxContext` exactly is. In the original implementation of `no-undef`, the following code resulted in false positive, which was perhaps caused by `SyntaxContext` problem, I guess.

```ts
(() => {
  function foo() {
    return new Bar();
  }
  class Bar {}
})();
```

kdy1: It was not, while the code in #463 is bug of the swc resolver.
I'll investigate further.

magurotuna: I see. Thanks for your help :)

kdy1: Does deno_lint does some transforms?
Root cause of the failing test is `visit_var_declarator` not being called.

kdy1: Oh, I did wrong test.

kdy1: Changing to
```rust

impl BindingCollector {
  fn declare(&mut self, i: Id) {
    dbg!(&i);
    self.declared.insert(i);
  }
}

impl Visit for BindingCollector {
  fn visit_fn_decl(&mut self, f: &FnDecl, _: &dyn Node) {
    dbg!(&f.ident);
    self.declare(f.ident.to_id());
    f.visit_children_with(self);
  }
}
```

fixes the issue. (I changed it to visit children nodes)

magurotuna: Great! I'll give it a try

magurotuna: hmm... tried your fix but the test still fails.

Code fix:
https://github.com/magurotuna/deno_lint/blob/test-magurotuna/src/rules/no_undef.rs#L56-L66

Test case:
https://github.com/magurotuna/deno_lint/blob/test-magurotuna/src/rules/no_undef.rs#L352-L368

Test result:

```
thread 'rules::no_undef::tests::no_undef_valid' panicked at 'Unexpected diagnostics found:
[
    LintDiagnostic {
        range: Range {
            start: Position {
                line: 4,
                col: 19,
                byte_pos: 56,
            },
            end: Position {
                line: 4,
                col: 22,
                byte_pos: 59,
            },
        },
        message: "Bar is not defined",
        code: "no-undef",
        hint: None,
    },
]

source:

    (() => {
      function foo() {
        return new Bar();
      }
      class Bar {}
    })();

', src/test_util.rs:228:5
```

Do I miss something?

kdy1:
```js
(() => {
      function foo() {
        return new Bar();
      }
      class Bar {}
    })();
```

This is bug of swc resolver. I'm currently working on it. I think the test below is a same bug.

magurotuna: > This is bug of swc resolver. I'm currently working on it. I think the test below is a same bug.This is bug of swc resolver. I'm currently working on it. I think the test below is a same bug.

Oh I see. Then I'll wait for swc resolve to be fixed 👍

kdy1: I fixed it and published a new version of `swc_ecmascript` and `swc_ecma_transforms`.

magurotuna: @kdy1 
Thanks for the quick fix!
`deno_lint` depends on `swc_ecmascript`, doesn't depend on `swc_scma_transforms` directly. It seems like `swc_ecma_transforms` as a dependency of `swc_ecmascript` was not updated. (in https://github.com/swc-project/swc/blob/master/ecmascript/Cargo.toml#L27, the version of transforms is still `0.29.1`)
Is this correct?

kdy1: Oh. I forgot updating it but you can do `cargo update -p swc_ecma_transforms`.

magurotuna: nice command! That allows to pass the test. 
I have one more question. Now we've added `f.visit_children_with(self);` to `visit_fn_decl`. Should we add `visit_children_with` to the other visit methods e.g. `visit_class_decl`? I feel like this is required to handle a case like the following:

```ts
class SomeClass {
  someMethod() {
    function foo() {
      return new Bar();
    }
    class Bar {}
  }
}
```

kdy1: Yes. The binding collector should collect all bindings.
Alternatively, you can change it to `VisitAll`.

magurotuna: oh, right, `VisitAll` is the finest solution here. I'll use that :)

magurotuna:
Closing in favor of #497

Issue #393, opened 2020-10-15
example(dlint): uses wrong character to split lines


bartlomieju:
Just a heads up, this is incorrect because ECMAScript defines `\n`, `\r`, `\u{2028}`, and `\u{2029}` all as valid linebreak characters (although we havent fixed it in rslint yet either 😄)

_Originally posted by @RDambrosio016 in https://github.com/denoland/deno_lint/pull/391#discussion_r505495786_

`examples/dlint/main.rs` should be updated to split source file into lines on all enumerated characters.

bartlomieju:
Note to self: use [`SourceFile.lines`](https://docs.rs/swc_common/0.10.4/swc_common/struct.SourceFile.html#structfield.lines) to acquire line indexes from file. All newline characters are handled by SWC.

PR #488, opened 2020-11-01
refactor: return SourceFile from Linter::lint

bartlomieju:
This commit changes return type of "Linter::lint" to return
instance of "swc_common::SourceFile" alongside diagnostics. 
It allows to leverage meta data stored by SWC when formatting 
(eg. line indexes).

Fixes https://github.com/denoland/deno_lint/issues/393

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -817,9 +817,9 @@ dependencies = [
 
 [[package]]
 name = "swc_common"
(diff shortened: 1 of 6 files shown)


bartlomieju:
Blocked by https://github.com/swc-project/swc/pull/1198

PR #489, opened 2020-11-01

cknight:
This PR contributes to #159 by adding linting hints and docs to the following rules:
* no-extra-semi
* no-fallthrough
* no-func-assign
* no-global-assign
* no-import-assign

### DIFF
--- a/src/rules/no_func_assign.rs
+++ b/src/rules/no_func_assign.rs
@@ -54,16 +54,15 @@ function myFunc() {}
 ### Valid:
 ```typescript
 function foo() {}
-const someVar = bar;
+const someVar = foo;
 
 const a = function baz() {
   const someStr = "now I'm a string";
 }
 
-myFunc = existingFunc;
-function myFunc() {}
+const anotherFuncRef = existingFunc;
 
-const myFuncVar = function() {}
+let myFuncVar = function() {}
 myFuncVar = bar;  // variable reassignment, not function re-declaration
 ```
 "#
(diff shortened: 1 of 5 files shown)


src/rules/no_func_assign.rs
```diff
@@ -30,6 +30,44 @@ impl LintRule for NoFuncAssign {
...
+```typescript
+function foo() {}
+const someVar = bar;
+
+const a = function baz() {
+  const someStr = "now I'm a string";
+}
+
+myFunc = existingFunc;
+function myFunc() {}
```
magurotuna, line 64: this snippet appears both in invalid and in valid. Is this intentional?

src/rules/no_func_assign.rs
```diff
@@ -30,6 +30,44 @@ impl LintRule for NoFuncAssign {
...
+
+const myFuncVar = function() {}
+myFuncVar = bar;  // variable reassignment, not function re-declaration
```
magurotuna, line 67:
```suggestion
let myFuncVar = function() {}
myFuncVar = bar;  // variable reassignment, not function re-declaration
```

using `const` here may be confusing since reassignment occurs. I'd prefer `let`.

src/rules/no_func_assign.rs
```diff
@@ -30,6 +30,44 @@ impl LintRule for NoFuncAssign {
...
+}
+
+myFunc = existingFunc;
+function myFunc() {}
+```
+
+### Valid:
+```typescript
+function foo() {}
+const someVar = bar;
```
magurotuna, line 57:
```suggestion
function foo() {}
const someVar = foo;
```

It would be nicer to show that we are allowed to assign a function to other variables.

magurotuna (changes requested):
I left just a few comments, please check them :)

cknight:
Thanks as always, @magurotuna!  That's your review comments now addressed.

bartlomieju (approved):
LGTM, thank you @cknight!

PR #490, opened 2020-11-02
fix(no-eval): catch missing eval cases

### LINKED ISSUE #466: no-eval false negatives

ross-weir:
Fixes #466 

Detect `eval` usage in the following situations:

```
this.eval();
var foo = eval;
(0, eval)();
((eval))();
```

### DIFF
--- a/src/rules/no_eval.rs
+++ b/src/rules/no_eval.rs
@@ -63,15 +63,19 @@ impl<'c> NoEvalVisitor<'c> {
     Self { context }
   }
 
-  fn maybe_diagnose_identifier(&mut self, id: &str, span: &Span) {
-    if id == "eval" {
-      self.context.add_diagnostic_with_hint(
-        *span,
-        "no-eval",
-        "`eval` call is not allowed",
-        "Remove the use of `eval`",
-      );
+  fn maybe_add_diagnostic(&mut self, source: &str, span: Span) {
+    if source.contains("eval") {
+      self.add_diagnostic(span);
     }
+  }
+
+  fn add_diagnostic(&mut self, span: Span) {
+    self.context.add_diagnostic_with_hint(
+      span,
+      "no-eval",
+      "`eval` call is not allowed",
+      "Remove the use of `eval`",
+    );
   }
 }
 
@@ -81,7 +85,8 @@ impl<'c> Visit for NoEvalVisitor<'c> {
   fn visit_var_declarator(&mut self, v: &VarDeclarator, _: &dyn Node) {
     if let Some(expr) = &v.init {
       if let Expr::Ident(ident) = expr.as_ref() {
-        self.maybe_diagnose_identifier(&ident.sym.as_ref(), &v.span);
+        let ident_name = &ident.get_key().unwrap();
+        self.maybe_add_diagnostic(ident_name, v.span);
       }
     }
   }
@@ -90,12 +95,18 @@ impl<'c> Visit for NoEvalVisitor<'c> {
     if let ExprOrSuper::Expr(expr) = &call_expr.callee {
       match expr.as_ref() {
         Expr::Ident(ident) => {
-          self.maybe_diagnose_identifier(&ident.sym.as_ref(), &call_expr.span)
+          let ident_name = &ident.get_key().unwrap();
+          self.maybe_add_diagnostic(ident_name, call_expr.span)
         }
-        Expr::Member(member_expr) => self.maybe_diagnose_identifier(
-          &member_expr.get_key().unwrap(),
-          &call_expr.span,
-        ),
+        Expr::Member(member_expr) => {
+          let member_name = &member_expr.get_key().unwrap();
+          self.maybe_add_diagnostic(member_name, call_expr.span)
+        }
+        Expr::Paren(paren) => {
+          let paren_snippet =
+            &self.context.source_map.span_to_snippet(paren.span).unwrap();
+          self.maybe_add_diagnostic(paren_snippet, call_expr.span);
+        }
         _ => {}
       }
     }
@@ -110,8 +121,7 @@ mod tests {
   #[test]
   fn no_eval_test() {
     assert_lint_err::<NoEval>(r#"eval("123");"#, 0);
-    // TODO These tests should pass (#466)
-    // assert_lint_err::<NoEval>(r#"(0, eval)("var a = 0");"#, 0);
+    assert_lint_err::<NoEval>(r#"(0, eval)("var a = 0");"#, 0);
     assert_lint_err::<NoEval>(r#"var foo = eval;"#, 4);
     assert_lint_err::<NoEval>(r#"this.eval("123");"#, 0);
   }
--- a/src/rules/no_eval.rs
+++ b/src/rules/no_eval.rs
@@ -64,8 +64,8 @@ impl<'c> NoEvalVisitor<'c> {
     Self { context }
   }
 
-  fn maybe_add_diagnostic(&mut self, source: &str, span: Span) {
-    if source == "eval" {
+  fn maybe_add_diagnostic(&mut self, source: &dyn StringRepr, span: Span) {
+    if source.string_repr().unwrap() == "eval" {
       self.add_diagnostic(span);
     }
   }
@@ -84,16 +84,12 @@ impl<'c> NoEvalVisitor<'c> {
       // Nested paren callee ((eval))('var foo = 0;')
       Expr::Paren(paren) => self.handle_paren_callee(paren),
       // Single argument callee: (eval)('var foo = 0;')
-      Expr::Ident(ident) => {
-        let ident_name = &ident.string_repr().unwrap();
-        self.maybe_add_diagnostic(ident_name, ident.span)
-      }
+      Expr::Ident(ident) => self.maybe_add_diagnostic(ident, ident.span),
       // Multiple arguments callee: (0, eval)('var foo = 0;')
       Expr::Seq(seq) => {
         for expr in &seq.exprs {
           if let Expr::Ident(ident) = expr.as_ref() {
-            let ident_name = &ident.string_repr().unwrap();
-            self.maybe_add_diagnostic(ident_name, ident.span)
+            self.maybe_add_diagnostic(ident, ident.span)
           }
         }
       }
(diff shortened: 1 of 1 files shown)


src/rules/no_eval.rs
```diff
@@ -59,23 +62,52 @@ impl<'c> NoEvalVisitor<'c> {
...
+          let ident_name = &ident.sym.as_ref();
+          self.maybe_add_diagnostic(ident_name, call_expr.span)
+        }
+        Expr::Member(member_expr) => {
+          let member_name = &member_expr.get_key().unwrap();
+          self.maybe_add_diagnostic(member_name, call_expr.span)
+        }
+        Expr::Paren(paren) => {
+          let paren_snippet =
+            &self.context.source_map.span_to_snippet(paren.span).unwrap();
```
ross-weir, line 107: Couldn't figure out how to extract identifiers in the parens, let me know if there's a better way to do this!

magurotuna: `paren` in `Expr::Paren(paren)` is `swc_ecma_ast::ParenExpr` type, which has `expr` field in it. This `expr` field is `Box<Expr>` type, and if it is equal to `eval`, it means that this `expr` looks like `Box<Expr::Ident>` where the field `sym`of `Expr::Ident` is `eval`.
Also you should probably take care of nested parens i.e. `((eval))();`

magurotuna: For debugging, you may find `dbg!` so helpful. For example, if we add the following code in `Visit` implementation:

```rust
  fn visit_program(&mut self, p: &swc_ecmascript::ast::Program, _: &dyn Node) {
    dbg!(p);
    p.visit_children_with(self);
  }
```

then we can see what AST looks like.

```
[src/rules/no_eval.rs:69] p = Script(
    Script {
        span: Span {
            lo: BytePos(
                0,
            ),
            hi: BytePos(
                29,
            ),
            ctxt: #0,
        },
        body: [
            Expr(
                ExprStmt {
                    span: Span {
                        lo: BytePos(
                            0,
                        ),
                        hi: BytePos(
                            29,
                        ),
                        ctxt: #0,
                    },
                    expr: Call(
                        CallExpr {
                            span: Span {
                                lo: BytePos(
                                    0,
                                ),
                                hi: BytePos(
                                    29,
                                ),
                                ctxt: #0,
                            },
                            callee: Expr(
                                Paren(
                                    ParenExpr {
                                        span: Span {
                                            lo: BytePos(
                                                0,
                                            ),
                                            hi: BytePos(
                                                9,
                                            ),
                                            ctxt: #0,
                                        },
                                        expr: Seq(
                                            SeqExpr {
                                                span: Span {
                                                    lo: BytePos(
                                                        1,
                                                    ),
                                                    hi: BytePos(
                                                        8,
                                                    ),
                                                    ctxt: #0,
                                                },
                                                exprs: [
                                                    Lit(
                                                        Num(
                                                            Number {
                                                                span: Span {
                                                                    lo: BytePos(
                                                                        1,
                                                                    ),
                                                                    hi: BytePos(
                                                                        2,
                                                                    ),
                                                                    ctxt: #0,
                                                                },
                                                                value: 0.0,
                                                            },
                                                        ),
                                                    ),
                                                    Ident(
                                                        Ident {
                                                            span: Span {
                                                                lo: BytePos(
                                                                    4,
                                                                ),
                                                                hi: BytePos(
                                                                    8,
                                                                ),
                                                                ctxt: #1,
                                                            },
                                                            sym: Atom('eval' type=static),
                                                            type_ann: None,
                                                            optional: false,
                                                        },
                                                    ),
                                                ],
                                            },
                                        ),
                                    },
                                ),
                            ),
                            args: [
                                ExprOrSpread {
                                    spread: None,
                                    expr: Lit(
                                        Str(
                                            Str {
                                                span: Span {
                                                    lo: BytePos(
                                                        10,
                                                    ),
                                                    hi: BytePos(
                                                        28,
                                                    ),
                                                    ctxt: #0,
                                                },
                                                value: Atom('console.log(42);' type=dynamic),
                                                has_escape: false,
                                            },
                                        ),
                                    ),
                                },
                            ],
                            type_args: None,
                        },
                    ),
                },
            ),
        ],
        shebang: None,
    },
)
```

ross-weir: That's extremely useful. Thanks a lot for the help!

src/rules/no_eval.rs
```diff
@@ -1,10 +1,14 @@
...
 use super::Context;
 use super::LintRule;
+use crate::swc_util::StringRepr;
+use swc_common::Span;
 use swc_ecmascript::ast::CallExpr;
 use swc_ecmascript::ast::Expr;
 use swc_ecmascript::ast::ExprOrSuper;
+use swc_ecmascript::ast::VarDeclarator;
 use swc_ecmascript::visit::noop_visit_type;
+use swc_ecmascript::visit::swc_ecma_ast::ParenExpr;
```
magurotuna, line 11:
```suggestion
use swc_ecmascript::ast::ParenExpr;
```

I'm not sure why `ParenExpr` is also exported from `swc_ecmascript::visit::swc_ecma_ast`, but if you don't have any special reason, it's better to import like this.

src/rules/no_eval.rs
```diff
@@ -59,23 +63,61 @@ impl<'c> NoEvalVisitor<'c> {
   fn new(context: &'c mut Context) -> Self {
     Self { context }
   }
+
+  fn maybe_add_diagnostic(&mut self, source: &dyn StringRepr, span: Span) {
+    if source.string_repr().unwrap() == "eval" {
+      self.add_diagnostic(span);
+    }
```
magurotuna, line 70: This `unwrap` seems a kind of dangerous to me as it will likely cause panics in real world. How about writing like this?

```suggestion
    if source.string_repr().as_deref() == Some("eval") {
      self.add_diagnostic(span);
    }
```

src/rules/no_eval.rs
```diff
@@ -59,23 +63,61 @@ impl<'c> NoEvalVisitor<'c> {
...
-            call_expr.span,
-            "no-eval",
-            "`eval` call is not allowed",
-            "Remove the use of `eval`",
-          );
+      match expr.as_ref() {
+        Expr::Ident(ident) => self.maybe_add_diagnostic(ident, call_expr.span),
+        Expr::Member(member_expr) => {
+          self.maybe_add_diagnostic(member_expr, call_expr.span)
         }
```
magurotuna, line 118: This could result in false positive. For example:

```ts
const obj = {
  eval: (val) => { console.log(val); },
};

obj.eval("yeah");
```

In this case calling `obj.eval` is totally fine, but your code will report it mistakenly.

magurotuna (changes requested):

magurotuna:
I'm referring to ESLint's implementation (https://github.com/eslint/eslint/blob/master/lib/rules/no-eval.js). Now I feel like `no-eval` rule is quite difficult to implement if we try to port it perfectly. For instance, in order to check if `this.eval(..)` is valid or not, we also need to check if `this` is the global object or not.

ross-weir:
> I'm referring to ESLint's implementation (https://github.com/eslint/eslint/blob/master/lib/rules/no-eval.js). Now I feel like `no-eval` rule is quite difficult to implement if we try to port it perfectly. For instance, in order to check if `this.eval(..)` is valid or not, we also need to check if `this` is the global object or not.
> 
> I think what we have to do first is port test cases from ESLint(https://github.com/eslint/eslint/blob/master/tests/lib/rules/no-eval.js), which can help us figure out how this rule should be implemented.

Yeah this is quite a bit more complicated than I anticipated. I'm certainly happy to keep chipping away at it but it's going to take me a bit of time compared to you experienced folks :-)

How would you like me to proceed here? I'm thinking something like:

- Remove the `Member` expr false positive
- Port across the tests as `TODO` tests

Then at least the rule is improved slightly and we have something to aim for with the TODO tests going forward?

magurotuna:
> - Remove the Member expr false positive

Yeah, I think it would be nice to remove this false positive, because "false negative" is better than "false positive" in terms of linter.

> - Port across the tests as TODO tests

This seems also good. Personally I like to write test cases first and let them pass one by one as I implement.

To be honest, I don't think we should port this rule perfectly from ESLint, because it is not so important rule and there are more edge cases compared to its importance - we are allowed to make a compromise more or less.
How much compromise we make could be shown by test cases.

ross-weir:
> To be honest, I don't think we should port this rule perfectly from ESLint, because it is not so important rule and there are more edge cases compared to its importance - we are allowed to make a compromise more or less.
How much compromise we make could be shown by test cases.

Makes sense. I ported some tests that I _think_ shouldn't be so bad (I don't really know though, happy to remove/add cases), also added a test case for the false positive.

src/rules/no_eval.rs
```diff
@@ -84,14 +124,42 @@ impl<'c> Visit for NoEvalVisitor<'c> {
...
+  #[test]
+  fn no_eval_invalid() {
+    assert_lint_err! {
+      NoEval,
+      "eval('123');": [{col: 0, message: MESSAGE, hint: HINT}],
+      "(0, eval)('var a = 0');": [{col: 4, message: MESSAGE, hint: HINT}],
+      "((eval))('var a = 0');": [{col: 2, message: MESSAGE, hint: HINT}],
+      "var foo = eval;": [{col: 4, message: MESSAGE, hint: HINT}],
+
+      // TODO
```
magurotuna, line 145: Why not add a link to this PR `https://github.com/denoland/deno_lint/pull/490` here for other developers to understand the context?

magurotuna:
@ross-weir 

> I ported some tests that I think shouldn't be so bad (I don't really know though, happy to remove/add cases), also added a test case for the false positive.

Great 👍  I think it's okay for the time being.

> Completed no-eval for #431 as part of this PR as I was adding new tests.

thanks!

src/rules/no_eval.rs
kitsonk, line 145:
```suggestion
      // TODO (see: https://github.com/denoland/deno_lint/pull/490)
```

magurotuna (approved):
LGTM

kitsonk (approved):
LGTM

PR #493, opened 2020-11-02

cknight:
This PR contributes to #159 with linting hints and docs on the following rules:
* no-inferrable-types
* no-inner-declarations
* no-invalid-regexp

### DIFF
--- a/src/rules/no_inner_declarations.rs
+++ b/src/rules/no_inner_declarations.rs
@@ -30,6 +30,49 @@ impl LintRule for NoInnerDeclarations {
     let mut visitor = NoInnerDeclarationsVisitor::new(context, valid_decls);
     visitor.visit_program(program, program);
   }
+
+  fn docs(&self) -> &'static str {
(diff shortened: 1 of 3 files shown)


src/rules/no_inner_declarations.rs
```diff
@@ -30,6 +30,49 @@ impl LintRule for NoInnerDeclarations {
...
+```typescript
+function doSomething() {}
+if (someBool) {}
+
+var a = 10;
+function someFunc(someVal:number): void {
+  var foo = true;
+  if (someVal > 4) {
+    let b = 10;
+    const fn  = function doSomethingElse() {}
```
magurotuna, line 70:
```suggestion
    const fn = function doSomethingElse() {}
```

magurotuna (changes requested):
so minor nit. Otherwise LGTM 👍

bartlomieju (approved):
LGTM

PR #496, opened 2020-11-03
fix(camelcase): provide better hint message when object pattern has default value

### LINKED ISSUE #416: Provide better hint in camelcase rule
Given the code:

```ts
const { no_camelcased = false } = bar;
``` 

when applying `camelcase` rule to this code, we get a hint that looks like:

```
Consider replacing `{ no_camelcased }` with `{ no_camelcased: noCamelcased }`
```

It's a little confusing because `= false` is gone. 
So, we would be happier if hint was as follows:

```
Consider replacing `{ no_camelcased = false }` with `{ no_camelcased: noCamelcased = false }`
```

_Originally posted https://github.com/denoland/deno_lint/pull/302#issuecomment-712252783_

magurotuna:
Fixes #416 

Hint message of `camelcase` rule is sometimes unfriendly. Consider the following code:

```ts
const { category_id = 1 } = query;
```

Running `camelcase` on this code, we get a hint like:
`` Consider replacing `{ category_id }` with `{ category_id: categoryId } ``
where the default value `1` is gone. If one is not so familiar with JavaScript / TypeScript, he/she may be confused and cannot figure out how the default value works with the suggested hint.

So this PR changes hint message to:
`` Consider replacing `{ category_id = .. }` with `{ category_id: categoryId = .. } ``
so that it can tell users how to add default values by showing `..` mark.

### DIFF
--- a/src/rules/camelcase.rs
+++ b/src/rules/camelcase.rs
@@ -137,17 +137,22 @@ enum IdentToCheck {
   /// Key and value name in object pattern, for example:
   ///
   /// ```typescript
-  /// const { foo } = obj1; // key_name: foo, value_name: None
+  /// // key_name: foo, value_name: None, has_default: false
+  /// const { foo } = obj1;
   ///
-  /// const { foo: bar } = obj2; // key_name: foo, value_name: Some(bar)
+  /// // key_name: foo, value_name: Some(bar), has_default: false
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

PR #499, opened 2020-11-03
feat(example): add `--rule` option to dlint

magurotuna:
This PR adds `--rule` option to `dlint run` commandm, which allows us to run __a certain rule__ only.

It is very useful when developers work on fixing a certain rule's bug. For example, I work on `getter-return` and when I think I've done with it, I want to make sure that there's no regression. Of course we have test cases in `deno_lint` for that purpose, but sometimes it is not enough. So I always run a command like this to see a result:

```sh
ls /path/to/deno/**/*.js -1 | grep -v -e 'node_modules' -e 'target' | xargs cargo run --example dlint -- run 2>&1 | less
```

This command produces so many errors that I cannot confirm there's no regression with my fix.
In this case, if we use `--rule` option, we'll be able to get errors we want to see, which should help find any regression easily.

### DIFF
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -132,8 +132,17 @@ fn run_linter(paths: Vec<String>, filter_rule_name: Option<&str>) {
     let source_code =
       std::fs::read_to_string(&file_path).expect("Failed to read file");
 
+    let rules = if let Some(rule_name) = filter_rule_name {
+      get_recommended_rules()
+        .into_iter()
+        .filter(|r| r.code() == rule_name)
+        .collect()
+    } else {
+      get_recommended_rules()
+    };
+
     let mut linter = LinterBuilder::default()
-      .rules(get_recommended_rules())
+      .rules(rules)
       .lint_unknown_rules(true)
       .lint_unused_ignore_directives(true)
       .build();
@@ -141,15 +150,6 @@ fn run_linter(paths: Vec<String>, filter_rule_name: Option<&str>) {
     let (source_file, file_diagnostics) = linter
       .lint(file_path.to_string(), source_code)
       .expect("Failed to lint");
-
-    let file_diagnostics = if let Some(rule_name) = filter_rule_name {
-      file_diagnostics
-        .into_iter()
-        .filter(|d| d.code == rule_name)
-        .collect()
-    } else {
-      file_diagnostics
-    };
 
     error_counts.fetch_add(file_diagnostics.len(), Ordering::Relaxed);
     let _g = output_lock.lock().unwrap();
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -145,7 +145,7 @@ fn run_linter(paths: Vec<String>, filter_rule_name: Option<&str>) {
     let file_diagnostics = if let Some(rule_name) = filter_rule_name {
       file_diagnostics
         .into_iter()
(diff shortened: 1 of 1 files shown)


magurotuna:
Running `ls /path/to/deno/**/*.js -1 | grep -v -e 'node_modules' -e 'target' -e 'tests' -e 'third_party' -e '00_typescript.js'| xargs cargo run --example dlint -- run --rule camelcase`, the result looks like:


```
error[camelcase]: Identifier 'id_1' is not in camel case.
   --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/system_loader_es5.js:100:11
    |
100 |           id_1 = _c[0],
    |           ^^^^
    |
    = help: Consider renaming `id_1` to `ID_1`
error[camelcase]: Identifier 'r_1' is not in camel case.
   --> /home/yusuktan/Repos/github.com/magurotuna/deno/cli/system_loader_es5.js:127:34
    |
127 |       var m, d, e, s, i, _a, _b, r_1;
    |                                  ^^^
    |
    = help: Consider renaming `r_1` to `R_1`
Found 2 problems
```

Only `camelcase` errors are being extracted.

examples/dlint/main.rs
```diff
@@ -124,23 +124,33 @@ fn display_diagnostics(
...
   let error_counts = Arc::new(AtomicUsize::new(0));
   let output_lock = Arc::new(Mutex::new(())); // prevent threads outputting at the same time
 
   paths.par_iter().for_each(|file_path| {
     let source_code =
       std::fs::read_to_string(&file_path).expect("Failed to read file");
 
-    let mut linter = LinterBuilder::default().rules(get_rules()).build();
+    let mut linter = LinterBuilder::default()
+      .rules(get_recommended_rules())
```
bartlomieju: Why not filter rule list before running linter?

magurotuna: I did that way first, but I realized that doing so led to unexpected result. That said, consider the case where we use the code of 836c2ff239464fb3c14f592dff6cb7894e65d0fd and the target script is:

```ts
// deno-lint-ignore no-explicit-any
let foo: any = 1;
console.log(foo);
```

Running `cargo run --example dlint -- run --rule prefer-const hoge.ts`, then we get:

```
error[ban-unknown-rule-code]: Unknown rule for code "no-explicit-any"
 --> hoge.ts:1:1
  |
1 | // deno-lint-ignore no-explicit-any
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
error[prefer-const]: 'foo' is never reassigned. Use 'const' instead
 --> hoge.ts:2:5
  |
2 | let foo: any = 1;
  |     ^^^
  |
Found 2 problems
```

This shows that the linter treats rules other than the specified one in `--rule` as "unknown" rule. To address this problem, I moved filtering a rule down.

Also I think we could address it by fixing `filter_diagnostics` method in `linter.rs`, though I'm not sure which way is better.

bartlomieju: Ah I see, that makes sense

bartlomieju:
@magurotuna maybe we should address #502 before landing this?

magurotuna:
@bartlomieju OK, that makes sense.

magurotuna:
@bartlomieju PTAL

bartlomieju (approved):

PR #505, opened 2020-11-03
fix ban-unused-ignore and ban-unknown-rule-code

### LINKED ISSUE #502: "ban-unknown-rule-code" should distinguish between known and unknown rules
Configuring linter to run only a single rule will cause a lot of `ban-unknown-rule-code` errors
due to other rules not being run. This is wrong - `ban-unknown-rule-code` should know
all available rules codes even if certain rule is not being run. 

Instead, `ban-unused-ignore` should be raised in case of running with only a single rule configured.

Ref https://github.com/denoland/deno_lint/pull/499#discussion_r516777180

magurotuna:
Fixes #502

### DIFF
--- a/src/linter.rs
+++ b/src/linter.rs
@@ -7,10 +7,10 @@ use crate::diagnostic::{LintDiagnostic, Position, Range};
 use crate::ignore_directives::parse_ignore_comment;
 use crate::ignore_directives::parse_ignore_directives;
 use crate::ignore_directives::IgnoreDirective;
-use crate::rules::LintRule;
+use crate::rules::{get_recommended_rules, LintRule};
 use crate::scopes::Scope;
 use std::cell::RefCell;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::rc::Rc;
 use std::time::Instant;
 use swc_common::comments::SingleThreadedComments;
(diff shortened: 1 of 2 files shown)


src/linter.rs
```diff
@@ -236,10 +236,17 @@ impl Linter {
...
 
-    let rule_codes = rules
+    #[allow(clippy::borrowed_box)]
+    fn to_code(rule: &Box<dyn LintRule>) -> String {
+      rule.code().to_string()
+    }
+
+    let executed_rule_codes =
+      rules.iter().map(to_code).collect::<HashSet<String>>();
+    let available_rule_codes = get_recommended_rules()
```
bartlomieju, line 246: This should probably be `get_all_rules()`

src/linter.rs
```diff
@@ -236,10 +236,17 @@ impl Linter {
     let ignore_directives = context.ignore_directives.clone();
     let diagnostics = &context.diagnostics;
 
-    let rule_codes = rules
+    #[allow(clippy::borrowed_box)]
+    fn to_code(rule: &Box<dyn LintRule>) -> String {
+      rule.code().to_string()
+    }
```
bartlomieju, line 242: IMO superfluous, please remove

bartlomieju (changes requested):

src/linter.rs
magurotuna: Alright, I'll remove soon

bartlomieju (approved):
LGTM

PR #506, opened 2020-11-04
refactor: use HashMap for GLOBALS instead of array

magurotuna:
Currently the static variable `GLOBALS` is declared as an array type.
`GLOBALS` has 132 elements and it is used to check if an identifier is the global one or not in `no-global-assign` and `no-undef`. This check is currently done by naive linear search, which requires `O(N)` time comlexity. Theoretically, it is worse than `HashMap` or `BTreeMap`.

To see the actual performance, I created a simple benchmark and ran it.
Result: https://github.com/magurotuna/array-hashmap-btreemap-benchmark
The result shows that `HashMap` is best on average. If we attempt to look for `MessageEvent` (which is placed at nearly middle of `GLOBALS` array), it takes 20ns in `HashMap` while it does 31ns in array.

`HashMap` is wrapped in `Lazy` so it requires initialization cost at the first access. I think this cost can be ignored compared to the time reduction. Plus `no-global-assign` and `no-undef` do search for `GLOBAL` a lot of times.

Therefore I have adopted `HashMap` here. I think this refactor should make the performance better even though it might be a little.


bartlomieju (approved):
LGTM, thanks

magurotuna:
closing as this is a bit of unfair benchmark as pointed out here: https://discord.com/channels/684898665143206084/684911491035430919/773548683290738729

PR #514, opened 2020-11-07
feat: add require-await rule

### LINKED ISSUE #503: Add require-await rule
[`require-await`](https://eslint.org/docs/rules/require-await) was used in `deno` repository and helped to catch inefficiencies in code. We should add this rule and make it `recommended`. 

CC @kdy1 can you look into it?

magurotuna:
Closes #503 

### TODO

- [x] doc

### DIFF
--- a/src/rules/require_await.rs
+++ b/src/rules/require_await.rs
@@ -59,6 +59,73 @@ impl LintRule for RequireAwait {
     let mut visitor = RequireAwaitVisitor::new(context);
     program.visit_with(program, &mut visitor);
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows async functions that have no await expression
+
+In general, the primary reason to use async functions is to use await expressions inside.
+If an async function has no await expression, it is most likely an unintentional mistake.
+
+### Invalid:
+```typescript
+async function f1() {
+  doSomething();
+}
+
+const f2 = async () => {
+  doSomething();
+};
+
+const f3 = async () => doSomething();
+
+const obj = {
+  async method() {
+    doSomething();
+  }
+};
+
+class MyClass {
+  async method() {
+    doSomething();
+  }
+}
+```
+
+### Valid:
+```typescript
+await asyncFunction();
+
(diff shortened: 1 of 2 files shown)


src/rules/require_await.rs
```diff
@@ -0,0 +1,549 @@
...
+  }
+}
+
+impl FunctionInfoBuilder {
+  fn kind(mut self, kind: FunctionKind) -> Self {
+    self.kind = Some(kind);
+    self
+  }
+
+  #[allow(clippy::wrong_self_convention)]
```
bartlomieju, line 194: These clippy supressions are bad, maybe use `&mut self` instead?

src/rules/require_await.rs
```diff
@@ -0,0 +1,549 @@
...
+    self.function_info = Some(new_function_info);
+
+    // Visit the function's inside
+    func.visit_children_with(self);
+
+    // Check if the function should be reported
+    self.check_function_info(func);
+
+    // Restore upper function info
+    let upper = mem::take(&mut self.function_info.as_mut().unwrap().upper);
```
bartlomieju, line 271: Probably could just call `.take()` on the `upper` instead of using `mem::take`?

src/rules/require_await.rs
```diff
@@ -0,0 +1,549 @@
...
+}
+
+#[derive(Default)]
+struct FunctionInfoBuilder {
+  kind: Option<FunctionKind>,
+  is_async: Option<bool>,
+  is_generator: Option<bool>,
+  is_empty: Option<bool>,
+  upper: Option<Box<FunctionInfo>>,
+}
```
bartlomieju, line 170: @magurotuna is this builder really required? Seems you could construct `FunctionInfo` in place where it's required - seems like a bunch of not really needed code

bartlomieju:
@magurotuna I simplified the code a little, please take a look

magurotuna:
Very nice refactoring. Thanks! @bartlomieju

PR #520, opened 2020-11-08
refactor(no-dupe-keys): switch to `asset_lint_err!` macro

magurotuna:
Part of #431 

Additionally, adds more tests and replaces `Visit` with `VisitAll` to handle nested switch statements.

(So this PR may also be part of #330)

### DIFF
--- a/src/rules/no_duplicate_case.rs
+++ b/src/rules/no_duplicate_case.rs
@@ -1,6 +1,7 @@
 // Copyright 2020 the Deno authors. All rights reserved. MIT license.
 use super::Context;
 use super::LintRule;
+use derive_more::Display;
 use std::collections::HashSet;
 use swc_common::Spanned;
 use swc_ecmascript::visit::noop_visit_type;
@@ -8,6 +9,20 @@ use swc_ecmascript::visit::Node;
 use swc_ecmascript::visit::Visit;
 
 pub struct NoDuplicateCase;
+
+const CODE: &str = "no-duplicate-case";
+
+#[derive(Display)]
+enum NoDuplicateCaseMessage {
+  #[display(fmt = "Duplicate values in `case` are not allowed")]
+  Unexpected,
+}
+
+#[derive(Display)]
+enum NoDuplicateCaseHint {
+  #[display(fmt = "Remove or rename the duplicate case clause")]
+  RemoveOrRename,
+}
 
 impl LintRule for NoDuplicateCase {
   fn new() -> Box<Self> {
@@ -19,7 +34,7 @@ impl LintRule for NoDuplicateCase {
   }
 
   fn code(&self) -> &'static str {
-    "no-duplicate-case"
+    CODE
   }
 
   fn lint_program(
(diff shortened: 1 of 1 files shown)


src/rules/no_duplicate_case.rs
```diff
@@ -128,9 +188,172 @@ switch (someText) {
...
+        {
+          line: 7,
+          col: 6,
+          message: NoDuplicateCaseMessage::Unexpected,
+          hint: NoDuplicateCaseHint::RemoveOrRename,
+        }
+      ],
+
+      // TODO(magurotuna): to pass the following tests, we have to remove comments and whitespaces somehow.
+      // "var a = 1, p = {p: {p1: 1, p2: 1}}; switch (a) {case p.p.p1: break; case p. p // comment\n .p1: break; default: break;}": [
```
bartlomieju, line 312: @magurotuna can you elaborate on this one?

magurotuna: @bartlomieju 
I want to do so, but I don't come up with an idea yet.
It seems like we should implement something to remove comments by ourselves, which I don't think is a good way.

@kdy1 
Can I ask for your help?
I'd like to check if there are duplicates in test sections of a switch statement. In the current implementation, it's done by comparing  two tests in text representation, like this:

```rust
let test_txt = source_map
  .span_to_snippet(test_span)
  .unwrap()
  .replace(|c: char| c.is_whitespace(), "");

if seen.contains(&test_txt) {
  // there is duplicate
}
```

However, this way we cannot handle cases where tests have any comment inside. e.g:

```typescript
// we want the following snippet to be reported

switch (a) {
  case p.p1:
    break;
  case p. /* some comment */ p1:
    break;
}
```

So I think removing comments is straight forward approach here.
If you come up with any better solution, I'd be happy to hear it.

kdy1: You can clone the discriminator of switch statement and remove span and then compare. It's fast enough.

There's a function named [`drop_span`](https://swc.rs/rustdoc/swc_ecma_utils/fn.drop_span.html) and [`DropSpan`](https://swc.rs/rustdoc/swc_ecma_utils/struct.DropSpan.html) in `swc_ecmascript::utils`.

bartlomieju: > There's a function named drop_span and DropSpan in swc_ecmascript::utils. 

Good to know, I think we have duplicate implementation in `src/swc_util.rs`

magurotuna: Great, I'll try it, thanks for the quick reply.

magurotuna: @bartlomieju 

Can I work on removing `DropSpan` from `src/swc_util.rs`?

bartlomieju: @magurotuna sure :)

magurotuna:
@bartlomieju Done. PTAL

bartlomieju (approved):
LGTM, nice work 👍

magurotuna:
oops, my bad, the PR title was wrong...
I meant `refactor(no-duplicate-case): switch to `asset_lint_err!` macro`

PR #524, opened 2020-11-09
refactor: delegate DropSpan to swc_ecmascript::utils

magurotuna:
We have a utility for dropping span from `Expr` in `src/swc_util.rs` and use it in `no-dupe-else-if` implementation.
But as described in https://github.com/denoland/deno_lint/pull/520#discussion_r519424880,  the same utility exists in `swc_ecmascript`. So in this PR, I removed our utility, and delegated it to `swc_ecmascript`.


bartlomieju (approved):
LGTM, nice cleanup

PR #532, opened 2020-11-14
fix(prefer-const): ease handling for destructuring assignment

### LINKED ISSUE #522: Prefer const too aggressive on destructuring
In my opinion, prefer const should not "trigger" in situations where one of the variables gets reassigned during destructuring.  For example, given:

```js
let [a, b] = ["a", "b"];
b = "c";
```

This would error that `a` is not re-assigned, but this should be allowed, as `b` gets reassigned and you can mix mutability on destructuring.

magurotuna:
Fixes #522 

~~WIP~~


magurotuna:
I've finished the fix. What we want to fix looks simple, it's just that we want to imitate ESLint's `{ "destructuring": "any" }`option. In spite of its simplicity, the fix for it was quite complicated.
The basic idea of this fix is, if variables are declared or re-assigned within the same pattern in destructuring assignments, they are treated as "group". For example,  

```ts
let a, b;
({a, b} = x);
```

In the above case, `a` and `b` belong to the same "group".
Next example:

```ts
let a, b, c, d;
({a, b} = x);
({b, c, d} = y);
```

In this case, `a`, `b`, `c`, `d` all belong to the same "group".

Variables belonging to the same group share their statuses about whether they are reassigned or not. See the following example to explain this:

```ts
let a, b; // Both `a` and `b` are "Declared"
({a, b} = x);  // Both `a` and `b` are "Initialized"
a = 1; // `a` is "Reassigned". `b` belongs to the same group as `a`, so `a` is also regarded as "Reassigned".
```

In order to manage "group"s, I adopted DisjointSet, which allows querying if two elements belong to the same group efficiently.

As a result, this fix allows the following test cases to pass:

```ts
let {a, b} = obj; b = 0;
let a, b; ({a, b} = obj); b++;
let a, b, c; ({a, b} = obj1); ({b, c} = obj2); b++;
let {a = 0, b} = obj; b = 0; foo(a, b);
let {a: {b, c}} = {a: {b: 1, c: 2}}; b = 3;
let a, b; ({a = 0, b} = obj); b = 0; foo(a, b);
let { name, ...otherStuff } = obj; otherStuff = {};
(function() { let {a: x = -1, b: y} = {a:1,b:2}; y = 0; })();
(function() { let [x = -1, y] = [1,2]; y = 0; })();
let {a: x = -1, b: y} = {a:1,b:2}; y = 0;
let [x = -1, y] = [1,2]; y = 0;
```

bartlomieju (approved):
LGTM

PR #540, opened 2020-11-16
fix(no-func-assign): handle FnExpr and nested assignments

magurotuna:
Part of #431, #330

What's done in this PR:

- fix scope analyzer to handle `FnExpr` properly
- handle nested assignments using `VisitAll` trait
- switch to `assert_lint_err!` macro
- port tests from ESLint


bartlomieju (approved):
LGTM

PR #541, opened 2020-11-17
fix(no-dupe-keys): deal with BigInt keys

magurotuna:
Thanks to https://github.com/swc-project/swc/pull/1192, BigInt keys can be parsed correctly, which allows us to address a TODO comment in `no-dupe-keys`.

(EDIT) Plus addressed CI issues :)


bartlomieju (approved):
Cool, thanks!

PR #543, opened 2020-11-18
fix(no-inferrable-types): handle nested things properly

magurotuna:
~~WIP~~

- [x] use `VisitAll` instead of `Visit`

(EDIT)
Part of #431 and #330 

The current implementation doesn't handle either nested things or multiple variable declarations. In this PR, I addressed those problems by replacing `Visit` with `VisitAll` and fixing `visit_var_decl` implmenetation.

### DIFF
--- a/src/rules/no_inferrable_types.rs
+++ b/src/rules/no_inferrable_types.rs
@@ -2,6 +2,7 @@
 use super::Context;
 use super::LintRule;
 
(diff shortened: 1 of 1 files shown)


magurotuna:
@bartlomieju PTAL

bartlomieju (approved):
LGTM

PR #557, opened 2020-12-04
chore: remove empty object literal from ban types

lucacasonato:
I think using `{}` as a TS type is perfectly valid in many scenarios. Example:

```ts
interface Page<T> {
  routes: Record<string, string>,
  data: T,
}

Page<{}> // as in, the data is an empty object.
```


crypt096 (approved):

kitsonk:
The problem is that more times than it is valid. I discuss this in #572 in part. I am not supportive of removing it. It is a very common source for type errors. You aren't really specifying an empty object, in the example above, you are effectively specifying `any`.

PR #559, opened 2020-12-05
fix(no-inferrable-types): handle private class fields

### LINKED ISSUE #558: linter ignores private properties
i got this property
`#input: string = "";`
It doesn't show that it is a problem
if I do
`input: string = "";`
Then it does show that is is a problem
```
(no-inferrable-types) inferrable types are not allowed
    input:         string        =    "";
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    at \\?\C:\Users\User\Desktop\coding\wrappers\ffmpeg\deno\mod.ts:38:4

    hint: Remove the type, it is easily inferrable```

magurotuna:
This PR fixes `no-inferrable-types` so that it will check private class fields as well.
Also, adds tests to make sure this rule works as expected for the following cases:

- private instance fields
- private static fields
- private instance methods
- private static methods

Fixes #558

### DIFF
--- a/src/rules/no_inferrable_types.rs
+++ b/src/rules/no_inferrable_types.rs
@@ -5,8 +5,9 @@ use super::LintRule;
 use derive_more::Display;
 use swc_ecmascript::ast::{
   ArrowExpr, CallExpr, ClassProp, Expr, ExprOrSuper, Function, Ident, Lit,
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
Lgtm

PR #560, opened 2020-12-06
fix(getter-return): proper handling for function in getter

magurotuna:
This PR addresses TODO comments in `getter-return` so that the rule will display appropriate error messages in cases where getter functions contain function expression, function declaration, or arrow expression.

For example, running `getter-return` rule on the following code:

```typescript
const foo = {
  get bar() {
    function f() {
      return true;
    }
  },
};
```

we see `Expected 'bar' to always return a value.`, but after this PR is merged, we will see `Expected to return a value in 'bar'.`


bartlomieju (approved):
LGTM

PR #562, opened 2020-12-07
refactor(getter-return): replace std::mem::take with Option::take

magurotuna:
Just a minor refactoring. Replaces `std::mem::take` with `Option::take`.


bartlomieju (approved):
LGTM

PR #563, opened 2020-12-08
feat(plugin): expose ControlFlow to JS

magurotuna:
PoC

Let's say we have `foo.js` with the following content:

```javascript
function f() {
  return 1;
  return "f";
}

function g() {
  throw new Error();
  return "g";
}
```

running `cargo run --example dlint -- run foo.ts --plugin ./examples/dlint/plugins/test_plugin3.js`, then we can see:

image

### DIFF
--- a/examples/dlint/control-flow.js
+++ b/examples/dlint/control-flow.js
@@ -1,0 +1,18 @@
+class ControlFlow {
+  static isReachable(stmt) {
+    const { isReachable } = Deno.core.jsonOpSync("query_control_flow_by_span", {
+      span: stmt.span,
+    });
+    return isReachable;
+  }
+
+  static stopsExecution(stmt) {
+    const { stopsExectuion } = Deno.core.jsonOpSync(
+      "query_control_flow_by_span",
+      {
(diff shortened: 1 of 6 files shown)


examples/dlint/control-flow.js
```diff
@@ -0,0 +1,18 @@
+class ControlFlow {
+  static isReachable(stmt) {
+    const { isReachable } = Deno.core.jsonOpSync("query_control_flow_by_span", {
```
bartlomieju, line 3: Nitpick: can you add `op_` prefix to names of ops? I find it a lot easier to search for op call sites

bartlomieju (commented):
@magurotuna this is great!

examples/dlint/control-flow.js
magurotuna: Oh I forgot that. I'll do it tomorrow (because it's late at night here :)

magurotuna: Done b6a85be

bartlomieju (approved):
LGTM, no comments

PR #568, opened 2020-12-14
fix(prefer-as-const): handle `var_decl` that has no `decl`

### LINKED ISSUE #567: prefer-as-const panics on empty file
Opening a new document in an editor and starting to type causes panics in the LSP:

```
thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 0', /Users/kitsonk/.cargo/registry/src/github.com-1ecc6299db9ec823/deno_lint-0.2.13/src/rules/prefer_as_const.rs:99:26
```

It seems like we need to keep "empty" documents from being linted.

magurotuna:
Fixes #567 

~~I'll undraft it after some more refinement.~~

Refs:
#431 #330

### DIFF
--- a/src/rules/prefer_as_const.rs
+++ b/src/rules/prefer_as_const.rs
@@ -5,8 +5,8 @@ use super::LintRule;
 use derive_more::Display;
 use swc_common::{Span, Spanned};
 use swc_ecmascript::ast::{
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, thanks @magurotuna

PR #570, opened 2020-12-15
fix(valid-typeof): allow comparison with variables other than undefined, null, Object

### LINKED ISSUE #566: Comparing a literal string union variable to `typeof` throws a `valid-typeof` error
When I run `deno lint --unstable` on this typescript I am getting a error
```ts
function test(typeTest: "string" | "number" | "boolean", input: any) {
	if(typeTest === typeof input)
		console.log("Yey")
}
```
```
(valid-typeof) Invalid typeof comparison value
	if (expected === typeof input) {
     ^^^^^^^^
    at file.ts:1:64
```

magurotuna:
Fixes #566 

Please see https://github.com/denoland/deno_lint/issues/566#issuecomment-743189703 before reviewing because this fix relaxes the current restriction of `valid-typeof` rule, being against the previous discussion at #38.

In addition, I've done with the following things:
- use `VisitAll` instead of `Visit` to handle nested cases (#330)
- update documentation

### DIFF
--- a/src/rules/valid_typeof.rs
+++ b/src/rules/valid_typeof.rs
@@ -1,11 +1,9 @@
 // Copyright 2020 the Deno authors. All rights reserved. MIT license.
 use super::{Context, LintRule};
 use swc_common::Spanned;
-use swc_ecmascript::ast::BinaryOp::{EqEq, EqEqEq, NotEq, NotEqEq};
-use swc_ecmascript::ast::Expr::{Lit, Unary};
-use swc_ecmascript::ast::Lit::Str;
-use swc_ecmascript::ast::UnaryOp::TypeOf;
-use swc_ecmascript::ast::{BinExpr, Program};
+use swc_ecmascript::ast::{
+  BinExpr, BinaryOp, Expr, Ident, Lit, Program, UnaryOp,
+};
 use swc_ecmascript::visit::{noop_visit_type, Node, Visit};
 
 pub struct ValidTypeof;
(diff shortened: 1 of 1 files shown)


crypt096 (approved):

bartlomieju:
@kitsonk @lucacasonato @nayeemrmn what do you think about this PR?

kitsonk:
It would be good to get an example of what this would trigger on (e.g. a "bad" example).  I am confused on what it would add over and above existing TypeScript strict type checking.

magurotuna:
@kitsonk 
Basically, this fix allows our `valid-typeof` to behave in the same way as ESLint's `valid-typeof` with `requireStringLiterals: false`. To put it simply, it reports "typo"s in string literals that are compared with `typeof xxx` values, while anything else is always okay.
The following code would be  "bad" examples (taken from https://eslint.org/docs/rules/valid-typeof). These are "typo"s.

```typescript
typeof foo === "strnig"
typeof foo == "undefimed"
typeof bar != "nunber"
typeof bar !== "fucntion"
```

There's one difference between ESLint's `valid-typeof` with  `requireStringLiterals: false` and the implementation in this PR.
ESLint allows comparison with things like `undefined` literal, while the implementation in this PR _doesn't_ allow them. For example, the following cases are treated as _valid_ by ESLint, but _invalid_ by ours.

```typescript
typeof foo === undefined
typeof foo == null
typeof foo !== Object
```

I think TypeScript checking also treat these kinds of comparison as _valid_, but they are most likely mistakes by a programmer, so it'd be good to lint it.

In summary, this PR changes two things:

- allows comparison with normal variables (e.g. `typeof foo === someVariable`). This is the initial motivation coming from #566.
- in spite of the above change, comparison with special literals such as `undefined`, `null`, or `Object` is left to be _invalid_. (different from ESLint).

kitsonk:
[Of those](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEA3W8oC5UDsCeBuAUPgC7YAOIA9gGarwC8D8ARAM5EyYCWA5k8WZRpR6dZgFdMoKpwC2IYHxLlqtAISimmCQCMQMRQJXD1GqmLCYinCpj79lQkaIlTOmefcG1GWiBE9G8CbwAPLaAFbgRPhAA), only these two don't produce errors in TypeScript (which is somewhat disappointing, as I would have thought they would have had overlap type errors):

```ts
typeof a === undefined
typeof a == null
```

I am of the opinion we should avoid overlap of such things, personally.

magurotuna:
@kitsonk 
Oh okay, then `valid-typeof` rule seems almost useless in the context of TypeScript.
So it means that our `valid-typeof` rule should check comparison with `undefined` and `null` only.

One concern I have is whether or not we need to care about JavaScript. If we avoided overlap with TypeScript, erroneous codes like `typeof foo === "strnig"`, if written in JavaScript, wouldn't be checked at all. Therefore I think that the minimum overlap to lint JavaScript well is necessary.

ry:
@magurotuna What's the status of this PR? Are you still trying to get this merged? If so, please rebase, otherwise please close.

magurotuna:
Closing because it's stale. More consideration and discussion will be needed.

PR #577, opened 2020-12-26
chore: delegate FsModuleLoader to deno_core

magurotuna:
This PR bumps `deno_core` to 0.74.0 and removes our own `FsModuleLoader` so that it will be delegated to `deno_core`.

refs:
- #415 
- https://github.com/denoland/deno/pull/8523


bartlomieju (approved):
LGTM

PR #589, opened 2021-01-15
fix(camelcase): don't check ambient declarations + handle TypeScript things

### LINKED ISSUE #587: The camel-case rule should probably not apply to ambient declarations
So I've been dabbling a little bit with an AssemblyScript port and ran into this scenario

```ts
cat test.ts
declare function internal_intrinsic(): void;
deno lint --unstable test.ts
(camelcase) Identifier 'internal_intrinsic' is not in camel case.
declare function internal_intrinsic(): void;
                 ^^^^^^^^^^^^^^^^^^
    at /home/caspervonb/test.ts:1:17

    hint: Consider renaming `internal_intrinsic` to `internalIntrinsic`

Found 1 problem
Checked 1 file
```

We probably shouldn't be doing that, reasoning being that ambient declarations are the environment's fault and the user is unable to do anything about it.

magurotuna:
Fixes #587, and additionally, lets `camelcase` rule check TypeScript things such as type aliases, interfaces, and so on.

### DIFF
--- a/src/rules/camelcase.rs
+++ b/src/rules/camelcase.rs
@@ -1042,7 +1042,7 @@ mod tests {
             {
               col: 13,
               message: "Identifier 'snake_case' is not in camel case.",
-              hint: "Consider renaming `snake_case` to `snakeCase`",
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

PR #592, opened 2021-01-16
fix(dlint): make order of diagnostics output deterministic

### LINKED ISSUE #591: Output from dlint example is not deterministic
Before printing diagnostics in `dlint` they should be first sorted by files to ensure deterministic output; currently there is no sorting so running `dlint` multiple times will reorder diagnostics.

magurotuna:
Fixes #591

### DIFF
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -17,9 +17,8 @@ use rayon::prelude::*;
 use serde::Serialize;
 use std::collections::BTreeMap;
 use std::path::PathBuf;
-use std::sync::mpsc;
-use std::sync::Arc;
-use std::thread;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
 use swc_common::BytePos;
 
 mod config;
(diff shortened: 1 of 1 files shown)


bartlomieju (commented):
@magurotuna thanks, LGTM, I have a request though: could you try to rewrite `dlint` in such a way that it pushes the diagnostics over the channel and then they are sorted and displayed in main thread without using mutexes? @RDambrosio016 suggested on Discord that printing diagnostics from each thread was a major slow down in his setup, I wonder if it would affect dlint's as well.

magurotuna:
using mutex(f1beab4)

```
/usr/bin/time -p ./target/release/examples/dlint run ~/Repos/github.com/magurotuna/deno/std/**/*.ts
real 0.04
user 0.55
sys 0.06
```

using channel(f828125):

```
/usr/bin/time -p ./target/release/examples/dlint run ~/Repos/github.com/magurotuna/deno/std/**/*.ts
real 0.04
user 0.55
sys 0.07
```

On my machine, it seems like both mutex and channel show almost the same performance🤔  (It's a very simple benchmark though)

bartlomieju (approved):
@magurotuna thanks for checking! Since there's no noticeable change I reverted your last commit, thanks for the fix!

PR #593, opened 2021-01-17
feat: add lint_with_ast API

### LINKED ISSUE #580: Add API to lint AST instead of source code
For the purpose of https://github.com/denoland/deno/issues/8641 we should have additional APIs on `Linter` struct that allows to pass already parsed AST instead of source code.

Current API:
https://github.com/denoland/deno_lint/blob/14a4dccfbd84e13f3fc2e245961f3c84f3ffc339/src/linter.rs#L212-L216

Proposed API:
```rust
  pub fn lint_ast(
    &mut self,
    file_name: String,
    ast: swc_ecmascript::ast::Program,
  ) -> Result<
    (Rc<swc_common::SourceFile>, Vec<LintDiagnostic>),
    SwcDiagnosticBuffer,
  >
```

magurotuna:
This PR adds a new public API `lint_with_ast`. Here's its signature:

```rust
pub fn lint_with_ast(
    &mut self,
    file_name: String,
    ast: &swc_ecmascript::ast::Program,
    comments: swc_common::comments::SingleThreadedComments,
    source_map: Rc<SourceMap>,
  ) -> Result<
    (Rc<swc_common::SourceFile>, Vec<LintDiagnostic>),
    SwcDiagnosticBuffer,
  >
```

where `self` is struct of `Linter`.

The proposed API signature in #580 didn't have `comments` and `source_map` as parameters, but it seems they are also required to lint.
I tried to see if it was possible for `comments` to be an immutable reference rather than the value itself, but in vain. Still it works well, though.

Closes #580

### DIFF
--- a/src/ast_parser.rs
+++ b/src/ast_parser.rs
@@ -162,6 +162,8 @@ impl AstParser {
     syntax: Syntax,
     source_code: &str,
   ) -> Result<(ast::Program, SingleThreadedComments), SwcDiagnosticBuffer> {
+    // NOTE: calling `self.source_map.new_source_file` mutates `source_map`
+    // even though it's of type `Rc`.
     let swc_source_file = self.source_map.new_source_file(
       FileName::Custom(file_name.to_string()),
       source_code.to_string(),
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):
LGTM, no comments, thanks @magurotuna!

PR #595, opened 2021-01-17
doc: Add docs and tests for no_var

umucorn:
This PR contributes to #159 by adding linting docs to `no_var`.

Should I add non-failing tests for `no_var` using `const` and `let` keywords? I'm not sure because `const` and `let` keywords are not even visited by this rules code.

### DIFF
--- a/src/rules/no_var.rs
+++ b/src/rules/no_var.rs
@@ -82,4 +82,12 @@ mod tests {
       0,
     );
   }
+
+  #[test]
+  fn no_var_valid() {
+    assert_lint_ok::<NoVar>(
+      r#"let foo = 0; const bar = "bar""#,
+      0
+    );
+  }
 }
--- a/src/rules/no_var.rs
+++ b/src/rules/no_var.rs
@@ -10,6 +10,7 @@ use swc_ecmascript::visit::Visit;
 pub struct NoVar;
 
 const MESSAGE: &str = "`var` keyword is not allowed.";
+const CODE: &str = "no-var";
 
 impl LintRule for NoVar {
   fn new() -> Box<Self> {
@@ -17,7 +18,7 @@ impl LintRule for NoVar {
   }
 
   fn code(&self) -> &'static str {
-    "no-var"
+    CODE
   }
 
   fn lint_program(
@@ -36,7 +37,7 @@ impl LintRule for NoVar {
 
 ### Invalid:
 ```typescript
-var foo = 'bar';
+var foo = "bar";
 ```
 
 ### Valid:
@@ -63,9 +64,7 @@ impl<'c> Visit for NoVarVisitor<'c> {
 
   fn visit_var_decl(&mut self, var_decl: &VarDecl, _parent: &dyn Node) {
     if var_decl.kind == VarDeclKind::Var {
-      self
-        .context
-        .add_diagnostic(var_decl.span, "no-var", MESSAGE);
+      self.context.add_diagnostic(var_decl.span, CODE, MESSAGE);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


magurotuna:
Thanks for adding documentation!

> Should I add non-failing tests for no_var using const and let keywords?

Yes, that would be fine. In fact, `no-var` has been listed in #330 (and in #431) due to the insufficiency of tests. I'd be happy if you could add tests for it :)

magurotuna (approved):
LGTM 👍 
If you'd like to add tests, feel free to do so.

src/rules/no_var.rs
```diff
@@ -26,6 +28,24 @@ impl LintRule for NoVar {
...
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Enforces the use of block scoped variables over more error prone function scoped variables. Block scoped variables are defined using `const` and `let` keywords.
+
+`const` and `let` keywords ensure the variables defined using these keywords are not accessible outside their block scope. On the other hand, variables defined using `var` keyword are only limited by their function scope.
+
+### Invalid:
+```typescript
+var foo = 'bar';
```
magurotuna, line 39:
```suggestion
var foo = "bar";
```

It's trivial, but I think double quotations should be better, because `deno fmt` command, which is the built-in formatted of Deno, prefers them to single quotations.

src/rules/no_var.rs
```diff
@@ -43,11 +63,9 @@ impl<'c> Visit for NoVarVisitor<'c> {
...
   fn visit_var_decl(&mut self, var_decl: &VarDecl, _parent: &dyn Node) {
     if var_decl.kind == VarDeclKind::Var {
-      self.context.add_diagnostic(
-        var_decl.span,
-        "no-var",
-        "`var` keyword is not allowed",
-      );
+      self
+        .context
+        .add_diagnostic(var_decl.span, "no-var", MESSAGE);
```
magurotuna, line 68: Could you make `"no-var"` const just like `MESSAGE`?  Here's an example: https://github.com/denoland/deno_lint/blob/master/src/rules/no_eval.rs#L17

src/rules/no_var.rs
```diff
@@ -63,5 +81,32 @@ mod tests {
       r#"var someVar = "someString"; const c = "c"; let a = "a";"#,
       0,
     );
+
+    assert_lint_err!(
+      NoVar,
+      "var foo = 'bar';": [{
+        col: 0,
+        message: MESSAGE,
+      }],
```
magurotuna, line 90: Would be great to let this `assert_lint_err!` macro contain all invalid test cases.

```suggestion
    assert_lint_err!(
      NoVar,
      r#"var someVar = "someString"; const c = "c"; let a = "a";"#: [{
        col: 0,
        message: MESSAGE,
      }],
      "var foo = 'bar';": [{
        col: 0,
        message: MESSAGE,
      }],
```

src/rules/no_var.rs
```diff
@@ -63,5 +81,32 @@ mod tests {
...
+          message: MESSAGE,
+        }
+      ]
+    );
+  }
+
+  #[test]
+  fn no_var_valid() {
+    assert_lint_ok::<NoVar>(r#"let foo = 0; const bar = "bar""#);
   }
```
magurotuna, line 111: Can you move `no_var_valid` function above the `no_var_invalid` function? Because all the tests for the rules follow this convention of order :)

umucorn:
I have added tests according to #431.

magurotuna (changes requested):
Very nice work. Just a couple of comments I have 👍

src/rules/no_var.rs
umucorn: Thank you, I think it should cover all of these invalid cases. But I think it already covers in the following lines though, or I am missing something

src/rules/no_var.rs
umucorn: Yes, I moved it to top now

src/rules/no_var.rs
magurotuna: Ah yes, my point has been addressed in your fix 459fa16. I just meant to remove `assert_lint_err::<NoVar>(...)` so it looks totally fine now :)

src/rules/no_var.rs
```diff
@@ -76,37 +75,32 @@ mod tests {
...
 
   #[test]
-  fn no_var_invalid() {
-    assert_lint_err::<NoVar>(
-      r#"var someVar = "someString"; const c = "c"; let a = "a";"#,
-      0,
-    );
+  fn no_var_valid() {
+    assert_lint_ok::<NoVar>(r#"let foo = 0; const bar = "bar""#);
+  }
```
magurotuna, line 80: Just one more request - could you also replace this `assert_lint_ok` function with `assert_lint_ok!` macro? like https://github.com/denoland/deno_lint/blob/master/src/rules/no_namespace.rs#L62-L68

magurotuna (changes requested):

src/rules/no_var.rs
umucorn: Done

magurotuna (approved):
LGTM, great work @umucorn 

I'd like to ask @bartlomieju to review and merge it.

bartlomieju (approved):
LGTM, thanks @umucorn!

Issue #571, opened 2020-12-15
Try re-implementing some rule using dprint-swc-ecma-ast-view 


magurotuna:
https://github.com/dsherret/dprint-swc-ecma-ast-view

@dsherret created this awesome crate that wraps swc's AST, which allows navigating AST more freely than swc's visitor we're currently using.
I think it will make lots of rules easy to implement, e.g. `no-await-in-loop` in nearly the same way as RSLint does. So I'd say it's worth trying to use it.

dsherret:
There's a few small details I need to work out, but it's basically ready. I'm going to spend another weekend on it and then it should be good for use!

dsherret:
@magurotuna ok, should be good to use in 0.3.0 I believe. Let me know if you have any suggestions for anything or find any bugs.

magurotuna:
@dsherret Great! I'll take a stab at it :)

magurotuna:
I'm kind of busy recently, which prevents me from trying to implement it. Although I don't have time to implement, I elaborate on how to make it in my mind. I'll manage it as soon as possible :)

bartlomieju:
@magurotuna should we still keep this issue open?

magurotuna:
sorry, didn't notice the mention. There are still lots of rules that need to be changed to use ast-view. I'll open an issue to keep track of progress.

PR #604, opened 2021-01-26
POC: re-implement `no-with` and `no-await-in-loop` with dprint-swc-ecma-ast-view

magurotuna:
This is POC for #571

`no-with` has so simple implementation that it won't need the power of dprint-swc-ecma-ast-view, but I chose it as a first step to see if the new implementation works fine in harmony with the existing rules.

One thing I've found unfortunate is that dprint-swc-ecma-ast-view only accepts swc's `Module`, instead of `Program`. I opened the issue about it (https://github.com/dprint/dprint-swc-ecma-ast-view/issues/4)

CC @dsherret @bartlomieju


magurotuna:
Next, as a more complex rule than `no-with`, I attempted to re-implement `no-await-in-loop`.
It seems like the same functionality has been accomplished with far less code. In fact, the previous implementation of `no-await-in-loop` has 189 lines of code:
https://github.com/denoland/deno_lint/blob/fc51afc371a4f09cf3b7601f8dc135f7dd97107a/src/rules/no_await_in_loop.rs#L177-L363

whereas the new code has only 59 lines (including comments):
https://github.com/denoland/deno_lint/blob/fc51afc371a4f09cf3b7601f8dc135f7dd97107a/src/rules/no_await_in_loop.rs#L117-L175

src/rules/no_await_in_loop.rs
```diff
@@ -88,6 +114,66 @@ impl<'c> NoAwaitInLoopVisitor<'c> {
...
 }
 
+impl<'c> Handler for NoAwaitInLoopVisitor<'c> {
+  fn await_expr(&mut self, await_expr: &AstView::AwaitExpr) {
+    fn inside_loop(
+      await_expr: &AstView::AwaitExpr,
+      node: AstView::Node,
+    ) -> bool {
+      use NodeKind::*;
+      match node.kind() {
```
dsherret, line 124: Instead of matching on `node.kind()` and doing `.expect`, it's possible to match directly on the node.

Additionally, instead of doing `.inner.await_token` it's possible to do `.await_token()` instead, which will get `inner.await_token()` (there should be a helper method for ever property found on `inner`, so `.inner.x` -> `.x()`).

For example:

```ts
use Node::*;
match node {
  FnDecl(_) | FnExpr(_) | ArrowExpr(_) => false,
  ForOfStmt(stmt)
    if stmt.await_token().is_some() =>
  {
    // etc...
  }
  // etc...
}
```

src/rules/no_await_in_loop.rs
```diff
@@ -32,6 +36,28 @@ impl LintRule for NoAwaitInLoop {
     visitor.visit_program(program, program);
   }
 
+  fn lint_program_with_ast_view(
```
dsherret, line 39: 0.6.0 supports providing a `Program` instead as per your recommendation 🙂

src/rules/no_await_in_loop.rs
```diff
@@ -88,6 +114,66 @@ impl<'c> NoAwaitInLoopVisitor<'c> {
...
+          // When it encounters `ForInStmt` or `ForOfStmt`, we should treat it as `inside_loop = true`
+          // except for the case where the given `await_expr` is contained in the `right` part.
+          // e.g. for (const x of await xs) { ... }
+          //                      ^^^^^^^^ <-------- `right` part
+          let right = match node.kind() {
+            ForInStmt => &*node.expect::<AstView::ForInStmt>().inner.right,
+            ForOfStmt => &*node.expect::<AstView::ForOfStmt>().inner.right,
+            _ => unreachable!(),
+          };
+          !right.span().contains(await_expr.span())
```
dsherret, line 146: I had no idea about this `.contains` method on span. Good to know!

src/rules/no_await_in_loop.rs
```diff
@@ -88,6 +114,66 @@ impl<'c> NoAwaitInLoopVisitor<'c> {
...
+            Some(p) => p,
+            None => return false,
+          };
+          inside_loop(await_expr, parent)
+        }
+      }
+    }
+
+    if inside_loop(await_expr, await_expr.into_node()) {
+      self.add_diagnostic(await_expr.inner.span);
```
dsherret, line 172: Could do `await_expr.span()`.

dsherret (commented):
@magurotuna looks really nice! I think it can even be further improved.

src/rules/no_await_in_loop.rs
magurotuna: Thanks! Sounds good suggestion, I rewrote the code.

dsherret: Looks really good!

src/rules/no_await_in_loop.rs
```diff
@@ -24,12 +21,21 @@ impl LintRule for NoAwaitInLoop {
...
     &self,
     context: &mut Context,
-    program: &swc_ecmascript::ast::Program,
+    program_info: dprint_swc_ecma_ast_view::ProgramInfo<'a>,
   ) {
-    let mut visitor = NoAwaitInLoopVisitor::new(context);
-    visitor.visit_program(program, program);
+    with_ast_view(program_info, |pg| {
+      NoAwaitInLoopHandler.traverse(pg, context);
+    });
```
dsherret, line 38: The "view" should be created really fast, but ideally you would only do this once per file instead of once per rule.

magurotuna: Good point. Potentially it could. 
We have to take into account that [`lint_program_with_ast_view` method on `LintRule` trait](https://github.com/denoland/deno_lint/blob/17ce8d0f71b786732f1341fe398bb724b0db109f/src/rules.rs#L100) needs to fall back to `lint_program` if a rule doesn't implement `lint_program_with_ast_view`, because we need to maintain both the existing rule implementations that use swc's AST and new rule implementations that use dprint ast-view's stuff until all the existing rules are replaced.

So, the default implementation of `lint_program_with_ast_view` should look like:

```rust
  /// Executes lint using `dprint-swc-ecma-ast-view`.
  /// Falls back to the `lint_program` method if not implemented.
  fn lint_program_with_ast_view<'a>(
    &self,
    context: &mut Context,
    program: dprint_swc_ecma_ast_view::Program;
,
  ) {
    use dprint_swc_ecma_ast_view::Program::*;
    let swc_program = match program {
      Module(m) => swc_ecmascript::ast::Program::Module(m.inner.clone()),
      Script(s) => swc_ecmascript::ast::Program::Script(s.inner.clone()),
    };
    self.lint_program(context, &swc_program);
  }
```

If the signature of `lint_program_with_ast_view` was like above, we could run all rules like:

```rust
    dprint_swc_ecma_ast_view::with_ast_view(program_info, |pg| {
      // Run builtin rules
      for rule in &self.rules {
        rule.lint_program_with_ast_view(&mut context, pg);
      }
    });
```

This code is fine as the ast-view is created once, but I'm concerned about the cost of cloning swc's Module or Script in these lines:

```rust
      Module(m) => swc_ecmascript::ast::Program::Module(m.inner.clone()),
      Script(s) => swc_ecmascript::ast::Program::Script(s.inner.clone()),
```

Any good ideas?

dsherret: Considering `lint_program` takes a reference to an swc `Program` enum, perhaps instead of cloning, we could introduce a new `Program` enum that instead holds a reference to the swc `Module` or `Script`?

```
pub enum ProgramRef<'a> {
    Module(&'a Module),
    Script(&'a Script),
}
```

Then that could be provided to `lint_program` instead.

magurotuna: @dsherret Ahh looks cool! That way we could avoid wasteful cloning. Will rewrite it!

magurotuna:
@dsherret 
I rewrote it to use `ProgramRef` instead of cloning. Could you have a look? (might be hard to see the diff as it's pretty big)

dsherret:
@magurotuna looks good to me! 🙂

src/linter.rs
```diff
@@ -415,10 +416,20 @@ impl Linter {
...
-    // Run builtin rules
-    for rule in &self.rules {
-      rule.lint_program(&mut context, program);
-    }
+    let program_info = ProgramInfo {
+      program,
+      source_file: None,
+      tokens: None,
+      // TODO(@magurotuna): let it have SingleThreadedComments
+      comments: None,
```
bartlomieju, line 424: This should be fixed before landing?

src/rules.rs
```diff
@@ -86,15 +86,48 @@ pub mod triple_slash_reference;
...
+  /// Returns the unique code that identifies the rule
   fn code(&self) -> &'static str;
+
+  /// Returns the tags this rule belongs to, e.g. `recommended`
   fn tags(&self) -> &'static [&'static str] {
     &[]
   }
+
+  /// Returns the documentation string for this rule, describing what this rule is for with several
+  /// examples.
```
bartlomieju, line 130: Thanks for adding docstrings

bartlomieju (commented):
Thanks @magurotuna; this looks good to me, I only have a single comment (not sure if that's a blocker for landing)

src/linter.rs
magurotuna: Not really, I think it can be fixed when we try to refactor some rules that look at comments to use dprint-swc-ecma-ast-view, so it's okay for now.

magurotuna:
Thanks for reviewing. I'll resolve the conflict within a few hours.

bartlomieju (approved):

PR #605, opened 2021-01-31
refactor(no-self-assign): switch to `asset_lint_err!` macro

magurotuna:
Part of #431 and #330

- switches to `assert_lint_err!`
- adds hint
- uses `VisitAll` to handle nested assignments properly


bartlomieju (approved):
LGTM

PR #616, opened 2021-02-19
Upgrade swc_ecmascript to 0.23.1

### LINKED ISSUE #615: deno lint mis-parses private method parameters?
Run the following through `deno lint`:
```js
class T {
  #foo(x, y) {}
  #bar(x, y) {}
}
```

`deno lint` claims that I have declared `x` and `y` twice, as if they were were properties, instead of parameters.

magurotuna:
This PR upgrades swc_ecmascript to v0.23.1 which has a breaking change on `Ident`.
It consequently resolves #615, so I have added a new test case to make sure deno_lint doesn't complain the following code:

```ts
class T {
  #foo(x) {}
  #bar(x) {}
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,9 +46,9 @@ checksum = "afddf7f520a80dbf76e6f50a35bca42a2331ef227a28b3b6dc5c2e2338d114b1"
 
 [[package]]
 name = "ast_node"
-version = "0.7.1"
+version = "0.7.2"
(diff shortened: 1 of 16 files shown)


bartlomieju:
Cool, thanks @magurotuna, I'll wait before landing this PR until other crates that use `swc` in Deno also update to latest version.

bartlomieju (approved):

PR #617, opened 2021-02-21
chore: add missing copyright and new CI step to check it

magurotuna:
I found copyrights are missing in some files while browsing them.
I have added them and also created a new CI step that checks if all Rust files have copyrights.


kitsonk:
@magurotuna thanks for the missing copyrights.  As for the tooling in the CI, please see [the comment](https://github.com/denoland/deno/pull/8283#issuecomment-730356452) @ry made for a similar PR in Deno.

magurotuna:
@kitsonk 
Thanks for the reference!

> Also I think sometimes when we pull in external code but make very few changes, it makes sense to just maintain the original copyright header and not add our own.

Makes a lot of sense.
I retract the new CI step and script.

kitsonk (approved):
LGTM

PR #618, opened 2021-02-22
fix(no-unreachable): ignore type declarations

littledivy:
Fixes #619

PR #621, opened 2021-02-23
fix(no-unreachable): do not ignore TsEnum. Ignore TsModuleDecl.

littledivy:
This PR:

1. Let's the rule apply on `enum`s.
Enums transform to runtime objects. Therefore, they can be unreachable. 

2. Ignore `module` declarations.
```typescript
throw new Error(); // <-- will throw at runtime

declare module "SomeModule" {
  export function fn(): void;
} // <-- types can be imported.

let a = 1; // <-- this is rather unreachable.
```

My bad for not testing it thoroughly in #618


magurotuna (approved):
LGTM

> Enums transform to runtime objects. Therefore, they can be unreachable.

Good catch 👍

bartlomieju (approved):

PR #624, opened 2021-02-27
refactor(no-this-alias): switch to `assert_lint_err!` macro and cleanup

magurotuna:
Part of #431, #330

- switch to `assert_lint_err!` macro
- replace `Visit` with `VisitAll` to deal with nested cases
- introduce [`if_chain`](https://docs.rs/if_chain/1.0.1/if_chain/) crate so that chained `if` and `if let` won't have redundant indentations. `if_chain!` macro is simple but so useful that even clippy use it.

### DIFF
--- a/src/rules/no_this_alias.rs
+++ b/src/rules/no_this_alias.rs
@@ -1,10 +1,10 @@
 // Copyright 2020-2021 the Deno authors. All rights reserved. MIT license.
 use super::Context;
 use super::LintRule;
-use swc_ecmascript::ast::{ArrowExpr, Expr, Function, Pat, VarDecl};
+use swc_ecmascript::ast::{Expr, Pat, VarDecl};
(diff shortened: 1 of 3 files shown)


src/rules/no_this_alias.rs
```diff
@@ -41,52 +45,26 @@ impl<'c> NoThisAliasVisitor<'c> {
...
-              "no-this-alias",
-              "assign `this` to declare a value is not allowed",
-            );
-          }
+      if_chain! {
+        if let Some(init) = &decl.init;
+        if matches!(&**init, Expr::This(_));
+        if matches!(&decl.name, Pat::Ident(_));
+        then {
+          self.context.add_diagnostic(var_decl.span, CODE, MESSAGE);
```
bartlomieju, line 58: A bit magical, but oh well :)

bartlomieju (approved):
LGTM

PR #625, opened 2021-02-27
feat(no-unused-vars): add recommended tag to `no-unused-vars` and refine it

### LINKED ISSUE #598: Add `no-unused-imports`?
Are there plans to add this rule? I know personally it'd be amazing if `deno lint` could also throw errors/warnings on unused imports

magurotuna:
This PR adds `no-unused-vars` to the recommended set to enable it by default, and refines the rule so that it can work fine for deno_std.

Applying this rule to the current deno_std, we get 37 errors. It seems like there are no false positives.
I will soon open another PR to the deno_std repo to address these errors.

Part of #431 
Resolves #598

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,9 +114,9 @@ mod lint_tests {
   fn warn_unknown_rules() {
     let src = r#"
  // deno-lint-ignore some-rule
- function foo() {
+ function _foo() {
    // deno-lint-ignore some-rule-2 some-rule-3
-   let bar_foo = true
+   let _bar_foo = true
  }
       "#;
     let diagnostics = lint_recommended_rules(src, true, false);
@@ -130,7 +130,7 @@ mod lint_tests {
     let diagnostics = lint_recommended_rules(
       r#"
  // deno-lint-ignore some-rule
- function foo() {
+ function _foo() {
    // pass
  }
       "#,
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
Cool, thanks @magurotuna, this is very useful rule. Do you want to land it before #604?

magurotuna:
@bartlomieju 
I'd like #604 to be landed before this PR. Maybe there will be a conflict after #604 gets merged but I'll resolve it then. :-)

bartlomieju:
@magurotuna have you tried running this rule on `deno_std`? Were there any false positives?

magurotuna:
@bartlomieju I checked that before, and errors that happened in deno_std at that time were fixed by denoland/deno_std#764. In addition, the false positives I found were fixed in this PR.

However, running this rule on the current deno_std, the following error appears:

```
error[no-unused-vars]: `msg` is never used
   --> /home/yusuktan/Repos/github.com/magurotuna/deno_std/node/assert.ts:226:7
    |
226 |   let msg;
    |       ^^^
    |
    = help: If this is intentional, prefix it with an underscore like `_msg`
Found 1 problems
```

This diagnose looks a _true_ positive, so I'll fix it and open a PR to deno_std.

bartlomieju (approved):

PR #629, opened 2021-03-04
fix(no-undef):  Handle bindings of arrow expressions.

### LINKED ISSUE #596: no-undef gets triggered after destructure parameters on functions
I found a quite weird corner case here

If I were to create a function with the following structure

```ts
export const checkAdminAccess = async (
  { a }: { a: number },
  next: () => Promise<void>,
) => {}
```

Deno lint will warn me that next is not defined, however doing any changes to the function parameters (such as changing the first parameter for a non-destructurable one) will remove the warning

kdy1:
Closes #596.


bartlomieju (approved):
LGTM, thanks Kang

PR #632, opened 2021-03-08
refactor(no-setter-return): refine with ast-view and add sufficient tests

magurotuna:
Part of #330, #431

This PR does:

- rewrite with dprint_swc_ecma_ast_view
- refine the implementation to avoid false negatives
- add lots of tests (most come from ESLint)


magurotuna:
@lucacasonato do you have any idea about this error? https://github.com/denoland/deno_lint/pull/632/checks?check_run_id=2058739263#step:17:307

bartlomieju (approved):

PR #641, opened 2021-03-21
fix(no-unsafe-finally): deal with if-statements or functions in finally blocks, etc.

magurotuna:
Part of #330, #431

- rewrite the implementation with dprint-swc-ecma-ast-view
- add hint message
- refine the implementation so it can properly work for a) if-statements or functions in finally blocks, b) labeled statements
- add/port more tests referring to [ESLint](https://github.com/eslint/eslint/blob/master/tests/lib/rules/no-unsafe-finally.js)

### DIFF
--- a/src/rules/no_unsafe_finally.rs
+++ b/src/rules/no_unsafe_finally.rs
@@ -12,8 +12,25 @@ const HINT: &str = "Use of the control flow statements (`return`, `throw`, `brea
 
 #[derive(Display)]
 enum NoUnsafeFinallyMessage {
-  #[display(fmt = "Unsafe usage of {} statement", _0)]
-  UnsafeUsage(String),
+  #[display(fmt = "Unsafe usage of break statement")]
+  Break,
+  #[display(fmt = "Unsafe usage of continue statement")]
+  Continue,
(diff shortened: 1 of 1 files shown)


src/rules/no_unsafe_finally.rs
```diff
@@ -1,12 +1,38 @@
...
+use super::{Context, LintRule, ProgramRef};
+use crate::handler::{Handler, Traverse};
+use derive_more::Display;
+use dprint_swc_ecma_ast_view::{self as AstView, NodeTrait};
+use swc_common::{Span, Spanned};
 
 pub struct NoUnsafeFinally;
 
+const CODE: &str = "no-unsafe-finally";
+const HINT: &str = "Use of the control flow statements (`return`, `throw`, `break` and `continue`) in a `finally` block will most likely lead to undesired behavior. It's recommended to take a second look.";
```
bartlomieju, line 11: Please remove `It's recommended to take a second look.`. Also it seems that this line exceeds 120 chars, could it be split into multiline?

magurotuna: Okay, fixed it!

bartlomieju (approved):

PR #649, opened 2021-03-30
chore: remove no-undef from recommended set

bartlomieju:


sant123:
> Removed because it was complaining about stuff like process or DOM elements in mixed codebases. 

From @bartlomieju.

Thanks for the explanation!

PR #651, opened 2021-04-03
feat: implement new lint `no-deprecated-deno-api`

### LINKED ISSUE #650: Add `no-deprecated-deno`
This lint would error on all uses of the APIs that we are going to deprecate as per https://github.com/denoland/deno/issues/9795. The lint only has to appear on the most obvious cases (e.g. `new Deno.Buffer`, but not `const { Buffer } = Deno; new Buffer();`).

The lint rule should be `recommended`.

For Deno 1.9 the lint should highlight the following APIs:
- `Deno.Buffer`
- `Deno.readAll`
- `Deno.readAllSync`
- `Deno.writeAll`
- `Deno.writeAllSync`

Rule name is open for bikeshedding :-)

magurotuna:
Resolves #650 

This new lint rule warns the usage of the deprecated APIs in `Deno` namespace, such as:

```ts
new Deno.Buffer();
new Deno["Buffer"];
new Deno[`Buffer`];

Deno.readAll(reader);
Deno["readAll"](reader);
Deno[`readAll`](reader);
Deno.readAllSync(reader);
Deno["readAllSync"](reader);
Deno[`readAllSync`](reader);

Deno.writeAll(writer, data);
Deno["writeAll"](writer, data);
Deno[`writeAll`](writer, data);
Deno.writeAllSync(writer, data);
Deno["writeAllSync"](writer, data);
Deno[`writeAllSync`](writer, data);
```

Unfortunately, the following example isn't warned due to the limitation of scope analyzer:

```ts
function foo () {
  const Deno = 42;
}
Deno.readAll(reader);
```

cc @lucacasonato

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -27,6 +27,7 @@ pub mod no_constant_condition;
 pub mod no_control_regex;
 pub mod no_debugger;
 pub mod no_delete_var;
+pub mod no_deprecated_deno_api;
 pub mod no_dupe_args;
 pub mod no_dupe_class_members;
 pub mod no_dupe_else_if;
@@ -160,6 +161,7 @@ pub fn get_all_rules() -> Vec<Box<dyn LintRule>> {
     no_control_regex::NoControlRegex::new(),
     no_debugger::NoDebugger::new(),
     no_delete_var::NoDeleteVar::new(),
+    no_deprecated_deno_api::NoDeprecatedDenoApi::new(),
     no_dupe_args::NoDupeArgs::new(),
     no_dupe_class_members::NoDupeClassMembers::new(),
     no_dupe_else_if::NoDupeElseIf::new(),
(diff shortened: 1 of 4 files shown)


lucacasonato (commented):
Overall looks good. It would be great if the lint hint for each deprecated API would show the URL where the new API can be imported from. So something like this:

**Message**: Deno.Buffer is deprecated and scheduled for removal in Deno 2.0
**Hint**: use Buffer from https://deno.land/std/io/buffer.ts instead

lucacasonato:
Can you also add `Deno.iter` and `Deno.iterSync`? They have also been deprecated.

magurotuna:
@lucacasonato Done :)

bartlomieju (approved):
LGTM, thanks @magurotuna!

PR #660, opened 2021-04-09
ci: add gnu-tar on macOS as a workaround

magurotuna:
~~Not sure if it works so let's see it~~

This workaround seems to work.


magurotuna:
@bartlomieju PTAL

bartlomieju (approved):

PR #664, opened 2021-04-15
fix(no-unused-vars): properly mark JSX element names as used

### LINKED ISSUE #663: `no-unused-vars` doesn't support JSX
`no-unused-vars` emits an error for the following code, even though `React` is required.

```js
import React from "..."; // `React` is never used
export default <div />;
```

Not even when a function is explicitly used in a JSX tag does `no-unused-vars` understand what's going on: 

```js
import React from "..."; // `React` is never used
function Component() { return null; } // `Component` is never used
export default <Component />;
```

Unfortunately, I'll need to downgrade to Deno 1.8, where this rule wasn't enabled by default.

magurotuna:
Fixes #663

### DIFF
--- a/src/rules/no_unused_vars.rs
+++ b/src/rules/no_unused_vars.rs
@@ -1122,24 +1122,6 @@ import foo from 'foo';
 import bar from 'foo';
 export interface Bar extends foo.i18n<bar> {}
       ",
-
-      // TODO(kdy1): Unignore
-      //       "
-      // import { TypeA } from './interface';
-      // export const a = <GenericComponent<TypeA> />;
-      //       ",
-
-      // TODO(kdy1): Unignore
-      //       "
-      // const text = 'text';
-      // export function Foo() {
-      //   return (
-      //     <div>
-      //       <input type=\"search\" size={30} placeholder={text} />
-      //     </div>
-      //   );
-      // }
-      //       ",
       "
 import { observable } from 'mobx';
 export default class ListModalStore {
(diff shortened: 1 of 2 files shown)


src/rules/no_unused_vars.rs
```diff
@@ -83,6 +84,10 @@ struct Collector {
   /// Type of this should be hashset, but we don't have a way to
   /// restore hashset after handling bindings
   cur_defining: Vec<Id>,
+
+  /// Whether the code has JSX tag syntax or not.
+  /// If it does, imported `React` has to be treated as implicitly "used".
```
lucacasonato, line 89: The char that is used for this can be changed using the tsconfig file. When using `preact` the implicitly used vars are `h` and `Fragment`.

magurotuna: Oh I missed that point, thanks.
I'm not so familiar with this sort of thing, but from looking at the following documents, is it sufficient if just all of `React`, `h` and `Fragment` are marked as "used"? Or can developers choose arbitrary names they want to use as JSX factory function, even in the current version of TypeScript?

https://preactjs.com/guide/v10/typescript/
https://www.typescriptlang.org/docs/handbook/jsx.html

lucacasonato: Yes, they can choose arbitrary names. There are two configuration options for this in the tsconfig. `jsxFactory` and `jsxFragmentFactory`. The fragment factory name is used for JSX fragments (`<></>`), while the regular JSX factory is used for everything else (`<div></div>`, `<Foo />`). I suggest you have the implementor specify the `jsxFactory` and `jsxFragmentFactory` when configuring the linter.

magurotuna: That's unfortunate, things get complicated😣 but thanks for the clarification.

>  I suggest you have the implementor specify the jsxFactory and jsxFragmentFactory when configuring the linter.

That would be ideal, but however at the moment there's no capability of configuration, and moreover, I don't think deno_lint should be configurable in the way the behavior of rules can be changeable. Instead, I'd prefer to put a restriction on the implementors so that they _cannot_ choose arbitrary names for `jsxFactory` or `jsxFragmentFactory`. (Not quite sure if it's feasible though)

lucacasonato: The implementer can already get these values from the tsconfig, so it would not be "extra" configuration. @bartlomieju WDYT?

lucacasonato: Maybe we just have to disable this rule on JSX files.

magurotuna: > Maybe we just have to disable this rule on JSX files.

It sounds reasonable to me. False negative is more acceptable than false positive.

bartlomieju: I'm fine with both approaches. Right now `deno lint` does not take configuration flag, but we could potentially add it to sniff out values from tsconfig.

magurotuna: Just fixed so it will skip linting `.jsx` and `.tsx` files. Please take a look again.

bartlomieju: Looks ok to me

lucacasonato (commented):
LGTM now. @bartlomieju could you take a look too? Want to get this landed for 1.9.1.

bartlomieju (approved):

src/rules/no_unused_vars.rs
```diff
@@ -67,6 +74,10 @@ impl LintRule for NoUnusedVars {
   }
 }
 
+fn is_jsx_file(filename: &str) -> bool {
+  filename.ends_with(".jsx") || filename.ends_with(".tsx")
+}
```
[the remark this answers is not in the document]
magurotuna: I see. I'll open an issue for that.

PR #668, opened 2021-04-18
fix(prefer-const): handle switch statements properly

### LINKED ISSUE #665: `prefer-const` false positive when reassigned in `switch`
`prefer-const` emits an error, even though the variable is assigned in the `case`.

```js
let a; // 'a' is never reassigned
switch (Date.now() % 2) {
  case 0:
    a = 1;
}

console.log(a);
```

Changing the `switch` to an `if`works as expected.

magurotuna:
Fixes #665


bartlomieju (approved):
LGTM, thanks!

PR #675, opened 2021-04-24
fix(control-flow): handle `break` statements in loops properly

### LINKED ISSUE #674: Statement is unreachable after a conditional `throw`
Linting the following code:

```ts
const a = 'foo';
while(true){
    if(a == 'foo')
        break;
    throw new Error('bar');
}
console.log('foobar');
```

I get: 

```
console.log('foobar');
^^^^^^^^^^^^^^^^^^^^^^
```

The issue is, said statement is actually reachable.

The warning goes away if:
- the `throw` line is removed entirely
- the `throw` is nested in an `else` block.

magurotuna:
Fixes #674

### DIFF
--- a/src/control_flow.rs
+++ b/src/control_flow.rs
@@ -280,7 +280,7 @@ impl Analyzer<'_> {
         BlockKind::Label(label) => {
           if let Some(Some(id)) = &self.scope.found_break {
             if *id == label {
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

PR #677, opened 2021-05-01
refactor: use `all_comments` on ast-view

magurotuna:
This PR refactors several things:

- use `all_comments` I added to `dprint-swc-ecma-ast-view` in https://github.com/dprint/dprint-swc-ecma-ast-view/pull/18, instead of dealing with `trailing_comments` and `leading_comments` ourselves.
- move `Context` to its own module and make its fields private. Instead, it exposes getter methods to access the fields so we can control mutability more appropriately.
- add `dprint_swc_ecma_ast_view::Program` to `Context` as a field in order to call `all_comments` at any time. Accordingly, `Context` has a lifetime parameter `'view`, which indicates that `Context` cannot outlive the ast-view.
- remove `has_linted: bool` from the `Litner` struct. Instead the first parameters of `Linter::lint` and `Linter::lint_with_ast` have altered into `mut self` from `&mut self`. This way we can guarantee the actual linting occurs only once.

### DIFF
--- a/src/rules/default_param_last.rs
+++ b/src/rules/default_param_last.rs
@@ -18,7 +18,11 @@ impl LintRule for DefaultParamLast {
     "default-param-last"
   }
 
-  fn lint_program(&self, context: &mut Context, program: ProgramRef<'_>) {
+  fn lint_program<'view>(
+    &self,
+    context: &mut Context<'view>,
+    program: ProgramRef<'view>,
+  ) {
     let mut visitor = DefaultParamLastVisitor::new(context);
     match program {
       ProgramRef::Module(ref m) => m.visit_all_with(&DUMMY_NODE, &mut visitor),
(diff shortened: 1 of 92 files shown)


src/rules/default_param_last.rs
```diff
@@ -18,7 +18,11 @@ impl LintRule for DefaultParamLast {
     "default-param-last"
   }
 
-  fn lint_program(&self, context: &mut Context, program: ProgramRef<'_>) {
+  fn lint_program<'view>(
```
bartlomieju, line 21: A bit unlucky that all those signatures now require lifetime parameter

bartlomieju (approved):
This is a very nice refactor. I'm a bit unhappy for the changes to `lint_program` that adds required lifetime, but given how this restructures several major interfaces I think it's worth it.

LGTM,  very nice @magurotuna

Issue #666, opened 2021-04-16
no-self-assign error message is bad


lucacasonato:
```
error[no-self-assign]: length
  --> /Users/lucacasonato/projects/github.com/denoland/deno/op_crates/fetch/22_body.js:80:23
   |
80 |       second.length = second.length;
   |                       ^^^^^^^^^^^^^
   |
   = help: Mistake
```

This error message should be reworked. It need a better help text, and a better error text.

kamchy:
I'll try to look into this issue.

magurotuna:
@bartlomieju @lucacasonato This issue can be closed as it's been resolved by #681.

PR #681, opened 2021-05-05
Add missing fmt parameter in use-side of display macro

kamchy:
#666 issue is probably caused by missing macro parameter.  When added, the rule displays expected messages.


magurotuna (approved):
LGTM :+1: Nice fix!

bartlomieju (approved):
LGTM, thank you @kamchy

PR #685, opened 2021-05-07
chore: apply clippy suggestions

magurotuna:
This PR

- applies clippy suggestions that recommend we use `assert!` instead of `assert_eq!` in comparing with boolean
- updates the rustc version in README.md


bartlomieju (approved):
Thanks @magurotuna, it's strange that CI passed even though there are changes to clippy :/ am a bit confused as I run it locally as well.

bartlomieju:
Sounds good, thanks!

PR #686, opened 2021-05-08
fix(no-unused-vars): properly handle usage in arguments and `FnExpr`

### LINKED ISSUE #667: 1.9.0 - `no-unused-vars` diagnosis is wrong in cases where the value is accessible via a closure
1.9.0 complaints about values that will be found in the closure, here's an example:

```typescript
const i = setInterval(async () => {
  try {
    const v = await fn();
    if (v) {
      clearInterval(i);
      done(v);
    }
  } catch (_) {
    // ignore
  }
}, opts?.interval);
```

The linter will:

> (no-unused-vars) `i` is never used
>    const i = setInterval(async () => {

magurotuna:
Fixes #667


bartlomieju (approved):
LGTM, nice docs!

PR #689, opened 2021-05-09
chore: upgrade swc and dprint-swc-ecma-ast-view

magurotuna:
Upgrades `swc` and `dprint-swc-ecma-ast-view` and adjusts to the interface change in `dprint-swc-ecma-ast-view`.


ry (approved):
Thanks

Issue #690, opened 2021-05-09
Constructor overload arguments never used warnings


KyleJune:
For method overloads we don't show any warnings for arguments going unused but for constructor overloads we do.

```ts
export class Example {
  x: number;

  constructor(x: number);
  constructor(y: string);
  constructor(xOrY: number | string) {
    this.x = typeof xOrY === "string" ? parseInt(xOrY) : xOrY;
  }

  add(x: number): number;
  add(y: string): number;
  add(xOrY: number | string): number {
    const x: number = typeof xOrY === "string" ? parseInt(xOrY) : xOrY;
    return this.x * x;
  }
}
```

In the example above it warns about x and y not being used in the constructor overloads but doesn't warn about that in the add method's overloads. It should not be warning about these overloads being unused since they cannot be used directly.

```
$ deno lint --unstable
(no-unused-vars) `x` is never used
  constructor(x: number);
              ^^^^^^^^^
    at /mnt/c/Users/kylej/projects/deno/oauth2_server/example.ts:4:14

    hint: If this is intentional, prefix it with an underscore like `_x`

(no-unused-vars) `y` is never used
  constructor(y: string);
              ^^^^^^^^^
    at /mnt/c/Users/kylej/projects/deno/oauth2_server/example.ts:5:14

    hint: If this is intentional, prefix it with an underscore like `_y`

Found 2 problems
Checked 1 file
```

magurotuna:
I'll fix it.

PR #691, opened 2021-05-10
v0.5.0

bartlomieju:


magurotuna:
@bartlomieju I will start to work on fixing #690 in about 4 hours and I think it will take just a few hours to get it fixed. Maybe it would be good to have this bug fix included in 0.5.0.

bartlomieju:
> @bartlomieju I will start to work on fixing #690 in about 4 hours and I think it will take just a few hours to get it fixed. Maybe it would be good to have this bug fix included in 0.5.0.

@magurotuna sure, thank you! I'll wait for PR before releasing :)

PR #692, opened 2021-05-10
fix(no-unused-vars): handle overloaded constructors

### LINKED ISSUE #690: Constructor overload arguments never used warnings

magurotuna:
Fixes #690

### DIFF
--- a/src/rules/no_unused_vars.rs
+++ b/src/rules/no_unused_vars.rs
@@ -396,14 +396,6 @@ impl Visit for Collector {
       declarator.init.visit_with(declarator, a);
     });
   }
-
-  fn visit_constructor(&mut self, c: &Constructor, _: &dyn Node) {
-    if c.body.is_none() {
-      return;
-    }
-
-    c.visit_children_with(self);
-  }
 }
 
 fn get_id(r: &TsEntityName) -> Id {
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, thanks!

PR #694, opened 2021-05-11
doc(use_isnan): add doc

togami2864:
Adds the documentation for `use_isnan` rule.
#159

### DIFF
--- a/src/rules/use_isnan.rs
+++ b/src/rules/use_isnan.rs
@@ -63,13 +63,13 @@ Therefore, this rule makes you use the `isNaN()` or `Number.isNaN()` to judge th
 ### Invalid:
 ```typescript
 if (foo == NaN) {
-    // ...
+  // ...
 }
 ```
 
 ```typescript
 if (foo != NaN) {
-    // ...
+  // ...
 }
 ```
 
--- a/src/rules/use_isnan.rs
+++ b/src/rules/use_isnan.rs
@@ -74,9 +74,16 @@ if (foo != NaN) {
 ```
 
 ```typescript
-switch (foo) {
-  case NaN:
+switch (NaN) {
+  case foo:
     // ...
+}
+```
+
+```
+switch (foo) {
+    case NaN:
+      // ...
 }
 ```
 
@@ -89,13 +96,6 @@ if (isNaN(foo)) {
 
 ```typescript
 if (!isNaN(foo)) {
-    // ...
-}
-```
-
-```typescript
-switch (foo) {
-  case isNaN(foo):
     // ...
 }
 ```"#
--- a/src/rules/use_isnan.rs
+++ b/src/rules/use_isnan.rs
@@ -82,21 +82,21 @@ switch (NaN) {
 
 ```
 switch (foo) {
-    case NaN:
-      // ...
+  case NaN:
+    // ...
 }
 ```
 
 ### Valid:
 ```typescript
 if (isNaN(foo)) {
-    // ...
+  // ...
 }
 ```
 
 ```typescript
 if (!isNaN(foo)) {
-    // ...
+  // ...
 }
 ```"#
   }
--- a/src/rules/use_isnan.rs
+++ b/src/rules/use_isnan.rs
@@ -49,6 +49,43 @@ impl LintRule for UseIsNaN {
       ProgramRef::Script(ref s) => visitor.visit_script(s, &DUMMY_NODE),
     }
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows comparisons to 'NaN'.
+
+Because `NaN` is unique in JavaScript by not being equal to anything, including itself, the results of comparisons to `NaN` are confusing:
+
+- `NaN === NaN` or `NaN == NaN` evaluate to false
+- `NaN !== NaN` or `NaN != NaN` evaluate to true
+
+Therefore, this rule makes you use the isNaN() or Number.isNaN() to judge the value is `NaN` or not.
+
+### Invalid:
+```typescript
+if (foo == NaN) {
+    // ...
+}
+```
+
+```typescript
+if (foo != NaN) {
+    // ...
+}
+```
+
+### Valid:
+```typescript
+if (isNaN(foo)) {
+    // ...
+}
+```
+
+```typescript
+if (!isNaN(foo)) {
+    // ...
+}
+```"#
+  }
 }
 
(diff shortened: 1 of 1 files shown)


src/rules/use_isnan.rs
```diff
@@ -49,6 +49,43 @@ impl LintRule for UseIsNaN {
       ProgramRef::Script(ref s) => visitor.visit_script(s, &DUMMY_NODE),
     }
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows comparisons to 'NaN'.
```
magurotuna, line 54:
```suggestion
    r#"Disallows comparisons to `NaN`.
```

Backticks would be nicer here because it will be rendered as a `<code>` tag so that readers can easily notice that it is a pre-defined value in JavaScript.

src/rules/use_isnan.rs
```diff
@@ -49,6 +49,43 @@ impl LintRule for UseIsNaN {
...
+
+Because `NaN` is unique in JavaScript by not being equal to anything, including itself, the results of comparisons to `NaN` are confusing:
+
+- `NaN === NaN` or `NaN == NaN` evaluate to false
+- `NaN !== NaN` or `NaN != NaN` evaluate to true
+
+Therefore, this rule makes you use the isNaN() or Number.isNaN() to judge the value is `NaN` or not.
```
magurotuna, line 61:
```suggestion
Therefore, this rule makes you use the `isNaN()` or `Number.isNaN()` to judge the value is `NaN` or not.
```

Ditto. Wrapping these in backticks will get rendered pretty well.

src/rules/use_isnan.rs
```diff
@@ -49,6 +49,43 @@ impl LintRule for UseIsNaN {
...
+if (foo == NaN) {
+    // ...
+}
+```
+
+```typescript
+if (foo != NaN) {
+    // ...
+}
+```
```
magurotuna, line 74: Could you add other invalid examples like `switch(NaN)` and `case NaN`? I think this should make the document much more helpful.

magurotuna (changes requested):
Thank you very much @togami2864!
Overall looks good to me. I left just a few suggestions.

togami2864:
Thanks. I fixed it.

src/rules/use_isnan.rs
```diff
@@ -49,6 +49,57 @@ impl LintRule for UseIsNaN {
...
+if (!isNaN(foo)) {
+    // ...
+}
+```
+
+```typescript
+switch (foo) {
+  case isNaN(foo):
+    // ...
+}
```
magurotuna, line 100: This looks a little bit strange in semantics, because `isNaN(foo)` evaluates to boolean but `foo` is number, thus `foo` is never equal to  `isNaN(foo)`.
I think we need to use `if` instead of `switch` in order to convert the `case NaN:` example into a valid one. So we could probably remove this example.

Instead, how about adding the following to invalid examples? It is also invalid and should be converted into `if` statement, so it would be worth presenting to readers.

```ts
switch (NaN) {
  case foo:
    // ...
}
```

magurotuna (changes requested):

src/rules/use_isnan.rs
togami2864: Definitely. I misunderstood.
I'll remove  previous changes and add `switch(NaN)`  `case NaN`  as invalid example.

src/rules/use_isnan.rs
```diff
@@ -49,6 +49,57 @@ impl LintRule for UseIsNaN {
...
+
+- `NaN === NaN` or `NaN == NaN` evaluate to false
+- `NaN !== NaN` or `NaN != NaN` evaluate to true
+
+Therefore, this rule makes you use the `isNaN()` or `Number.isNaN()` to judge the value is `NaN` or not.
+
+### Invalid:
+```typescript
+if (foo == NaN) {
+    // ...
```
magurotuna, line 66:
```suggestion
  // ...
```

We adopt two-space indentation. It would be great to follow this convention.

src/rules/use_isnan.rs
```diff
@@ -49,6 +49,57 @@ impl LintRule for UseIsNaN {
...
+  case foo:
+    // ...
+}
+```
+
+```
+switch (foo) {
+    case NaN:
+      // ...
+}
```
magurotuna, line 87:
```suggestion
switch (foo) {
  case NaN:
    // ...
}
```

Ditto

src/rules/use_isnan.rs
```diff
@@ -49,6 +49,57 @@ impl LintRule for UseIsNaN {
...
+```
+
+### Valid:
+```typescript
+if (isNaN(foo)) {
+    // ...
+}
```
magurotuna, line 94:
```suggestion
if (isNaN(foo)) {
  // ...
}
```

src/rules/use_isnan.rs
```diff
@@ -49,6 +49,57 @@ impl LintRule for UseIsNaN {
...
+```
+
+```typescript
+if (!isNaN(foo)) {
+    // ...
+}
```
magurotuna, line 100:
```suggestion
if (!isNaN(foo)) {
  // ...
}
```

magurotuna (changes requested):
Thanks for refining the examples, now the contents look pretty good!

Sorry for nitpicking, but I'd like you to fix the last few things about indentation.

src/rules/use_isnan.rs
magurotuna: Maybe you forgot this :)

src/rules/use_isnan.rs
```diff
@@ -49,6 +49,57 @@ impl LintRule for UseIsNaN {
...
+```typescript
+if (foo == NaN) {
+    // ...
+}
+```
+
+```typescript
+if (foo != NaN) {
+    // ...
+}
```
magurotuna, line 73:
```suggestion
if (foo != NaN) {
  // ...
}
```

and this (I also missed it 😄 )

togami2864:
I made a lot of tiny mistakes😱 I'll be more careful.
Thank you for your reviewing.

magurotuna (approved):
@togami2864 
No worries, now it's perfect! Thanks so much for touching it up:+1:
I'd like @bartlomieju to do the final review.

bartlomieju (approved):
LGTM as well, thank you @togami2864!

PR #696, opened 2021-05-15
chore: add GitHub issue templates

magurotuna:
This PR adds two GitHub issue templates, "Blank Issue" and "Bug Report".

Issue templates should be really helpful not only for visitors (especially first-time visitors) to easily file issues but also for developers because we will be able to get uniform, sufficient information from a filled-out template, which will allow us to debug and find the cause.


bartlomieju (approved):
Thanks!

PR #702, opened 2021-05-20
fix(ban-ts-comment): enrich hint to include concrete example

### LINKED ISSUE #701: Enrich `ban-ts-comment` message/hint
Users who want to suppress the TypeScript checker may get puzzled about how they can get along with the TS directives and the diagnosis from deno_lint. To clearly learn how, they need to visit lint.deno.land and to look at the example of `ban-ts-coment`.
https://lint.deno.land/#ban-ts-comment

To address this situation, it would be great to entich the rule's message and/or hint. Particularly, the current hint doesn't mention any examples about how the reason should be written. Concrete examples would be really helpful.

magurotuna:
Fixes #701

And also refactors the code a little bit.

### DIFF
--- a/src/rules/ban_ts_comment.rs
+++ b/src/rules/ban_ts_comment.rs
@@ -18,13 +18,42 @@ use swc_common::Span;
 /// - ts-nocheck: allowed with comment
 pub struct BanTsComment;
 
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, very nice, thank you!

PR #703, opened 2021-05-21
doc(no-unreachable): add doc

togami2864:
Adds the documentation for `no-unreachable` rule.
#159

### DIFF
--- a/src/rules/no_unreachable.rs
+++ b/src/rules/no_unreachable.rs
@@ -93,21 +93,6 @@ function foo() {
     return 1;
   }
 }
-```
-
-```typescript
-function bar() {
-  return x;
-  var x;
-}
-```
-
-```typescript
-switch (foo) {
-  case 1:
-    break;
-    var x;
-}
 ```"#
   }
 }
--- a/src/rules/no_unreachable.rs
+++ b/src/rules/no_unreachable.rs
@@ -39,7 +39,7 @@ impl LintRule for NoUnreachable {
   fn docs(&self) -> &'static str {
     r#"Disallows the unreachable code after the control flow statements.
 
-Because the control flow statements(`return`,`throw`, `break` and `continue`) unconditionally exit a block of code, any statements after them cannot be executed.
+Because the control flow statements (`return`, `throw`, `break` and `continue`) unconditionally exit a block of code, any statements after them cannot be executed.
 
 ### Invalid:
 ```typescript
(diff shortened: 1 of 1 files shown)


src/rules/no_unreachable.rs
```diff
@@ -35,6 +35,81 @@ impl LintRule for NoUnreachable {
...
+}
+```
+
+```typescript
+switch (foo) {
+  case 1:
+    break;
+    var x;
+}
+```"#
```
magurotuna, line 111: The reason why `var x;` in these examples are valid may be quite unclear to people who are not so experienced. Maybe it's great to add some comment, like...

```suggestion
```typescript
function bar() {
  return x;
  // Because it's declared with `var`, this `x` gets hoisted.
  var x;
}
```

```typescript
switch (foo) {
  case 1:
    break;
    // Because it's declared with `var`, this `x` gets hoisted.
    var x;
}
```"#
```

(oh the suggestion is rendered awfully due to nested code blocks...)

But in this case, I would rather remove these two examples, because it's so rare that we use `var` declarations in today's JavaScript/TypeScript. As everyone knows, we should always use `let` or `const` instead.

magurotuna (commented):
Thanks!
I have one personal opinion on the valid cases including variable hoisting. What do you think about it?

src/rules/no_unreachable.rs
togami2864: True. These two examples are not required.(also, support for ie11 is going to be over....)
I'll remove them.

src/rules/no_unreachable.rs
```diff
@@ -35,6 +35,66 @@ impl LintRule for NoUnreachable {
       ProgramRef::Script(ref s) => visitor.visit_script(s, &DUMMY_NODE),
     }
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows the unreachable code after the control flow statements.
+
+Because the control flow statements(`return`,`throw`, `break` and `continue`) unconditionally exit a block of code, any statements after them cannot be executed.
```
magurotuna, line 42:
```suggestion
Because the control flow statements (`return`, `throw`, `break` and `continue`) unconditionally exit a block of code, any statements after them cannot be executed.
```

magurotuna (changes requested):
I found another minor nit :)

magurotuna (approved):
LGTM :100: Nice contribution!

lucacasonato (approved):
Thanks!

PR #707, opened 2021-05-27
feat: implement new lint `no-invalid-triple-slash-reference`

### LINKED ISSUE #706: Add lint rule to avoid bad triple slash references
We've had a lot of confusing around triple slash references using the wrong property. [For example](https://github.com/esbuild/deno-esbuild/blob/28b04eedb1583a7c1c55aa83fc678c5b4a783c63/mod.js#L1) 
```
/// <reference path="./mod.d.ts" />
```
should actually be
```
/// <reference types="./mod.d.ts" />
```
Ideally deno lint would complain about bad triple slash references.

magurotuna:
Closes #706

### DIFF
--- a/src/rules/no_invalid_triple_slash_reference.rs
+++ b/src/rules/no_invalid_triple_slash_reference.rs
@@ -1,0 +1,223 @@
+// Copyright 2020-2021 the Deno authors. All rights reserved. MIT license.
+use super::{Context, LintRule, ProgramRef};
+use derive_more::Display;
+use once_cell::sync::Lazy;
+use regex::Regex;
+use swc_common::comments::{Comment, CommentKind};
+use swc_common::Span;
+
+pub struct NoInvalidTripleSlashReference;
+
+const CODE: &str = "no-invalid-triple-slash-reference";
+
+#[derive(Display)]
+enum NoInvalidTripleSlashReferenceMessage {
+  #[display(
+    fmt = "triple-slash references with `path` have no effect in Deno"
+  )]
+  Path,
+}
+
+#[derive(Display)]
+enum NoInvalidTripleSlashReferenceHint {
+  #[display(
+    fmt = r#"Replace `path` with `types`, like `/// <reference types="{}" />`"#,
+    _0
+  )]
+  Replace(String),
+}
+
+impl LintRule for NoInvalidTripleSlashReference {
+  fn new() -> Box<Self> {
+    Box::new(NoInvalidTripleSlashReference)
+  }
+
+  fn tags(&self) -> &'static [&'static str] {
+    &["recommended"]
+  }
+
+  fn code(&self) -> &'static str {
+    CODE
+  }
+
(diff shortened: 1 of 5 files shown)


src/rules/no_invalid_triple_slash_reference.rs
```diff
@@ -0,0 +1,223 @@
...
+
+/// Returns `Some` if the comment should be reported.
+fn check_comment(comment: &Comment) -> Option<(Span, String)> {
+  if comment.kind == CommentKind::Block {
+    return None;
+  }
+
+  static INVALID_TSR_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r#"^/\s*<reference\s*path\s*=\s*["|'](?P<path>.*)["|']\s*/>"#)
+      .unwrap()
```
ry, line 104: Rather than blocking only "reference path", could it be an allow-list (presumably we want to allow only "reference types=" and "reference lib=") so that things like 
```
/// <reference foo="bar">
```
would also trigger this rule?

ry (commented):
Awesome! Thank you!

src/rules/no_invalid_triple_slash_reference.rs
kitsonk: I agree with @ry, we only want to allow `/// <reference types="..." />` in JavaScript/JSX files (we shouldn't have the diagnostic for TypeScript/TSX files).

Also we should be consistent with the logic that is in the `module_graph`:

https://github.com/denoland/deno/blob/59237d195f6afa0d8927128751ba5ff4562d7cce/cli/module_graph.rs#L204-L224

magurotuna: allow-list sounds good.
And I've noticed that my wording in the doc, "triple-slash references with `path` have no effect in Deno", is likely wrong... actually the `path` directive allows adding a dependency to a module in Deno as well.

So my current understanding is that this lint rule should be triggered in cases such as:

- triple-slash references in which none of the allowed directives (`path`, `types`, `lib`) is specified (i.e. invalid format)
- triple-slash references with `path` iff the file is JS or JSX (i.e. the `path` directive is used unintentionally)

I'm fixing it right now.

kitsonk: To be clear, in TypeScript files (TypeScript/TSX) triple-slash directives are allowed and work.

In JavaScript files (JSX, TSX), only the `types` directive is allowed.

magurotuna: Ah OK, I misunderstood the second point. Thanks for clarifying :)

magurotuna: @ry @kitsonk 
Fixed. Now, when the target file is JS-like, this rule tries to detect the directives other than `types`, and when it's not JS-like, it checks if there is an invalid format of triple-slash reference.
Please take a look again 😃

ry (approved):
LGTM - this will really help reduce TS errors - thanks!

PR #708, opened 2021-05-29
fix(no-unused-vars): handle type-only imports correctly

### LINKED ISSUE #705: no-unused-vars warning when variable and type have the same name
Linter thinks type is unused if the variable name is the same as the type name
### Lint Name
`no-unused-vars`

### Code Snippet
a.ts
```ts
import type { Filters } from "./types.ts";
export function audioFilters(...Filters: Filters[]): void {
  for (const filter of Filters) {
    console.log(filter);
    return;
  }
}
```

types.ts
```ts
export interface Filters {
  complex?: boolean;
  filterName: string;
  options: Record<string, string | number>;
}
```
### Expected Result

No warnings

### Actual Result

linter keeps warning that I'm not using `Filters`

### Additional Info

this happens when the type name is the same as the variable name\
using `filters` instead of `Filters` as variable name fixes this for me

### Version

deno 1.10.2 (release, x86_64-unknown-linux-gnu)
v8 9.1.269.27
typescript 4.2.2

magurotuna:
Fixes #705 by upgrading swc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1155,9 +1155,9 @@ dependencies = [
 
 [[package]]
 name = "swc_ecma_transforms_base"
-version = "0.15.4"
+version = "0.15.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf55207e36eaa746e474bbf57ae2ee7455626f6711d0e7c14065bfc13bc59ed5"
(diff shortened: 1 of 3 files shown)


bartlomieju:
Thanks, before landing I need to upgrade SWC in other related projects to make sure we don't have duplicate versions.

bartlomieju (approved):
LGTM

PR #709, opened 2021-05-31
fix: show version from Cargo.toml

brandly:
Before, `--version` is shown in the `--help` text, but the output doesn't list a version number.

```
$ dlint --version
dlint 
```

clap can pull the version from `Cargo.toml` [as shown here](https://github.com/clap-rs/clap/blob/master/examples/09_auto_version.rs). With this change, the version is shown as expected.

```
$ dlint --version
dlint 0.5.0
```


magurotuna (approved):
Good catch, LGTM :+1:

kitsonk (approved):
LGTM

PR #715, opened 2021-06-04
doc(no_mixed_spaces_and_tabs): add doc

togami2864:
Adds the documentation for  no_mixed_spaces_and_tabs rule.
#159


src/rules/no_mixed_spaces_and_tabs.rs
```diff
@@ -89,6 +89,34 @@ impl LintRule for NoMixedSpacesAndTabs {
       }
     }
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows mixed spaces and tabs for indentation.
+
+Most code rules need to use of tabs or spaces for indentation. Therefore, if a line of code is indented with both tabs and spaces, it will usually result in an error.
```
magurotuna, line 96: When reading this sentence I got a little bit confused. 

> Most code rules need to use ...

When I see "rule" I think of it as something that the syntax of a language forces developers to follow, but the language itself doesn't force it. So how about using "conventions" instead? It sounds better when referring to code styles.

> use of tabs or spaces for indentation

`of` looks unnecessary here.

> Most code rules need to use of tabs or spaces for indentation.

This whole sentence is intended to state "Most code conventions out there require developers to use _either_ tabs or spaces for indentation" right? If it's correct, let's not use "need", and instead use a verb with more strongness, like "require" or "command". Additionally, what we should clarify here is that _tabs and spaces cannot be used at the same time_, so it is much better to insert "either" into the sentence IMO.

> it will usually result in an error.

To me this phrase sounds like some runtime error will happen, but actually won't, which can be confirmed by running the following snippet where we'll get no error.

```ts
function foo() {
	 /* <---- tab and space mixed */console.log(42);
}　

foo();
```

So I'd prefer to choose `mistake` over `error` here, like `it's most likely a mistake of a developer.`

src/rules/no_mixed_spaces_and_tabs.rs
```diff
@@ -89,6 +89,34 @@ impl LintRule for NoMixedSpacesAndTabs {
...
+
+### Invalid:
+
+```typescript
+function add(x, y) {
+      return x + y;
+}
```
magurotuna, line 103: It would be much nicer to add a comment that describes the line of `return` is indented with both a tab and spaces, because it's hard to recognize by just looking at the rendered snippet.

And it looks like the line of `return` is actually indented with only spaces, doesn't have tab character. Your intention may be like the following?

```ts
function add(x: number, y: number) {
	  return x + y; // indented with a tab + two spaces
}
```

(type annotations for parameters would be fine as well since it's TypeScript)

src/rules/no_mixed_spaces_and_tabs.rs
```diff
@@ -89,6 +89,34 @@ impl LintRule for NoMixedSpacesAndTabs {
...
+```
+
+```typescript
+    let x = 5,
+        y = 7;
```
magurotuna, line 108: Ditto, this also looks indented with only spaces, and should be better if there is a comment.

src/rules/no_mixed_spaces_and_tabs.rs
```diff
@@ -89,6 +89,34 @@ impl LintRule for NoMixedSpacesAndTabs {
...
+}
+```
+
+### Valid:
+
+```typescript
+function add(x, y) {
+    return x + y;
+}
```
magurotuna, line 117:
```suggestion
function add(x: number, y: number) {
  return x + y;
}
```

Would be good to follow `deno fmt` style, and to add type annotations to parameters.

src/rules/no_mixed_spaces_and_tabs.rs
```diff
@@ -89,6 +89,33 @@ impl LintRule for NoMixedSpacesAndTabs {
       }
     }
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows mixed spaces and tabs for indentation.
+
+Most code conventions require either tabs or spaces be used for indentation. Therefore, if a line of code is indented with both tabs and spaces, it's most likely a mistake of a developer.
```
magurotuna, line 96: Nice 👍

src/rules/no_mixed_spaces_and_tabs.rs
```diff
@@ -89,6 +89,33 @@ impl LintRule for NoMixedSpacesAndTabs {
...
+```typescript
+function add(x: number, y: number) {
+    return x + y; // indented with a tab + two spaces
+}
+```
+
+```typescript
+    let x = 5, // indented with a tab + two spaces
+        y = 7;
+```
```
magurotuna, line 109: Still seems like these examples consist of only spaces, right? Would be good to actually have a tab character and a space character.

src/rules/no_mixed_spaces_and_tabs.rs
```diff
@@ -89,6 +89,33 @@ impl LintRule for NoMixedSpacesAndTabs {
...
+    r#"Disallows mixed spaces and tabs for indentation.
+
+Most code conventions require either tabs or spaces be used for indentation. Therefore, if a line of code is indented with both tabs and spaces, it's most likely a mistake of a developer.
+
+### Invalid:
+
+```typescript
+function add(x: number, y: number) {
+          return x + y; // indented with a tab + two spaces
+}
```
magurotuna, line 103: Thanks for the fix again, but unfortunately, it still seems like it consists of spaces only. Perhaps your editor config forces you to use spaces (i.e. soft tabs), instead of hard tabs. If my guess is correct, I'd recommend you see the config of your editor and set it to hard-tabs (temporarily). Or please feel free to apply my suggestion below by clicking "Commit suggestion" button 👍 

```suggestion
function add(x: number, y: number) {
	  return x + y; // indented with a tab + two spaces
}
```

src/rules/no_mixed_spaces_and_tabs.rs
```diff
@@ -89,6 +89,33 @@ impl LintRule for NoMixedSpacesAndTabs {
...
+```
+
+```typescript
+          let x = 5, // indented with a tab + two spaces
+              y = 7;
```
magurotuna, line 108:
```suggestion
	let x = 5, // indented with a tab
	    y = 7; // indented with a tab + four spaces
```

magurotuna (changes requested):

bartlomieju:
Thank you for the PR @togami2864! I'll merge when we get LGTM from @magurotuna

togami2864:
Sorry for replying late @magurotuna. I fixed it.

magurotuna (approved):
Perfect! LGTM

CI fails but it should be addressed by #740. No worries.

bartlomieju (approved):
LGTM, thank you @togami2864!

PR #721, opened 2021-06-07
fix(control-flow): mark exprs as `may_throw = true` even when scope.end is already set

### LINKED ISSUE #716: False positive: `no-unreachable` unexpectedly flags statements inside catch clause
### Lint Name

no-unreachable

### Code Snippet

```ts
export function foo(trueOrFalse: boolean) {
  if (trueOrFalse) {
    // noop
  } else {
    // noop
  }
  try {
    bar();
    return 42;
  } catch (err) {
    console.error(err);
  }
}
```

### Expected Result

No error happens

### Actual Result

```
    console.error(err);
    ^^^^^^^^^^^^^^^^^^^
    at /home/yusuktan/Repos/github.com/magurotuna/deno_lint/maguro.ts:11:4

Found 1 problem
Checked 1 file
```

### Additional Info

If the if-else block is removed this false positive goes away.

I found this bug when browsing oak's code
https://github.com/oakserver/oak/blob/51c02fc1979c8e2f921f91099a390ecc4727fb7a/application.ts#L428-L433

### Version

deno 1.10.3 (release, x86_64-unknown-linux-gnu)
v8 9.1.269.27
typescript 4.2.2

magurotuna:
This PR fixes a false positive that occurs in situations where a try-catch block follows a if-else block. 
In addition, I separated `control_flow.rs` into two files because there are so many test cases, making it less easy for us to navigate the implementation and the tests in the editor.
This refactor makes the essential diff hard to find, so I'm going to add some comments to spot where it is.

Fixes #716


src/control_flow/analyze_test.rs
```diff
@@ -0,0 +1,888 @@
...
+  assert_flow!(flow, 30, false, Some(End::Continue)); // BlockStmt of if
+  assert_flow!(flow, 38, false, Some(End::Continue)); // BlockStmt of else
+  assert_flow!(flow, 43, false, Some(End::Continue)); // try stmt
+  assert_flow!(flow, 47, false, Some(End::forced_return())); // BlockStmt of try
+  assert_flow!(flow, 53, false, None); // `bar();`
+  assert_flow!(flow, 64, false, Some(End::forced_return())); // return stmt
+  assert_flow!(flow, 79, false, Some(End::Continue)); // catch clause
+  assert_flow!(flow, 91, false, Some(End::Continue)); // BlockStmt of catch
+  assert_flow!(flow, 97, false, None); // `console.error(err);`
+}
```
magurotuna, line 888: Added a test to ensure the control-flow analyzer works as expected.

src/rules/no_unreachable.rs
```diff
@@ -474,6 +474,23 @@ do {
...
+    // noop
+  }
+  try {
+    bar();
+    return 42;
+  } catch (err) {
+    console.error(err);
+  }
+}
+      "#,
```
magurotuna, line 493: Also added a direct regression test.

src/control_flow/mod.rs
```diff
@@ -0,0 +1,793 @@
...
+  fn visit_stmts(&mut self, stmts: &[Stmt], _: &dyn Node) {
+    for stmt in stmts {
+      self.visit_stmt_or_block(stmt);
+    }
+  }
+
+  fn visit_expr(&mut self, n: &Expr, _: &dyn Node) {
+    n.visit_children_with(self);
+
+    if matches!(self.scope.end, None | Some(End::Continue)) {
```
magurotuna, line 394: The only change for the implementation is here.
Previously this line looked like `if self.scope.end.is_none() {`. Actually we should set `self.scope.may_throw` to `true` in this case too.

bartlomieju (approved):
LGTM, no comments to implementation, provided tests are self-explanatory 👍

PR #722, opened 2021-06-10
refactor(ban-ts-comment): reimplement with ast-view

magurotuna:
Reimpmements `ban-ts-comment` with ast-view. As you can see, almost nothing has changed.


dsherret (approved):
That was easy! 😄

LGTM

PR #723, opened 2021-06-10
refactor(ban-types): reimplement with ast-view

magurotuna:
Reimplements `ban-types` with ast-view, and adds hints.

### DIFF
--- a/src/rules/ban_types.rs
+++ b/src/rules/ban_types.rs
@@ -99,7 +99,7 @@ type and the misunderstood `Object` type.
 There are very few situations where primitive wrapper objects are desired and
 far more often a mistake was made with the case of the primitive type.  You also
 cannot assign a primitive wrapper object to a primitive leading to type issues
-down the line
+down the line.
 
 With `Function`, it is better to explicitly define the entire function
 signature rather than use the non-specific `Function` type which won't give you
--- a/src/rules/ban_types.rs
+++ b/src/rules/ban_types.rs
@@ -26,7 +26,7 @@ impl BannedType {
   fn as_message(&self) -> &'static str {
     use BannedType::*;
     match *self {
-      String | Boolean | Number | Symbol => "For consistency, it is recommended to use the lower-case primitive",
+      String | Boolean | Number | Symbol => "The corresponding lower-case primitive should be used",
       Function => "This provides no type safety because it represents all functions and classes",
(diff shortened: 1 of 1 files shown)


src/rules/ban_types.rs
```diff
@@ -1,16 +1,71 @@
...
+      Number => "Use `number` instead",
+      Symbol => "Use `symbol` instead",
+      Function => "Define the function shape explicitly",
+      CapitalObject => {
+        r#"If you want a type meaning "any object", use `Record<string, unknown>` instead. Or if you want a type meaning "any value", you probably want `unknown` instead."#
+      }
+      LowerObject => "Use `Record<string, unknown>` instead",
+    }
+  }
+}
```
dsherret, line 50: Nice!

src/rules/ban_types.rs
```diff
@@ -1,16 +1,71 @@
...
+  Function,
+  CapitalObject,
+  LowerObject,
+}
+
+impl BannedType {
+  fn as_message(&self) -> &'static str {
+    use BannedType::*;
+    match *self {
+      String | Boolean | Number | Symbol => "For consistency, it is recommended to use the lower-case primitive",
```
dsherret, line 29: Nitpick: Technically it's for more than just consistency.

https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#number-string-boolean-symbol-and-object

src/rules/ban_types.rs
```diff
@@ -1,16 +1,71 @@
...
+      Function => "This provides no type safety bacause it represents all functions and classes",
```
dsherret, line 30: typo: `bacause` -> `because`

magurotuna: Fixed in bf81a85 ;-)

dsherret (approved):
LGTM. Thanks @magurotuna! It's nice that more of these are getting switched over 🙂

src/rules/ban_types.rs
magurotuna: Thanks for the reference, I didn't read that. Previous wording was based on [the comment in a typescript-eslint example](https://github.com/typescript-eslint/typescript-eslint/blob/79b510c00f25d76daf741936682e1222b0c994b5/packages/eslint-plugin/docs/rules/ban-types.md#examples), but it would be good to refer to the TS handbook :) e727b9c

PR #724, opened 2021-06-11
refactor(ban-untagged-ignore): reimplement with ast-view

magurotuna:
Reimplements `ban-untagged-ignore` with ast-view. It's very easy refactoring like #722 :)


bartlomieju (approved):
LGTM

PR #726, opened 2021-06-11
feat(handler): add `TraverseFlow` to control children traversing

magurotuna:
This PR adds a new enum `TraverseFlow` that indicates whether the node's children should be traversed or not.

In the current implementation, `Traverse::traverse` method will always traverse all the nodes existing in the AST. But I noticed that some lint rules (e.g. `camelcase`) could be implemented easily and simply if we could stop the traversing when some conditions are met.
Using `TraverseFlow` each handler method can control it. This is inspired by [std::ops::ControlFlow](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html).

The code diff is huge, but most of them are boilerplate; each handler method returns `TraverseFlow::Continue` by default, and in `Traverse::traverse` method, a value returned from handler (which is `TraverseFlow`) is stored into `flow` variable which is later checked before traversing children.


bartlomieju (commented):
I'm not a fan that you need to return `TraverseFlow::Continue` manually from each method. I see that this may be useful, but since it's gonna be rarely used could it be a method that signals to `Handler` to stop traversing?

magurotuna:
> since it's gonna be rarely used could it be a method that signals to Handler to stop traversing?

Makes much sense, but I can't come up with how to actually implement that...
If we want to signal to stop further traversing, we have to store a flag somewhere - the problem is, where do we store it?

One naive way would be like this:

```rs
// handler.rs

thread_local! {
  static STOP_TRAVERSE: RefCell<bool> = RefCell::new(false);
}

// `Handler` trait definition here

pub trait Traverse: Handler {
  fn stop(&self) {
    STOP_TRAVERSE.with(|stop_flag| {
      *stop_flag.borrow_mut() = true;
    });
  }

  fn traverse<'a, N>(&mut self, node: N, ctx: &mut Context)
  where
    N: NodeTrait<'a>,
  {
    // call a handler
    // each handler is allowed to set `STOP_TRAVERSE` to `true` if it wants

    // check `STOP_TRAVERSE` before traversing the tree
    STOP_TRAVERSE.with(|stop_flag| {
      if !*stop_flag.borrow() {
        // Walk the child nodes recursively.
        for child in node.children() {
          self.traverse(child, ctx);
        }
      }

      // reset the flag
      *stop_flag.borrow_mut() = false;
    });

    // ...
  }
}
```

In this example we store a flag each handler uses to signal to stop further traversing in a thread local variable. In this way, each handler doesn't have to return something if it is okay to continue traversing. If it wants to stop it, it just sets `STOP_TRAVERSE` to `true`.
Probably it works, but there should be a more sophisticated and elegant way to do it...

dsherret:
I think the current solution is good in that visiting the children is opt-out instread of an error-prone opt-in, but I was thinking the same thing as Bartek if this is barely used. Could the flag be stored on `Context`?

magurotuna:
@dsherret 
Storing on `Context` should work, but I don't really want to do so because `Context` is meant to store data that should be shared across running every lint rule. `Context` is created just after the AST view on a certain file is created, and lives until all lint rules are applied to the view. I prefer that a flag used for whether visiting children are stopped or not is encapsulated in the handler/traversal logic.

magurotuna:
@bartlomieju @dsherret 
After further consideration, I think that it wouldn't be bad to store the flag on `Context`. So I have fixed the code to go with this plan.
Can you take look at it once again?

dsherret (approved):
LGTM. Especially considering this is not going to be used that much, I think this is the better solution. We can always revert to the returned value in the future if we find out that would have been better.

PR #728, opened 2021-06-12
refactor(constructor-super): reimplement with ast-view

magurotuna:
Reimplements `constructor-super` with ast-view.


bartlomieju (approved):
Net lines removed = great PR. LGTM!

PR #734, opened 2021-06-15
fix(no-unused-vars): suggest aliasing in the case of named imports

### LINKED ISSUE #730: Unused import triggers no-unused-vars rule
### Lint Name

no-unused-vars

### Code Snippet

```ts
import { anything } from './file.ts';
```

### Expected Result
Warning related to unused import

### Actual Result
no-unused-vars warning

### Additional Info
To be specific, renaming... certainly works, but isn't as simple as the warning suggests

### Version
deno 1.11.0 (release, x86_64-unknown-linux-gnu)
v8 9.1.269.35
typescript 4.3.2

magurotuna:
I don't think I precisely understand what is wanted in #730, but I believe this PR fixes #730.

This PR fixes the hint issued by `no-unused-vars` in the case of named imports. Current hint looks like:

```
(no-unused-vars) `foo` is never used
import { foo } from "./foo.ts";
         ^^^
    at /private/tmp/maguro.ts:1:9

    hint: If this is intentional, prefix it with an underscore like `_foo`

Found 1 problem
Checked 1 file
```

but this is obviously wrong, because renaming `foo` to `_foo` results in import failure. The hint should suggest aliasing `foo` with an underscore, like `foo as _foo`. So the ideal hint would be:

```
error[no-unused-vars]: `foo` is never used
 --> maguro.ts:1:10
  |
1 | import { foo } from "./foo.ts";
  |          ^^^
  |
  = help: If this is intentional, alias it with an underscore like `foo as _foo`
Found 1 problems
```

This PR realizes this ideal style of hint for named imports.

---

For other import statements including default import and `* as` import, the output is going to be the same as the current one, as below:

```
error[no-unused-vars]: `bar` is never used
 --> maguro.ts:2:13
  |
2 | import * as bar from "./bar.ts";
  |             ^^^
  |
  = help: If this is intentional, prefix it with an underscore like `_bar`
error[no-unused-vars]: `baz` is never used
 --> maguro.ts:3:8
  |
3 | import baz from "./baz.ts";
  |        ^^^
  |
  = help: If this is intentional, prefix it with an underscore like `_baz`
```

### DIFF
--- a/src/rules/no_unused_vars.rs
+++ b/src/rules/no_unused_vars.rs
@@ -36,6 +36,12 @@ enum NoUnusedVarsHint {
     _0
   )]
   AddPrefix(String),
+  #[display(
+    fmt = "If this is intentional, alias it with an underscore like `{} as _{}`",
+    _0,
+    _0
+  )]
+  Alias(String),
 }
 
 impl LintRule for NoUnusedVars {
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, this is a very useful hint 👍  nice work

PR #738, opened 2021-06-17
fix(ban-unused-ignore): allow file-level ignore directive to suppress `ban-unused-ignore`

magurotuna:
This PR allows file-level ignore directives to suppress `ban-unused-ignore`.
This rule is "meta" rule, and thus handled differently than other normal rules internally, resulting in the diagnostics from this rule not being suppressed via the `// deno-lint-ignore-file` directive, which will be addressed by this PR. Also, adds documentation about how to ignore `ban-unused-ignore` itself to [Ignoring rules](https://lint.deno.land/ignoring-rules) page.

Many parts of the code diff are related to refactoring. 

Fixes https://github.com/denoland/deno/issues/11011
Blocked by #737 (because this PR makes use of the fact that array types get to implement IntoIterator since 1.53.0)


lawrence-dol:
@magurotuna : Using `dlint --config=...` does not suffer from this problem, understandably, so if `deno lint` is enhanced with the ability to specify a config file, and the config does not trigger these errors, that meets my need in a better way.

There is a strong argument to be made, if a config can be specified, that file-level or statement-level lint directives _ought_ warn if a inapplicable (or no longer applicable) directive is specified.

So this could be dropped in favor of adding support for a config file, as is already there for `dlint`, per #11012.

magurotuna:
@lawrence-dol 
Good to hear that you are happy with dlint. But I guess it will take not a small time to introduce the config file feature into `deno lint` because it touches the philosophy of the linter so we need to discuss it carefully. 

---
Note to self: Discussed with @bartlomieju on discord, we've decided to exclude support for the _line-level_ ignore directive on `ban-unused-ignore` because we want to avoid introducing an unnecessary complication that we have to handle `// deno-lint-ignore ban-unused-ignore` differently from other rules. Also, considering the situation where we feel like ignoring `ban-unused-ignore` itself, file-level should be enough.

magurotuna:
@bartlomieju This PR is ready for review. PTAL :)

bartlomieju (approved):
@magurotuna as always implementation, tests and documentation look great! Before landing I'll wait for review from @dsherret

src/ignore_directives.rs
```diff
@@ -151,62 +178,107 @@ target: Record<string, any>,
...
+"#,
+      |program, source_map| {
+        let file_directive = parse_file_ignore_directives(
+          "deno-lint-ignore-file",
+          &source_map,
+          program,
+        );
+
+        assert!(file_directive.is_none());
+      },
```
bartlomieju, line 281: These tests are much more readable 👍

dsherret (approved):
Very nice!

PR #740, opened 2021-06-20
chore: follow Rust naming convention

magurotuna:
Clippy complains about naming convention: https://github.com/denoland/deno_lint/runs/2865828573
This commit addresses it.

Blocks #715


ry (approved):
thanks

PR #743, opened 2021-06-20
fix(no-unused-vars): handle es private overloaded methods

### LINKED ISSUE #739: Private overload arguments get flagged as unused
The arguments inside the overloads get flagged by `no-unused-vars` as if the private overloads were in fact real functions

### Lint Name
no-unused-vars

### Code Snippet
```ts
class Test {
  #myFunction(value: string): string;
  #myFunction(value: number): number;
  #myFunction(value: string | number) {
    return value;
  }
}
```

### Expected Result
Since private functions became a supported feature by TS 4.3, they should behave as normal functions would do regarding overloads

### Version
deno 1.11.1 (release, x86_64-unknown-linux-gnu)
v8 9.1.269.35
typescript 4.3.2

magurotuna:
Fixes #739


bartlomieju (approved):
LGTM, good fix 👍

PR #749, opened 2021-06-22
fix(no-fallthrough): handle empty block in SwitchCase correctly

### LINKED ISSUE #746: Incorrect Flagging of Fallthrough
When braces are used around a switch's case block, and the block is empty, dlint incorrectly flags the case as a fallthrough. Also, note that the same error is reported twice, which is a separate bug.

### Lint Name

no-fallthrough

### Code Snippet

```js
switch(somevalue) {
    case 0: {
        // nothing to do for this value
        } break;
    case 1: {
        // nothing to do for this value
        } break;
    }
```

### Expected Result

No errors.

### Actual Result

```
 --> test.js:2:5
38;5;12m  |
2 |       case 0: {
  |  _____^
3 | |         // nothing to do for this values
4 | |         } break;
  | |_________________^
  |
  = help: Add `break` or comment `/* falls through */` to your case statement
 --> test.js:2:5
  |
2 |       case 0: {
  |  _____^
3 | |         // nothing to do for this values
4 | |         } break;
  | |_________________^
  |
  = help: Add `break` or comment `/* falls through */` to your case statement
Found 2 problems
```

### Additional Info

This is using dlint.

### Version

deno 1.11.1 (release, x86_64-pc-windows-msvc)
v8 9.1.269.35
typescript 4.3.2

dlint 0.6.0

magurotuna:
Fixes #746

(The second bug reported in #746 hasn't been fixed by this PR though. It isn't reproduced in my environment.)

### DIFF
--- a/src/rules/no_fallthrough.rs
+++ b/src/rules/no_fallthrough.rs
@@ -60,7 +60,7 @@ statement, intending only for a single case statement to be executed.  This
 rule enforces that you either end each case statement with a break statement or
 an explicit comment that fallthrough was intentional.  The fallthrough comment
 must contain one of `fallthrough`, `falls through` or `fall through`.
-    
+
 ### Invalid:
 ```typescript
 switch(myVar) {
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

Issue #752, opened 2021-06-23
Syntax errors panic instead of emitting a cleanly formatted error.


lawrence-dol:
### Lint Name

N/A.

### Code Snippet

```ts
function square(x) {
    return x * x;
    )}
```

### Expected Result

````
Syntax error: Unexpected token `)`. Expected this, import, async, function, [ for array literal, { for object literal, @ for decorator, function, class, null, true, false, number, bigint, string, regexp, ` for template literal, (, or an identifier.
  --> C:\...\test-module.js:3:4
   |
 3 |     )}
   |     ^
   |
Found 1 problem
````

### Actual Result

````
Validate: C:\...\test-module.js
thread 'main' panicked at 'Failed to lint: SwcDiagnosticBuffer { diagnostics: ["Unexpected token `)`. Expected this, import, async, function, [ for array literal, { for object literal, @ for decorator, function, class, null, true, false, number, bigint, string, regexp, ` for template literal, (, or an identifier at C:\\...\\test-module.js:3:4"] }', examples\dlint\main.rs:203:8
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
````

### Additional Info

This looks like a simple case of default error handling that could be formatted to be much easier to parse, and therefore more useful.

### Version

dlint 0.6.0

lawrence-dol:
This is sufficiently clean and useful now. Much easier to find the actual fault. Thanks!

PR #754, opened 2021-06-26
refactor(dlint): use `try_for_each` instead of `for_each`

magurotuna:
This patch replaces `try_each` with `try_for_each` in `dlint`, which allows us to use `?` operator inside the closure. As a result, we can propagate errors from inside the closure to outside. The benefit of this is that we can remove `expect` calls on `Option` so that users will get a little bit user-friendly error messages with Rust's default panic messages removed, like:

```
Error: Unexpected token `)`. Expected this, import, async, function, [ for array literal, { for object literal, @ for decorator, function, class, null, true, false, number, bigint, string, regexp, ` for template literal, (, or an identifier at maguro.ts:3:4
```

For comparison, the previous error message looked like:

```
thread 'main' panicked at 'Failed to lint: SwcDiagnosticBuffer { diagnostics: ["Unexpected token `)`. Expected this, import, async, function, [ for array literal, { for object literal, @ for decorator, function, class, null, true, false, number, bigint, string, regexp, ` for template literal, (, or an identifier at maguro.ts:3:4"] }', examples/dlint/main.rs:203:8
```

This previous message had so much information that it might confuse the end users unnecessarily.

For reviewers, the diff looks clean with the whitespace diffs hidden: https://github.com/denoland/deno_lint/pull/754/files?diff=split&w=1

----

@lawrence-dol Does this fix meet your need described in #752? Your expectation contains a snippet where you can easily find the place with invalid syntax, but this patch doesn't do it. The change for this would require more rework, and I'm leaning towards putting it off.

### DIFF
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -166,55 +166,57 @@ fn run_linter(
   }
 
   let file_diagnostics = Arc::new(Mutex::new(BTreeMap::new()));
-  paths.par_iter().for_each(|file_path| {
-    let source_code =
-      std::fs::read_to_string(&file_path).expect("Failed to load file");
+  paths
+    .par_iter()
+    .try_for_each(|file_path| -> Result<(), AnyError> {
+      let source_code = std::fs::read_to_string(&file_path)?;
 
-    let mut rules = if let Some(config) = maybe_config.clone() {
-      config.get_rules()
-    } else {
-      get_recommended_rules()
-    };
+      let mut rules = if let Some(config) = maybe_config.clone() {
+        config.get_rules()
+      } else {
+        get_recommended_rules()
+      };
 
-    if let Some(rule_name) = filter_rule_name {
-      rules = rules
-        .into_iter()
-        .filter(|r| r.code() == rule_name)
-        .collect()
-    };
+      if let Some(rule_name) = filter_rule_name {
+        rules = rules
+          .into_iter()
+          .filter(|r| r.code() == rule_name)
+          .collect()
+      };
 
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM 🚀

lawrence-dol:
@magurotuna : It's a significant improvement, but it would be much better if the format was closer to what I had suggested, if that's feasible. Particularly the line with the `filename:line:column` for editors to link the error to the source file.

```
Error: Unexpected token `)`. Expected this, import, async, function, [ for array literal, { for object literal, @ for decorator, function, class, null, true, false, number, bigint, string, regexp, ` for template literal, (, or an identifier.

  --> C:\...\maguro.ts:3:4
   |
 3 |     )}
   |     ^
   |
Found 1 problem
```

magurotuna:
@lawrence-dol How about using `deno fmt --check`? It will output like

```
$ deno fmt --check maguro.ts
Error checking: /Users/yusuktan/Repos/github.com/magurotuna/deno_lint/maguro.ts
   Line 3, column 5: Unexpected token `)`. Expected this, import, async, function, [ for array literal, { for object literal, @ for decorator, function, class, null, true, false, number, bigint, string, regexp, ` for template literal, (, or an identifier

      )}
      ~
Checked 1 file
```

lawrence-dol:
@magurotuna : That also emits all the things it doesn't like about the code formatting, which a large number of things in a minimized file 😆.

lawrence-dol:
@magurotuna : I would like to see my suggested output format done eventually, but this is "good enough".

PR #755, opened 2021-06-27
Recommend to use c++filt

kubo39:
Since 2016-11-16 c++filt [1] recognizes rust legacy demangling and will supports newer one(v0 name mangling) [2] [3], so it would be better choice.

[1]: https://sourceware.org/binutils/docs/binutils/c_002b_002bfilt.html
[2]: https://github.com/rust-lang/rfcs/pull/2603
[3]: https://sourceware.org/bugzilla/show_bug.cgi?id=27194

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,13 +148,15 @@ $ cargo test
 
 Prerequisites:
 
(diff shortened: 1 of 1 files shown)


magurotuna (approved):
LGTM 👍

bartlomieju (approved):
Thank you @kubo39

PR #756, opened 2021-06-28
feat(dlint): print rule doc with syntax highlighting

magurotuna:
This PR adds a feature that each rule documentation is printed to the terminal with syntax highlight. If this turns out to work fine, we'll add it to `deno lint` as well (https://github.com/denoland/deno/issues/8227).

Current screenshot:
image

- [x] use the same colorings as `deno repl` and `deno doc` (ref: https://github.com/denoland/deno/blob/main/cli/tools/repl.rs#L251)
- [x] add syntax highlight to JS/TS code
- [x] check if it's tty

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,7 +81,7 @@ jobs:
 
       - name: Generate rules json
         if: contains(matrix.os, 'ubuntu')
-        run: target/release/examples/dlint rules --json --all > docs.json
+        run: target/release/examples/dlint rules --json > docs.json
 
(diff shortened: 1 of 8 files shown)


magurotuna:
Added syntax highlight

image

caspervonb (approved):
LGTM!

bartlomieju (approved):
LGTM, this looks beautiful. Let's think how we can share this functionality between `deno_lint`, `deno_doc` and `deno` REPL (so for `deno lint`, `deno doc` and `deno` subcommands).

PR #767, opened 2021-07-10
docs(no-obj-calls): add doc

magurotuna:
This commit adds documentation for the `no-obj-calls` rule and fixes behavior for an edge case where target built-in objects are shadowed by users.


bartlomieju (approved):
Good fix 👍

PR #769, opened 2021-07-10
BREAKING: always enable `ban-unknown-rule-code` and `ban-unused-ignore` & add these docs

magurotuna:
This PR contains a breaking change for the public interface on `LinterBuilder`. Currently `ban-unknown-rule-code` and `ban-unused-ignore` can be opted out by a downstream user via `LinterBuilder`, but this PR will always enable both of them. Accordingly, `lint_unknown_rules(bool)` and `lint_unused_ignore_directives(bool)` on the `LinterBuilder` are removed.

In addition, docs for these two rules are added so that we can see them on https://lint.deno.land.

### DIFF
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -189,9 +189,7 @@ fn run_linter(
 
       let mut linter_builder = LinterBuilder::default()
         .rules(rules)
-        .syntax(determine_syntax(file_path))
-        .lint_unknown_rules(true)
-        .lint_unused_ignore_directives(true);
+        .syntax(determine_syntax(file_path));
 
       for plugin_path in &plugin_paths {
         let js_runner = js::JsRuleRunner::new(plugin_path);
--- a/src/context.rs
+++ b/src/context.rs
@@ -150,6 +150,9 @@ impl<'view> Context<'view> {
     filtered
   }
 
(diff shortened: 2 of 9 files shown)


bartlomieju (commented):
@magurotuna what's the reason for this change?

magurotuna:
@bartlomieju 
The primary motivation is to show the docs of `ban-unknown-rule-code` and `ban-unused-ignore` on https://lint.deno.land. When doing this, we need to decide whether they should be tagged with "recommended". If tagged with "recommended", the docs will be shown on the site by default. If not, they won't by default. 
I think it would be better to treat them as "recommended", since they are generally useful for everyone utilizing the linter.
Tagging them as "recommended" also means that they are run by default. Here the problem arises that these two rules are not always run; actually they are run only when opted in via `LinterBuilder`. 

So in short, if we decide that the two rules should be included in "recommended" set, they should be always run regardless of `LinterBuilder`. Whether a rule is run or not should be decided based on the fact that it is tagged as "recommended" for consistency. Otherwise, a few rules that are not "recommended" would be run, which might confuse users.

bartlomieju (approved):
Thanks for explaining @magurotuna, this makes total sense. Thanks, LGTM!

PR #778, opened 2021-07-17
fix(ban-types): suggest `Record<string, never>` when encountering `{}`

### LINKED ISSUE #775: `ban-types` should show hint that suggests `Record<string, never>` for typing empty object
Ref #774

magurotuna:
Fixes #775


bartlomieju (approved):
LGTM, thanks Yusuke, this is useful hint

PR #781, opened 2021-07-23
feat: implement new lint `prefer-primordials`

### LINKED ISSUE #757: New lint: `prefer-primordials` (dedicated to deno_core)
From https://github.com/denoland/deno/pull/10939#issue-668829704

> Node has a eslint plugin that validates that all globals that are used in a
given script use primordials. We should consider adding a plugin for this to
dlint too. See https://github.com/nodejs/node/blob/master/tools/eslint-rules/prefer-primordials.js

magurotuna:
Resolves #757 

I tested if this rule works fine for the actual code. Please see https://github.com/denoland/deno/pull/11500/


bartlomieju (approved):
LGTM, no comments, this is a great addition. Will need to bump `dlint` in `deno/third_party_tools` after releasing a new version

PR #783, opened 2021-07-28
docs(no-redeclare): add docs

magurotuna:
This PR adds documentation for `no-redeclare`.

### DIFF
--- a/src/rules/no_redeclare.rs
+++ b/src/rules/no_redeclare.rs
@@ -40,6 +40,49 @@ impl LintRule for NoRedeclare {
       ProgramRef::Script(ref s) => visitor.visit_script(s, &DUMMY_NODE),
     }
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows redeclaration of variables.
+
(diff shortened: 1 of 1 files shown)


src/rules/no_redeclare.rs
```diff
@@ -40,6 +40,49 @@ impl LintRule for NoRedeclare {
       ProgramRef::Script(ref s) => visitor.visit_script(s, &DUMMY_NODE),
     }
   }
+
+  fn docs(&self) -> &'static str {
+    r#"Disallows redeclaration of variables.
```
dsherret, line 45: This rule seems to affect also parameter, function, and class property declarations. Perhaps this sentence should be changed?

magurotuna: Good point, I missed those! I'll fix it today.

magurotuna: @dsherret Done, please take a look again.

dsherret (approved):
LGTM. Thanks @magurotuna!

PR #787, opened 2021-08-01
docs: move rule docs to separate markdown files

### LINKED ISSUE #786: Move rules' docs to separate markdown files and let them be included via `include_str!`
Currently, rules' docs are written directly in the Rust source code in markdown format, like https://github.com/denoland/deno_lint/blob/219f132d9c8a9864ce3eaea338ac6068cf0ecb7b/src/rules/no_with.rs#L36-L52

This works fine, but we could move them to separate markdown files and let them be included via `include_str!` like

```rust
  include_str!("path/to/doc.md")
}
```

The nice things about it are:

1. docs will be rendered on GitHub. (we have https://lint.deno.land already, but it'd be great to see them rendered on GitHub as well)
2. we'll be able to utilize markdown formatter

magurotuna:
This commit moves rule docs to separate markdown files and let them be included via `include_str!` macro. Also, `deno fmt` is run on the created markdown files.

Rendered example is [here](https://github.com/magurotuna/deno_lint/blob/include_str/docs/rules/adjacent_overload_signatures.md)

Closes #786


bartlomieju:
@magurotuna would it be possible to not include all those files if certain cargo flag is provided? `deno lint` doesn't print the docs anyway, so we could save a couple hundred KBs not including them.

magurotuna:
@bartlomieju 
That would be nice, I'll give it a try.

magurotuna:
@bartlomieju Added a new feature `docs` and made `LintRule::docs` method be behind this feature gate. PTAL again

bartlomieju:
@magurotuna this looks good to me; a few questions:
a) did you already format Markdown files?
b) does it make sense to split them, given that we do want to have Markdown printer in `deno` after all? (I guess it does, formatting is preferable)

Also @dsherret can you take a look and give your opinion?

magurotuna:
@bartlomieju 
a) Yes they are already formatted.
b) The biggest motivation is that we can format the docs if we separate them. Also, inline docs as Rust string literals are not good in that they are not highlighted in the editor and that we cannot utilize real-time markdown preview functionality some editors provide. Plus docs will be rendered on GitHub like [this](https://github.com/magurotuna/deno_lint/blob/include_str/docs/rules/adjacent_overload_signatures.md), which is kind of trivial though.

bartlomieju:
@magurotuna perfect, that sounds good to me.

magurotuna:
@bartlomieju That's also already done :) here's `format.ts` that is run in the CI process.
 https://github.com/denoland/deno_lint/blob/b476a645243b6f8533d1fbb68fdd612ac68d2e74/tools/format.ts#L30-L39

bartlomieju:
> @bartlomieju That's also already done :) here's `format.ts` that is run in the CI process.
> https://github.com/denoland/deno_lint/blob/b476a645243b6f8533d1fbb68fdd612ac68d2e74/tools/format.ts#L30-L39

Perfect, I missed that :))

bartlomieju (approved):
LGTM, thanks @magurotuna!

PR #789, opened 2021-08-09
v0.12.0

bartlomieju:


Cargo.toml
```diff
@@ -20,13 +20,13 @@ log = "0.4.14"
...
-swc_common = "0.11.0"
-swc_ecmascript = { version = "0.46.0", features = ["parser", "transforms", "utils", "visit"] }
+swc_common = "0.11.4"
+swc_ecmascript = { version = "0.52.1", features = ["parser", "transforms", "utils", "visit"] }
 regex = "=1.4.3"
 once_cell = "1.5.2"
 derive_more = { version = "0.99.14", features = ["display"] }
 anyhow = "1.0.40"
-dprint-swc-ecma-ast-view = "0.24.0"
+dprint-swc-ecma-ast-view = "0.25.0"
```
dsherret, line 29: Don't use this yet. It failed on CI, but still published.

bartlomieju: Okay, thanks for the heads up

dsherret: Actually, it's fine to use this, but now I have to investigate why the code generation isn't working in the latest nightly.

dsherret (approved):

PR #792, opened 2021-08-11
refactor: remove usage of swc's `SourceMap` and `SourceFile`

### LINKED ISSUE #790: Remove usage of `SourceMap` and `SourceFile`
I'm going to work on https://github.com/dprint/dprint-swc-ecma-ast-view/issues/28 first, but we should remove the usage of these as there is no need to have them and they do not implement `Sync`.

This is part of the work for reducing extra parses in Deno's language server.

dsherret:
Closes #790

FYI @magurotuna


src/ast_parser.rs
```diff
@@ -133,83 +102,60 @@ pub(crate) struct AstParser {
...
+      SwcDiagnostic::from_swc_error(file_name, &source_file, err)
+    })?;
 
-    let parse_result = parser.parse_program().map_err(move |err| {
-      let mut diagnostic_builder = err.into_diagnostic(&self.handler);
-      diagnostic_builder.emit();
-      SwcDiagnosticBuffer::from_swc_error(buffered_err, self)
+    let program = swc_common::GLOBALS.set(&self.globals, || {
+      program.fold_with(&mut ts_resolver(self.top_level_mark))
     });
```
dsherret, line 144: What does this do?

src/diagnostic.rs
```diff
@@ -1,25 +1,23 @@
...
 use serde::Serialize;
-use std::convert::TryInto;
 
 #[derive(Debug, Clone, Copy, PartialEq, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Position {
-  pub line: usize,
-  pub col: usize,
+  pub line_index: usize,
+  pub column_index: usize,
```
dsherret, line 8: This should all be 0-index now except when displayed or in the tests. (Still that weird 1-indexed line and 0-indexed column in the tests... maybe we could change that later to both 1-indexed over there)

src/rules.rs
```diff
@@ -55,7 +55,6 @@ pub mod no_invalid_regexp;
 pub mod no_invalid_triple_slash_reference;
 pub mod no_irregular_whitespace;
 pub mod no_misused_new;
-pub mod no_mixed_spaces_and_tabs;
```
dsherret, line 58: Should this rule exist? I think it's a formatting concern of `deno fmt` and not `deno lint` so I removed it, but I mainly removed it in this PR because it would be a bit of work to update it to work now (because it was using `source_map().span_to_lines(...)`).

examples/dlint/js.rs
```diff
@@ -149,7 +149,7 @@ impl Plugin for JsRuleRunner {
   fn run(
     &mut self,
     context: &mut Context,
-    program: Program,
+    program: ProgramRef,
```
dsherret, line 152: The plugins now take a `ProgramRef` because that's what might be provided to `parse_with_ast`. This also eliminates cloning of a whole AST potentially a bunch of times.

src/diagnostic.rs
bartlomieju: Keep in mind that this is used by LSP/vscode_deno, so changing it might break the former

bartlomieju (commented):
This looks good to me, I'll let @magurotuna review it as well

examples/dlint/js.rs
```diff
@@ -11,12 +11,12 @@ use deno_core::ZeroCopyBuf;
 use deno_lint::context::Context;
 use deno_lint::control_flow::ControlFlow;
 use deno_lint::linter::Plugin;
+use deno_lint::rules::ProgramRef;
```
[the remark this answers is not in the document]
dsherret: Actually, I think it would be ok to remove the `ProgramRef` now and use the one in `dprint_swc_ecma_ast_view`.

magurotuna:
I will review at night (in about 10 hours).

src/diagnostic.rs
magurotuna: I think we shouldn't use `usize` everywhere to represent line/column numbers because overuse of `usize` may result in developers' confusion in whether this type expresses 0-indexed value that is internally used or 1-indexed value that is for displaying diagnostics to end users. Instead, we could create new types, one representing internal values and the other values to be displayed.

---

Oh I found https://github.com/dsherret/text_lines which provides new types I described above. Maybe we could utilize it later on.

magurotuna (commented):
Looks good overall, but one point I want to ask is whether we should replace our `ProgramRef` with the one in `dprint_swc_ecma_ast_view`.

src/ast_parser.rs
bartlomieju: This is used to apply proper "syntax context" to all AST elements - when SWC performs transforms/folding it might change some of those context and "ts_resolver" ensures that all elements end up in proper lexical scope.

src/diagnostic.rs
dsherret: I'll add some comments to indicate it's 0-indexed. I think using `usize` is useful here because then you can index these types into an array of lines or a characters.

Yup, we could use text_lines in the future.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ default = []
 docs = []
 
 [dependencies]
+ast_view = { version = "0.29.3", package = "dprint-swc-ecma-ast-view" }
```
dsherret, line 24: The name `dprint_swc_ecma_ast_view` is way too long. Is this ok to do instead?

examples/dlint/js.rs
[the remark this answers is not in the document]
dsherret: Done now.

Cargo.toml
magurotuna: Fantastic

src/diagnostic.rs
magurotuna: I see, that makes sense!

src/rules.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2020-2021 the Deno authors. All rights reserved. MIT license.
 use crate::context::Context;
-use dprint_swc_ecma_ast_view::Program as ProgramView;
+
+pub use ast_view::Program;
+pub use ast_view::ProgramRef;
```
magurotuna, line 5: How about moving these `pub use` to the crate root (i.e. src/lib.rs)?
These two types are now included in the public interface of this crate, and also it'd be more intuitive for them to live and re-exported in the crate root than to be here as they don't seem to be related to "rules" very much.

bartlomieju (approved):

src/rules.rs
dsherret: Done in c9e2996e0c362130deb79cfc994d9041dde560bb.

magurotuna (approved):
LGTM 👍

PR #793, opened 2021-08-12
fix(ban_ts_comment): treat trailing spaces invalid

littledivy:
```diff
+ // @ts-ignore      
```
was treated valid by `ban_ts_comment`

### DIFF
--- a/src/rules/ban_ts_comment.rs
+++ b/src/rules/ban_ts_comment.rs
@@ -135,11 +135,11 @@ fn check_comment(comment: &Comment) -> Option<DirectiveKind> {
   }
 
   static EXPECT_ERROR_REGEX: Lazy<Regex> =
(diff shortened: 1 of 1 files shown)


magurotuna (commented):
The code looks good, but I have one question.
It seems like TypeScript accepts `// @ts-ignore` with trailing spaces or even extra characters as a valid directive. (see [the playground](https://www.typescriptlang.org/play?#code/PTAEAEBcGcFoEsDmA7A9gJwKah7vpMAPSdAQ1AGMALU9aAKABtNJRSAuUaE+ZRUALygALACYA3PSYtQAI07d0vfkLHigA))
In spite of this fact, you think trailing spaces should be warned by the linter, right?

littledivy:
> In spite of this fact, you think trailing spaces should be warned by the linter, right?

Yes - by "invalid" I meant linter should not treat trailing white space as a valid reason to a `@ts` directive.

magurotuna (approved):
LGTM 👍

Issue #765, opened 2021-07-08
Replace `markdown` and `ansi_term` with `pulldown-cmark`, `termcolor`


magurotuna:
In dlint we are using `markdown` and `ansi_term` to render markdown docs to the terminal, but when porting this feature to Deno CLI, it's desirable to replace these crates with ones that already exist in the dependency tree of deno cli. For compatibility dlint should probably depends on the same crates as well.
ref: https://github.com/denoland/deno/pull/11244#issuecomment-875321311

bartlomieju:
CC @dsherret I think you mentioned that you wanted to move away from `pulldown-cmark`?

dsherret:
Hey @magurotuna, there was a discussion today again about adding deno lint rules to the CLI docs and we thought linking to https://lint.deno.land would be sufficient enough information to give to users. The main reason is that we don't believe many people will be reading the lint rules from the CLI and linking them to https://lint.deno.land is a better experience with less for us to maintain in the long run. It would also reduce a bit of complexity because we wouldn't have to syntax highlight markdown and we wouldn't have to look into replacing ansi_term and markdown since we could remove them. What are your thoughts?

I feel bad though because you've put a lot of effort into this and done such a great job.

magurotuna:
@dsherret 
I agree that linking to https://lint.deno.land would be sufficient, but I don't think it's that much of a burden for us to maintain CLI doc because the source of docs is identical. Also, syntax highlight for markdown could be used for other purposes like `deno doc` command output, so even if we decide to not add the feature of lint docs as CLI, the code for markdown pretty printer will probably remain in the codebase.

magurotuna:
According to the design meeting held on July 8th (#11365), it seems like this feature is not what we should land. Close.

PR #810, opened 2021-08-22
refactor(dlint): replace `markdown` with `pulldown-cmark`

magurotuna:
This PR replaces `markdown` with `pulldown-cmark`. Both are used as markdown parser for printing markdown in pretty format to the terminal. Both work fine for this purpose, but `pulldown-cmark` is preferable because it already exists in the dependency tree of Deno.

To properly handle nested decorations like "bar" in "**foo ~~bar~~ baz**", I had to work with ANSI color codes directly, rather than using an external crate.

Ref #765, https://github.com/denoland/deno/pull/11244

### Demo

no-this-alias

image

no-deprecated-deno-api

image


bartlomieju:
Thanks @magurotuna, this looks good to me, the colors for highlighting look really good. @dsherret do you have any remarks?

examples/dlint/color.rs
```diff
@@ -1,195 +1,318 @@
...
+use if_chain::if_chain;
+use pulldown_cmark::{Options, Parser, Tag};
 use std::convert::TryFrom;
 use swc_ecmascript::parser::token::{Token, Word};
 
-pub trait Colorize {
-  fn colorize(self) -> String;
+pub fn colorize_markdown(input: &str) -> String {
+  let mut options = Options::empty();
+  options.insert(Options::ENABLE_STRIKETHROUGH);
```
dsherret, line 10: Maybe we don't need more options for dlint, but what I did in dprint was:

```rust
options.insert(Options::ENABLE_TABLES);
options.insert(Options::ENABLE_FOOTNOTES);
options.insert(Options::ENABLE_STRIKETHROUGH);
options.insert(Options::ENABLE_TASKLISTS);
```

dsherret (approved):
LGTM too

examples/dlint/color.rs
magurotuna: thanks for the info. We'll enable other options once it's needed :)

PR #813, opened 2021-08-23
feat(no-var): add "recommended" tag to `no-var`

magurotuna:
This PR adds `no-var` to the recommended set of rules.

`no-var` is currently not in the recommended set simply because ESLint's recommended set doesn't include the rule either. Not really sure why ESLint doesn't, but given that this rule is considered to be always useful when writing TypeScript/JavaScript in modern fashion, we should probably make it "recommended".


dsherret (approved):
LGTM and saw discussion on discord.

Issue #492, opened 2020-11-02
Consider putting rule-specific analyzers into single one


magurotuna:
We have many scope analyzers dedicated to each rule right now. In fact, as far as I see, the following rules have their own scope analyzer:

- prefer-const
- no-undef (#486)
- no-unused-vars
- no-import-assign
- no-global-assign

On the other hand, the following rules use the common scope analyzer (implemented in `scopes.rs`):

- no-ex-assign
- no-func-assign
- no-class-assign
- no-const-assign
- no-shadow-restricted-names
- no-obj-calls
- prefer-primordials
- no-new-symbol

Ideally, we should have only one scope analyzer - it would be best to merge the features of the rule specific analyzers into the common analyzer, which I believe can be realized.

bartlomieju:
@kdy1 can you look into this issue?

kdy1:
@bartlomieju Of course.

The rules below can be merged using a simple trait.

 - no-undef
 - no-import-assign
 - no-ex-assign (uses `scopes.rs`)
 - no-func-assign (uses `scopes.rs`)
 - no-class-assign (uses `scopes.rs`)
 - no-const-assign (uses `scopes.rs`)
 - no-const-assign (uses `scopes.rs`)
 - no-shadow-restricted-names (uses `scopes.rs`)

Alternatively, `no-undef` and `no-import-assign` can be modified to use data from `scopes.rs` easily. (I found that it's possible after looking at `no-ex-assign`.)

bartlomieju:
@kdy1 either of solutions works fine for me, decide yourself which one makes more sense.

bartlomieju:
@magurotuna is this still relevant after all your refactors?

magurotuna:
sorry I didn't notice the mention.
Yes it's still relevant, I'll look into it again

PR #822, opened 2021-08-29
refactor(no-global-assign): remove its own scope analyzer

magurotuna:
This PR removes the rule-dedicated scope analyzer from `no-global-assign`, and instead lets it use the common scope analyzer.

Towards #492


bartlomieju (approved):
LGTM

PR #824, opened 2021-08-30
feat: implement new lint `no-window-prefix`

### LINKED ISSUE #682: Add rule disallowing prefixing of web APIs with `window`
There should be a warning for `window.fetch`, recommending to replace with `fetch` or `globalThis.fetch` for compatibility between window, web workers, and service workers.

This should be limited to the known subset of APIs that are available in both the window, and workers.

magurotuna:
This PR adds a new lint rule `no-window-prefix` that bans the use of Web APIs via `window` object.

Closes #682

### DIFF
--- a/docs/rules/no_window_prefix.md
+++ b/docs/rules/no_window_prefix.md
@@ -20,7 +20,9 @@ Note that the following properties are allowed to call with `window`:
 - `window`
 - `Navigator`
 
-because these APIs are not supported in Workers.
+because these APIs are not supported in Workers. Additionally, `location` is
+also allowed because what it points to in the Window context is different from
+that in Web Workers.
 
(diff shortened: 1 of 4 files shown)


kitsonk:
I don't know off hand if we implement it, but a worker `self.location` is not the same object as `window.location`, they are two different DOM object prototypes too.  It make sense to except `.location` from the list as well.

lucacasonato (approved):
Looks great! Thanks @magurotuna  🙏

lucacasonato:
I do also agree with excluding 'location'

magurotuna:
Thanks for the review, added `location` to the allowlist.

bartlomieju (approved):

PR #825, opened 2021-08-31
fix(no-deprecated-deno-api): handle cases where `Deno` is shadowed in another scope

magurotuna:
This PR fixes `no-deprecated-deno-api` so that it can correctly handle cases where `Deno` is shadowed in another scope.


bartlomieju (approved):
LGTM, good fix

PR #829, opened 2021-08-31
feat: Add 'get_filtered_rules' helper

bartlomieju:


src/rules.rs
```diff
@@ -237,6 +239,68 @@ pub fn get_recommended_rules() -> Vec<Box<dyn LintRule>> {
...
+///   is empty then all rules will be excluded by default
+///
+/// - if `maybe_exclude` is `Some`, all rules with matching codes will
+///   be filtered out
+///
+/// - if `maybe_include` is `Some`, rules with matching codes will be added
+///   to the return list
+///
+/// Before returning the list will be deduplicated and sorted alphabetically.
+pub fn get_filtered_rules(
```
dsherret, line 257: Could this be used in dlint?

src/rules.rs
```diff
@@ -237,6 +239,68 @@ pub fn get_recommended_rules() -> Vec<Box<dyn LintRule>> {
...
+    let tags_set: HashSet<String> = HashSet::from_iter(tags.into_iter());
+    rules = rules
+      .into_iter()
+      .filter(|rule| {
+        let rule_tags = rule.tags().into_iter().map(|t| t.to_string());
+        let rule_tag_set: HashSet<String> = HashSet::from_iter(rule_tags);
+        !tags_set
+          .intersection(&rule_tag_set)
+          .collect::<Vec<_>>()
+          .is_empty()
```
dsherret, line 274: I think you could just do something along the lines of:

```rust
rule.tags().iter().any(|t| tags_set.contains(t))
```

dsherret (approved):
Looks ok, but I think it could be simplified by having a single `get_all_rules().into_iter().filter(|rule| ...)`. Also, perhaps we should add a few basic tests?

bartlomieju:
> Looks ok, but I think it could be simplified by having a single `get_all_rules().into_iter().filter(|rule| ...)`. Also, perhaps we should add a few basic tests?

I find it easier to reason about when it's split into discrete steps as described in the docstring.

src/rules.rs
bartlomieju: I changed it, but handling `String` and `&'static str` is frustrating...

dsherret:
@bartlomieju I was thinking something along the lines of:

```rust
let tags_set = maybe_tags.map(|tags| HashSet::from_iter(tags.into_iter());

let rules = get_all_rules()
  .into_iter()
  .filter(|rule| {
    if let Some(excludes) = &maybe_exclude {
      if excludes.contains(&rule.code().to_owned()) {
        return false;
      }
    }

    if let Some(includes) = &maybe_include {
      if includes.contains(&rule.code().to_owned()) {
        return true;
      }
    }

    if let Some(tags_set) = &tags_set {
      rule.tags().iter().any(|t| tags_set.contains(&t.to_string()))
    } else {
      true
    }
  })
  .collect::<Vec<_>>();
```

bartlomieju:
@dsherret I applied your suggestion with slight changes to keep the semantics of what was described in the docstring. Please let me know if that works for you.

dsherret (approved):
Yup. Looks good! Approving again.

PR #831, opened 2021-09-01
fix(tools): handle input with two or more hyphens correctly

magurotuna:
This PR fixes the scaffold script so it can handle input lint names with two or more hyphens like `foo-bar-baz` correctly.

### DIFF
--- a/tools/scaffold.ts
+++ b/tools/scaffold.ts
@@ -4,56 +4,66 @@
 import { cyan, green, red } from "https://deno.land/std@0.106.0/fmt/colors.ts";
 import { fromFileUrl, join } from "https://deno.land/std@0.106.0/path/mod.ts";
 
-if (Deno.args.length !== 1) {
-  console.error(red("ERROR"), "Lint name is not specified.");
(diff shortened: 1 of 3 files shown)


tools/scaffold.ts
```diff
@@ -28,13 +31,25 @@ await Promise.all([
...
+  `Also, don't forget to manually add a new lint rule to ${
+    cyan(
+      "get_all_rules",
+    )
+  } function in ${
+    cyan(
+      "src/rules.rs",
+    )
+  } so that the rule will get to be run actually.`,
 );
```
bartlomieju, line 54: Could you add some `Deno.test` here to check it works as expected? You'll probably need to gate using `import.meta.main` to run the actual script

magurotuna: Done in 0c099c4, but running the script actually in the test isn't desirable as it has side effects (e.g. create and edit files), so I just added some unit tests that confirm case-conversion and the contents of generated files are correct.

bartlomieju (approved):
LGTM, thanks!

PR #836, opened 2021-09-03
test(tools): add integration test for scaffold

magurotuna:
This PR adds an integration test for the scaffold script. Specifically, it checks if the created files pass `cargo check`.

I kind of hesitated to add this because running the script has side effects. But after further thought, actually we should have it in order to keep the files created by the scaffold scripts up-to-date and able to compile.


bartlomieju (approved):
LGTM, thanks!

PR #837, opened 2021-09-04
feat: implement new lint `prefer-ascii`

### LINKED ISSUE #832: New lint: `prefer-ascii` for Deno's internal JS
Optimization around deno_core is being made now in https://github.com/denoland/deno/pull/9713. This optimization requires that all the JS code that is eventually bundled in Deno is fully written only in ASCII i.e. one-byte characters. So it would be great to have a new lint rule that makes sure of that.

Similar to `prefer-primordials`, the new lint will be dedicated to the internal of Deno. Not intended to be used by Deno users.

CC @AaronO @bartlomieju

magurotuna:
This PR implements a new lint rule `prefer-ascii` that makes sure that the code is composed of only ASCII characters.

Rendered doc: https://github.com/magurotuna/deno_lint/blob/ace535166b4f42a603f9ff5c540d104b4e932647/docs/rules/prefer_ascii.md

In addition, moves `display_diagnostics` function into `diagnostics.rs` and adds test cases. The previous logic of this function didn't take multi-byte characters into account, which has been fixed as well. However, to properly display diagnostic annotations for double-width characters, we still need to wait for my patch to land: https://github.com/rust-lang/annotate-snippets-rs/pull/46
With the current version of annotate-snippets, displayed diagnostics may point to the wrong position if the snippet contains double-width characters, like:

image

Closes #832

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,8 +35,7 @@ dependencies = [
 [[package]]
 name = "annotate-snippets"
 version = "0.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c96c3d1062ea7101741480185a6a1275eab01cbe8b20e378d1311bc056d2e08"
+source = "git+https://github.com/magurotuna/annotate-snippets-rs?branch=check-display-width#bb47ac0e0eff5d2b9b2694829c477549c36d16ae"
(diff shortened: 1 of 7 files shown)


bartlomieju (approved):
LGTM! Thanks!

PR #842, opened 2021-09-06
fix: ignore `ban-unknown-rule-code` error without this rule

Brooooooklyn:


bartlomieju:
@magurotuna please take a look

src/rules/ban_unknown_rule_code.rs
```diff
@@ -7,6 +7,8 @@ use crate::{Program, ProgramRef};
 #[derive(Debug)]
 pub struct BanUnknownRuleCode;
 
+pub(crate) static CODE: &str = "ban-unknown-rule-code";
```
magurotuna, line 10: I know `static` also works, but in other rule implementations we use `const` to declare the rule `CODE`. So for consistency, it'd be great to use `const` here too.

src/linter.rs
```diff
@@ -141,10 +149,24 @@ impl Linter {
...
+
+    if self.check_unknown_rules
+      && !context
+        .file_ignore_directive()
+        .map(|f| f.has_code(ban_unknown_rule_code::CODE))
+        .unwrap_or(false)
+    {
+      filtered_diagnostics.extend(unknown_used_diagnostics);
+    }
+
```
magurotuna, line 167: Can we move all the logic associated with `ban-unknown-rule-code` into `Context::ban_unknown_rule_code`?
I mean, it should be better if `Context::ban_unknown_rule_code` also processes the flag on whether the ignore directive is used or not, and just returns diagnostics if any. In `filter_diagnostics`, we would just do:

```rs
filtered_diagnostics.extend(context.ban_unknown_rule_code());
```

like `ban-used-ignore` does.

src/linter.rs
```diff
@@ -91,13 +95,17 @@ impl Linter {
...
+      .iter()
+      .any(|a| a.code() == (BanUnknownRuleCode).code());
     Linter {
       ast_parser: AstParser::new(),
       ignore_file_directive,
       ignore_diagnostic_directive,
       syntax,
       rules,
       plugins,
+      check_unknown_rules,
```
magurotuna, line 108: In my opinion, I don't want the `Linter` struct to have `check_unknown_rules`. Can we store this boolean in `Context` instead?

src/lib.rs
```diff
@@ -92,6 +92,33 @@ mod lint_tests {
     assert_diagnostic(&diagnostics[1], "ban-unknown-rule-code", 4, 3, src);
   }
 
+  #[test]
+  fn no_ban_unknown_rule_code_error_without_rule() {
```
magurotuna, line 96: This test name sounds a little bit unclear to me. How about `dont_run_ban_unknown_rule_code_when_no_rules_provided` or something?

src/lib.rs
```diff
@@ -92,6 +92,33 @@ mod lint_tests {
...
+    let src = r#"
+ // deno-lint-ignore some-rule
+ function _foo() {
+   // deno-lint-ignore some-rule-2 some-rule-3
+   let _bar_foo = true
+ }
+      "#;
+    let diagnostics = lint(src, Arc::new(vec![]));
+    assert!(&diagnostics.is_empty());
```
magurotuna, line 105:
```suggestion
    assert!(diagnostics.is_empty());
```

src/lib.rs
```diff
@@ -92,6 +92,33 @@ mod lint_tests {
...
+    let src = r#"
+// deno-lint-ignore-file ban-unknown-rule-code
+
+// deno-lint-ignore some-rule
+export function foo() {
+  return true
+}
+      "#;
+    let diagnostics = lint_recommended_rules(src);
+    assert!(&diagnostics.is_empty());
```
magurotuna, line 119:
```suggestion
    assert!(diagnostics.is_empty());
```

magurotuna (changes requested):
This PR seems to be not so trivial that it needs no explanation, so if you gave a summary about what this PR attempts to fix, that would be much appreciated.
Anyway, as far as I understand, this PR does:

1. if the given set of rules doesn't have `ban-unknown-rule-code`, skip applying this rule
2. if the file-level ignore directive contains `ban-unknown-rule-code`, skip applying this rule

As for 1, I think this works "as intended" at least right now; `ban-unknown-rule-code`, as well as `ban-unused-ignore`, are always run even if no rules are specified. There is a historical reason behind it and it should be addressed in the long run, but in my opinion, ultimately these two rules should be implemented in `LintRule` trait somehow. We would need to do a little bit of large refactoring to achieve it, so yeah your solution looks good for the time being.

As for 2, I forgot about it when I implemented it! Thanks for the good catch. There are just a few comments about code design though.

Brooooooklyn:
> This PR seems to be not so trivial that it needs no explanation, so if you gave a summary about what this PR attempts to fix, that would be much appreciated.

I'm working on https://github.com/napi-rs/node-rs/tree/main/packages/deno-lint and trying to integrate it to one of my project.
I've been set the default ignore directive to `eslint-disable/eslint-disable-next-line` for compatible: https://github.com/napi-rs/node-rs/blob/f963c7cd2eb70070fb6c34c30574fe059356d750/packages/deno-lint/src/lib.rs#L181-L182

So I need a mechanism to disable the `ban-unknown-rule-code` when it's not been set.

bartlomieju:
@magurotuna are you happy with this solution or would you prefer to go with #847?

magurotuna:
@Brooooooklyn thanks for the explanation! Makes a lot of sense.

@bartlomieju sorry for the late reply I think this is fine for the moment. We can work on #847 later as a follow up.

magurotuna (approved):
LGTM

Brooooooklyn:
@bartlomieju Let's merge and release this one?

bartlomieju:
> @bartlomieju Let's merge and release this one?

@Brooooooklyn sorry for slow review, I actually did prerequisites for #847 and intend to implement it tonight, so we can rewrite `ban-unknown-rule-code` to being a proper rule. I'll try to release it in a week.

Brooooooklyn:
Okay, it's fine. 

> I'll try to release it in a week.

Looking forward it

Brooooooklyn:
@bartlomieju rebased

bartlomieju:
@Brooooooklyn please sign CLA

PR #844, opened 2021-09-06
chore(ci): only run lint on ubuntu and improve scaffold integration test

### LINKED ISSUE #841: `cargo check` in scaffold integration test causes an extra build on the CI
The `cargo check` in the scaffold integration test is slow on the CI because the CI will build in release mode and then run this in debug. We should consider some improvement here (maybe even just run with `--release`).

https://github.com/denoland/deno_lint/blob/d11ee57c557128567386e30af75e8d444485d176/tools/tests/scaffold_integration_test.ts#L35

https://github.com/denoland/deno_lint/runs/3519440748?check_suite_focus=true#step:11:372

dsherret:
I don't think it's worth it to run the linter on every platform given that there is basically zero platform specific code.

Additionally, closes #841.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,7 @@ jobs:
           deno test --unstable --allow-read=. --allow-write=. --allow-run ./tools
 
       - name: Lint
+        if: contains(matrix.os, 'ubuntu')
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, thanks!

PR #845, opened 2021-09-06
BREAKING: `rules::get_all_rules` etc. return `LintRule`s wrapped in `Arc`

magurotuna:
This PR changes the return types of `get_all_rules`, `get_recommended_rules`, and `get_filtered_rules` to `Arc`-wrapped values.

This will allow us to instantiate `LintRule`s just once before spawning worker threads, and then they are sharable across threads. Considering that instantiating rules is currently performed once per one file, which can be a significant amount of times if there are numerous lint target files, this will contribute to performance to some extent.


src/rules.rs
```diff
@@ -233,89 +303,32 @@ pub fn get_all_rules() -> Vec<Box<dyn LintRule>> {
...
 
   #[test]
   fn recommended_rules_sorted_alphabetically() {
-    let mut recommended_rules = get_recommended_rules();
-    recommended_rules.sort_by_key(|r| r.code());
-    for (sorted, unsorted) in
-      recommended_rules.into_iter().zip(get_recommended_rules())
+    let sorted_recommended_rules = {
+      let mut recommended_rules =
+        Arc::try_unwrap(get_recommended_rules()).unwrap();
```
magurotuna, line 314: A bunch of `#[derive(Debug)]` for all the lint structs are required to call `unwrap` here.

examples/dlint/js.rs
```diff
@@ -103,14 +103,26 @@ fn op_query_control_flow_by_span(
...
+    Box::new(Self {
+      plugin_path: plugin_path.to_string(),
+    })
+  }
+}
+
+struct JsRunner {
   runtime: JsRuntime,
   module_id: i32,
 }
```
magurotuna, line 122: Previous `JsRuleRunner` struct can neither be `Send` nor `Sync` because of `JsRuntime`. So I split it into two, one is `PluginRunner`, which is thread safe, the other is `JsRunner`, which is not.
What we'll share across worker threads is `PluginRunner` only.

bartlomieju (approved):
LGTM, thanks @magurotuna this is very desired change!

Issue #745, opened 2021-06-20
Feature Request: lint rule for empty enums? (including const enums)


somebody1234:
considering there's already a lint rule for empty interfaces, i think this would be not a bad idea at least

bartlomieju:
`const enum` is not supported in Deno anyway, but I'm open to adding rule for empty enums.

DmitryRomaniuk:
I'll pick up that issue :)

satyarohith:
Closed by #848

PR #848, opened 2021-09-07
add lint rule for empty enums

DmitryRomaniuk:
https://github.com/denoland/deno_lint/issues/745

### DIFF
--- a/docs/rules/no_empty_enum.md
+++ b/docs/rules/no_empty_enum.md
@@ -1,7 +1,7 @@
 Disallows the declaration of an empty enum
 
-An enum with no members serves no purpose. This rule will capture
-these situations as either unnecessary code or a mistaken empty implementation.
+An enum with no members serves no purpose. This rule will capture these
+situations as either unnecessary code or a mistaken empty implementation.
 
 ### Invalid:
 
@@ -13,6 +13,6 @@ enum Foo {}
 
 ```typescript
 enum Foo {
-  ONE = 'ONE'
+  ONE = "ONE",
 }
 ```
--- a/docs/rules/no_empty_enum.md
+++ b/docs/rules/no_empty_enum.md
@@ -7,7 +7,6 @@ these situations as either unnecessary code or a mistaken empty implementation.
 
 ```typescript
 enum Foo {}
-const enum Foo {}
 ```
 
 ### Valid:
@@ -15,9 +14,5 @@ const enum Foo {}
 ```typescript
 enum Foo {
   ONE = 'ONE'
-}
-
-const enum Bar {
-  TWO = 'TWO'
 }
 ```
--- a/docs/rules/no_empty_enum.md
+++ b/docs/rules/no_empty_enum.md
@@ -1,0 +1,23 @@
+Disallows the declaration of an empty enum
+
+An enum with no members serves no purpose. This rule will capture
+these situations as either unnecessary code or a mistaken empty implementation.
+
+### Invalid:
+
+```typescript
+enum Foo {}
+const enum Foo {}
+```
+
+### Valid:
+
+```typescript
+enum Foo {
+  ONE = 'ONE'
+}
+
+const enum Bar {
+  TWO = 'TWO'
+}
+```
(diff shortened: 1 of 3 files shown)


src/rules/no_empty_enum.rs
```diff
@@ -0,0 +1,120 @@
...
+  }
+}
+
+impl<'c, 'view> Visit for NoEmptyEnumVisitor<'c, 'view> {
+  fn visit_ts_enum_decl(
+    &mut self,
+    enum_decl: &TsEnumDecl,
+    _parent: &dyn Node,
+  ) {
+    if enum_decl.members.len() < 1 {
```
bartlomieju, line 62:
```suggestion
    if enum_decl.members.is_empty() {
```

docs/rules/no_empty_enum.md
```diff
@@ -0,0 +1,23 @@
+Disallows the declaration of an empty enum
+
+An enum with no members serves no purpose. This rule will capture
+these situations as either unnecessary code or a mistaken empty implementation.
+
+### Invalid:
+
+```typescript
+enum Foo {}
+const enum Foo {}
```
bartlomieju, line 10: Deno doesn't support `const enum`, I suggest to remove this from example as it might be confusing

bartlomieju (commented):
Thanks Dmitry! Looking good.

docs/rules/no_empty_enum.md
DmitryRomaniuk: thx, fixed

src/rules/no_empty_enum.rs
DmitryRomaniuk: thx, fixed

src/rules/no_empty_enum.rs
```diff
@@ -0,0 +1,120 @@
...
+    if enum_decl.members.is_empty() {
+      self.context.add_diagnostic_with_hint(
+        enum_decl.span,
+        CODE,
+        MESSAGE,
+        HINT,
+      );
+    }
+  }
+}
```
magurotuna, line 71: If we use `Visit` trait, we have to explicitly call a method to traverse AST nodes further. This means that the current code won't capture the error in the following example:

```ts
enum Foo {
  One = 1,
  Two = (() => {
    enum Bar {} // this won't be detected!
    return 42;
  })(),
}
```

Although this is a valid TypeScript, of course people might not write such a tricky code, but it should be better for the linter to detect it as well.
To address this issue, I recommend to use `VisitAll` trait instead. With this trait, we won't need to do something explicitly, the visitor will traverse all the nodes automatically.
(Or you can also use `Handler` trait - if you're interested in it, see the concrete example, like https://github.com/denoland/deno_lint/blob/9669c247be0e6ce4fcd376acdfd389251e83ed8a/src/rules/no_with.rs)

magurotuna (commented):
Nice work!

src/rules/no_empty_enum.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::ProgramRef;
+use deno_ast::swc::ast::TsEnumDecl;
+use deno_ast::swc::visit::Node;
+use deno_ast::swc::visit::Visit;
+
+#[derive(Debug)]
+pub struct NoEmptyEnum;
+
+const CODE: &str = "no-empty-enum";
+const MESSAGE: &str = "An empty enum is equivalent to `{{}}`.";
```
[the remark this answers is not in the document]
DmitryRomaniuk: thx, fixed

src/rules/no_empty_enum.rs
DmitryRomaniuk: I can't use ```VisitAll```, it doesn't work, so I change to ```Handler``` but added if statement, could you review again plz @magurotuna and added that example to test (I've never seen that tricky code before and can't imagine)

src/rules/no_empty_enum.rs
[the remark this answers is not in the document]
DmitryRomaniuk: thx

src/rules/no_empty_enum.rs
DmitryRomaniuk: And ci failed on format step 😥

magurotuna: Now the implementation with `Handler` looks perfect 👍 

> And ci failed on format step 😥

We have a script for formatting the code. So just running `./tools/format.ts` will help you.

magurotuna (approved):
LGTM - thank you!

bartlomieju (commented):
LGTM, thanks Dmitry!

PR #851, opened 2021-09-09
fix(no-window-prefix): switch from allow-list to deny-list

### LINKED ISSUE #850: `no-window-prefix` is too aggressive
`no-window-prefix` makes `deno lint` annoying to use for web targeted projects. For example calling `window.history` is disallowed.

I think we should consider switching `no-window-prefix` from an allow-list (onload, onunload, closed, alert, confirm, prompt, localStorage, sessionStorage, window, Navigator), to a deny-list. The deny list would contain all the items that are known to exist on both web workers and window workers.

`no-window-prefix` would only trigger if an API in the deny list is called. A quick script to generate this deny list from the d.ts files by getting the intersection of the two:

```js
import { doc } from "https://deno.land/x/deno_doc@v0.13.0/mod.ts";

const windowDoc = await doc("https://raw.githubusercontent.com/denoland/deno/main/cli/dts/lib.dom.d.ts");
const workerDoc = await doc("https://raw.githubusercontent.com/denoland/deno/main/cli/dts/lib.webworker.d.ts");

const windowItems = new Set(windowDoc.map((item) => item.name));
const workerItems = new Set(workerDoc.map((item) => item.name));

const intersection = new Set(
  [...windowItems].filter((x) => workerItems.has(x)),
);
intersection.add("Deno");

console.log(JSON.stringify([...intersection], null, 2));
```

This deny list should probably also be extended to cover all ECMAscript builtins.

So to clarify, the rule should hit if it sees the following: `window.${item in deny list}`.

deno 1.13.2+a5bcf70 (canary, x86_64-unknown-linux-gnu)

magurotuna:
This lets `no-window-prefix` rule use a deny-list instead of allow-list when checking if the property is considered valid or not.

It's possible to check if the deny-list is always up-to-date by running `tools/generate_no_window_prefix_deny_list.ts` in CI, but I am not sure if such a rigid approach is needed.

Fixes #850


bartlomieju (commented):
@lucacasonato does this change satisfy your needs?

bartlomieju (approved):
LGTM, thank you @magurotuna

Issue #854, opened 2021-09-10
`no-invalid-triple-slash-reference` rule incorrectly warns on `lib` triple slash references in `.js` files


jaydenseric:
TypeScript `lib` triple slash references should be allowed in `.js` files, but incorrectly result in `no-invalid-triple-slash-reference` rule warnings. See:

https://github.com/denoland/deno/discussions/11973

### Lint Name

`no-invalid-triple-slash-reference`

### Code Snippet

```js
// @ts-check
/// <reference lib="dom" />
/// <reference lib="deno.ns" />
```

### Expected Result

No lint errors.

### Actual Result

Lint errors:

```
This triple-slash reference directive is not allowed in JavaScript
In JavaScript only the `types` directive is allowed, like `/// <reference types="..." />`
```

### Additional Info

Until this is fixed, users can ignore the lint warnings:

```js
// @ts-check
// deno-lint-ignore no-invalid-triple-slash-reference
/// <reference lib="dom" />
// deno-lint-ignore no-invalid-triple-slash-reference
/// <reference lib="deno.ns" />
```

### Version

deno 1.13.2 (release, x86_64-apple-darwin)
v8 9.3.345.11
typescript 4.3.5

kitsonk:
I don't know if we want to detect the ts-check or not, I suspect we would still get false positives if check-js is on in a config file. Maybe best to just "allow" the library reference in JS files.

PR #855, opened 2021-09-10
fix: 'no-invalid-triple-slash-referecence' for JavaScript files

bartlomieju:
This commit allows to use `/// <reference lib="lib" />` in JavaScript files.

Fixes https://github.com/denoland/deno_lint/issues/854

### DIFF
--- a/src/rules/no_invalid_triple_slash_reference.rs
+++ b/src/rules/no_invalid_triple_slash_reference.rs
@@ -65,9 +65,9 @@ fn is_js_or_jsx(filename: &str) -> bool {
 
 #[derive(Debug, Eq, PartialEq)]
 enum ReportKind {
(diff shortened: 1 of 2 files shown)


bartlomieju:
@magurotuna please take a look

src/rules/no_invalid_triple_slash_reference.rs
```diff
@@ -133,10 +133,10 @@ fn check_comment(comment: &Comment, is_js_like: bool) -> Option<ReportKind> {
 
   if is_js_like {
     // In JavaScript, only the `types` directives are allowed
-    if is_types_ref(&comment.text) {
+    if is_types_ref(&comment.text) || is_lib_ref(&comment.text) {
```
magurotuna, line 136: Maybe the comment should also be changed to mention `lib` directives.

PR #857, opened 2021-09-11
feat(prefer-ascii): print character's unicode code point in hint

### LINKED ISSUE #852: prefer-ascii: Show `\x...` value on non-ascii character in hint
@magurotuna Great job !

A nice additional improvement would be to show the escaped `\x...` value

_Originally posted by @AaronO in https://github.com/denoland/deno/pull/9713#issuecomment-913021200_

magurotuna:
This PR adds a character's unicode code point to hint, and upgrades annotate-snippets to 0.9.1 so that diagnostics point to the right position when the target character is double-width.

Closes #852

### Before

image

### After
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,9 +34,9 @@ dependencies = [
 
 [[package]]
 name = "annotate-snippets"
-version = "0.9.0"
+version = "0.9.1"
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
LGTM, thanks

Issue #818, opened 2021-08-28
Tracking issue for switching to `dprint-swc-ecma-ast-view`


magurotuna:
Ref #571 

We are under switching to `dprint-swc-ecma-ast-view` in rules implementation. This is a tracking issue for it.

- [x] adjacent-overload-signatures
- [x] ban-ts-comment
- [x] ban-types
- [x] ban-unknown-rule-code
- [x] ban-untagged-ignore
- [x] ban-untagged-todo
- [x] ban-unused-ignore
- [x] camelcase
- [x] constructor-super
- [ ] default-param-last
- [ ] eqeqeq
- [ ] explicit-function-return-type
- [ ] explicit-module-boundary-types
- [ ] for-direction
- [ ] getter-return
- [ ] no-array-constructor
- [ ] no-async-promise-executor
- [x] no-await-in-loop
- [ ] no-case-declarations
- [ ] no-class-assign
- [ ] no-compare-neg-zero
- [ ] no-cond-assign
- [ ] no-const-assign
- [ ] no-constant-condition
- [ ] no-control-regex
- [ ] no-debugger
- [ ] no-delete-var
- [x] no-deprecated-deno-api
- [ ] no-dupe-args
- [ ] no-dupe-class-members
- [ ] no-dupe-else-if
- [ ] no-dupe-keys
- [ ] no-duplicate-case
- [ ] no-empty
- [ ] no-empty-character-class
- [ ] no-empty-interface
- [ ] no-empty-pattern
- [ ] no-eval
- [ ] no-ex-assign
- [ ] no-explicit-any
- [ ] no-extra-boolean-cast
- [ ] no-extra-non-null-assertion
- [ ] no-extra-semi
- [ ] no-fallthrough
- [ ] no-func-assign
- [ ] no-global-assign
- [ ] no-import-assign
- [ ] no-inferrable-types
- [ ] no-inner-declarations
- [ ] no-invalid-regexp
- [x] no-invalid-triple-slash-reference
- [x] no-irregular-whitespace
- [ ] no-misused-new
- [x] no-namespace
- [ ] no-new-symbol
- [ ] no-non-null-asserted-optional-chain
- [ ] no-non-null-assertion
- [ ] no-obj-calls
- [ ] no-octal
- [ ] no-prototype-builtins
- [ ] no-redeclare
- [ ] no-regex-spaces
- [ ] no-self-assign
- [x] no-setter-return
- [ ] no-shadow-restricted-names
[...]

PR #892, opened 2021-09-15
Migrate default-param-last from Visitor to Handler

DmitryRomaniuk:
https://github.com/denoland/deno_lint/issues/818


bartlomieju (approved):
LGTM!

PR #894, opened 2021-09-17
Fix no_extra_semi docs

bakkot:
In https://github.com/denoland/deno_lint/pull/787, the docs for no_extra_semi appear to have had the `valid` case copied to both the `invalid` and `valid` examples. This restores the `invalid` example to the original prior to #787.


bakkot:
Lol, just saw this is duplicated with #894, oops.

Issue #847, opened 2021-09-07
Proposal: Control the execution order of `LintRule` trait


magurotuna:
### Context

Currently, `ban-unused-ignore` and `ban-unknown-rule-code` are handled specially since these rules have to be run after other rules. Moreover, there is a priority even in these two rules; `ban-unused-ignore` has to be run after `ban-unknown-rule-code`.
In short, we have to run lint rules in the following order:

1. normal rules (other than the two rules below) + plugins
2. `ban-unknown-rule-code`
3. `ban-unused-ignore`

To do this, the logics of these special rules are not implemented in `LintRule` trait, which I think may lead to the unnecessarily increasing complexity of the codebase.

### Solution

It would be a solution to add the information about the order of execution to `LintRule` trait. So `LintRule` trait would look like:

```rust
pub trait LintRule: std::fmt::Debug + Send + Sync {
  fn new() -> Box<Self> where Self: Sized;

  fn lint_program<'view>(
    &self,
    context: &mut Context<'view>,
    program: ProgramRef<'view>,
  );

  fn lint_program_with_ast_view<'view>(
    &self,
    context: &mut Context<'view>,
    program: Program<'view>,
  );

  fn code(&self) -> &'static str;

  fn tags(&self) -> &'static [&'static str];

  #[cfg(feature = "docs")]
  fn docs(&self) -> &'static str;

  // New method, which returns the priority of this rule.
  // The lower the return value is, the earlier this rule will be run.
  fn priority(&self) -> u32;
}
```

and then set the priorities like the following:

|rule type|priority|
|---------|-------|
|normal rules + plugins|0|
|`ban-unknown-rule-code`|1|
|`ban-unused-ignore`|2|

With this information, we could sort the execution order or lint rules, and would no longer need to implement the two rules in a special way. Also, if someday we have another special rule that needs to be run, say, between `ban-unknown-rule-code` and `ban-unused-ignore`, all we have to do would be just adjust the priority numbers.

bartlomieju:
I'm in favor of this change 👍  thanks for a great write up @magurotuna. If we can make `ban-unknown-rule-code` and `ban-unused-ignore` into regular rules that's highly preferable.

bartlomieju:
Unfortunately this change clashes a bit with #845; we can't borrow lint rules list mutable because it's behind and `Arc`. I think we might want to change #845 with use `Vec<Arc<dyn LintRule>>` instead.

PR #901, opened 2021-09-19
refactor: store LintRule in an Arc instead of Box

bartlomieju:
Follow up to https://github.com/denoland/deno_lint/pull/845

As described in https://github.com/denoland/deno_lint/issues/847#issuecomment-922381168 having a list of rules being `Arc<Vec<_>>` makes it hard to use and some operations become very cumbersome to achieve. So instead of storing trait objects of `LintRule` in `Box`es we store them in `Arc`s.

CC @magurotuna

### DIFF
--- a/examples/dlint/config.rs
+++ b/examples/dlint/config.rs
@@ -30,7 +30,7 @@ pub struct Config {
 }
 
 impl Config {
-  pub fn get_rules(&self) -> Arc<Vec<Box<dyn LintRule>>> {
+  pub fn get_rules(&self) -> Vec<Arc<dyn LintRule>> {
     get_filtered_rules(
       Some(self.rules.tags.clone()),
       Some(self.rules.exclude.clone()),
@@ -158,7 +158,7 @@ mod tests {
     }}
   }
 
-  fn into_codes(rules: Arc<Vec<Box<dyn LintRule>>>) -> HashSet<&'static str> {
+  fn into_codes(rules: Vec<Arc<dyn LintRule>>) -> HashSet<&'static str> {
     rules.iter().map(|rule| rule.code()).collect()
   }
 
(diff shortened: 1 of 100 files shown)


magurotuna:
I don't really understand why `Arc<Vec<_>>` are hard to use for some operations. Could you elaborate on it?
As for #847, I think `Arc<Vec<_>>` should still work - all we have to do is to return the list of rules sorted by priority from `get_all_rules`, `get_recommended_rules`, and `get_filtered_rules`. After it's returned from these functions, mutable access to it is no longer necessary, isn't it?

bartlomieju:
> I don't really understand why `Arc<Vec<_>>` are hard to use for some operations. Could you elaborate on it?

Sure - if vector is behind an `Arc` then we can't really use any mutating methods (like `sort_by_key()`) because many consumers may hold this `Arc`. We could use `Arc::try_unwrap` but it only works if there's a single owner of the `Arc` and partially misses the point.

> As for #847, I think `Arc<Vec<_>>` should still work - all we have to do is to return the list of rules sorted by priority from `get_all_rules`, `get_recommended_rules`, and `get_filtered_rules`. After it's returned from these functions, mutable access to it is no longer necessary, isn't it?

That's true, but if we did that sorting in methods you listed then all of them would be sorted alphabetically besides `ban-unknown-rule-code` and `ban-unused-ignore` - so they would effectively be displayed at the bottom of the list. The priority of rules is internal detail of the linter and users shouldn't care about it - they still want to see rules sorted alphabetically. So I wanted to implement running by priority as an internal sort method done just before running rules. Another point is that rules coming from plugins will be separate vectors and we can't enforce their sorting until they are loaded.

Another point is that `Arc<Vec<Box<dyn LintRule>>>` is quite a complex type, much more than `Vec<Arc<dyn LintRule>>`. There's no problem in copying a vector of hundred arcs, it should still be super cheap operation, and would make handling of them more ergonomic.

magurotuna:
@bartlomieju Thanks for the explanation. Makes much sense to me.

magurotuna (approved):
LGTM

PR #902, opened 2021-09-22
feat: Add LintRule::priority()

bartlomieju:
Closes https://github.com/denoland/deno_lint/issues/847

This commit adds new method to "LintRule" trait called "priority".
The lower the priority the earlier the rule will be run. This value
defaults to 0 and most rules need not change it.

### DIFF
--- a/src/linter.rs
+++ b/src/linter.rs
@@ -14,6 +14,7 @@ use deno_ast::swc::common::SyntaxContext;
 use deno_ast::swc::parser::Syntax;
 use deno_ast::view::ProgramRef;
 use deno_ast::ParsedSource;
+use std::cmp::Ordering;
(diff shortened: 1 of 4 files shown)


src/linter.rs
```diff
@@ -192,11 +192,15 @@ impl Linter {
         top_level_ctxt,
       );
 
+      let mut sorted_rules = self.rules.clone();
+      crate::rules::sort_rules_by_priority(&mut sorted_rules);
```
magurotuna, line 196: Is this clone necessary? It seems to me that `crate::rules::sort_rules_by_priority(&mut self.rules)` will just work fine.

bartlomieju: Good catch

PR #913, opened 2021-09-26
Migrate no_cond_assign from Visitor to Handler

DmitryRomaniuk:
Part of https://github.com/denoland/deno_lint/issues/818
I added method to scopes

### DIFF
--- a/src/rules/no_const_assign.rs
+++ b/src/rules/no_const_assign.rs
@@ -59,74 +59,53 @@ impl LintRule for NoConstAssign {
 
 struct NoConstAssignHandler;
 
-impl NoConstAssignHandler {
-  fn check_pat(&mut self, pat: &Pat, span: Span, ctx: &mut Context) {
-    match pat {
-      Pat::Ident(ident) => {
-        self.check_scope_for_const(span, ident.id, ctx);
-      }
-      Pat::Assign(assign) => {
-        self.check_pat(&assign.left, span, ctx);
-      }
-      Pat::Array(array) => {
-        self.check_array_pat(array, span, ctx);
-      }
-      Pat::Object(object) => {
-        self.check_obj_pat(object, span, ctx);
-      }
-      _ => {}
+fn check_pat(pat: &Pat, span: Span, ctx: &mut Context) {
+  match pat {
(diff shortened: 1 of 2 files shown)


src/scopes.rs
```diff
@@ -61,6 +61,12 @@ impl Scope {
     self.vars.get(id)
   }
 
+  pub fn var_by_ident(&self, ident: &ast_view::Ident) -> Option<&Var> {
+    self
+      .vars
+      .get(&(ident.inner.sym.clone(), ident.inner.span.ctxt()))
+  }
```
magurotuna, line 68: This surely works, but how about the following? It looks more idiomatic IMO.

```suggestion
  pub fn var_by_ident(&self, ident: &ast_view::Ident) -> Option<&Var> {
    self.vars.get(&ident.inner.to_id())
  }
```

src/rules/no_const_assign.rs
```diff
@@ -59,66 +57,70 @@ impl LintRule for NoConstAssign {
...
-
-impl<'c, 'view> NoConstAssignVisitor<'c, 'view> {
-  fn new(context: &'c mut Context<'view>) -> Self {
-    Self { context }
-  }
+struct NoConstAssignHandler;
 
-  fn check_pat(&mut self, pat: &Pat, span: Span) {
+impl NoConstAssignHandler {
+  fn check_pat(&mut self, pat: &Pat, span: Span, ctx: &mut Context) {
```
magurotuna, line 63: Now these `check_*` methods can be changed to take `&self` instead of `&mut self`, right? If possible, it would be preferable.
Or it even looks like we could write these methods as normal functions rather than methods, since they don't depend on `self`. It's a matter of preference, but I personally prefer to have them as normal functions.

src/scopes.rs
DmitryRomaniuk: thanks, but I don't understand why ctx.scope().var(&name.inner.to_id()) doesn't work :( in no_const_assign.ts file

DmitryRomaniuk: I fixed

bartlomieju (approved):
LGTM, thanks!

PR #919, opened 2021-09-28
fix: bring back lint.deno.land

satyarohith:
- fix: pin fresh to a commit
- fix: enable client js hydration
- chore: deno fmt

### DIFF
--- a/www/deps.ts
+++ b/www/deps.ts
@@ -1,7 +1,7 @@
-export * from "https://raw.githubusercontent.com/lucacasonato/fresh/main/runtime.ts";
+export * from "https://raw.githubusercontent.com/lucacasonato/fresh/5bc35b8b955c143654d22936ad5618274cbe2a28/runtime.ts";
 import MarkdownIt from "https://dev.jspm.io/markdown-it@12.0.4";
 export { MarkdownIt };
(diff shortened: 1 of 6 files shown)


lucacasonato:
Now all working. I disabled runtimeJS now. @satyarohith PTAL

lucacasonato (approved):
Approval from me. If you are happy with it @satyarohith, please merge.

satyarohith (commented):
Thanks for introducing `<Head />`, Luca. :)

Few thoughts on the new design:
- The dark mode's description area feels "too dark".
- Should we have the expand button below the short description named "view more"?

CleanShot 2021-10-04 at 18 05 47@2x

CleanShot 2021-10-04 at 17 53 17@2x

lucacasonato:
> The dark mode's description area feels "too dark".

It is the GitHub dark theme colors. I think it would look nicer if the surrounding background was also dark, but I was too lazy to implement that right now. I did make the surrounding background dark in the `ignoring-rules` page.

> Should we have the expand button below the short description named "view more"?

Can do. This is just the default `<details>` + `<summary>` styling, and I didn't want to get into the woods with CSS today :sweat_smile:

satyarohith:
I'll merge it then; we can do the edits in a follow-up.

Issue #779, opened 2021-07-18
[Suggestion] Lint rule to disallow `arr[arr.length - 1]`


grian32:
With the new `arr.at(-1)` syntax introduced V8 9.2, perhaps it would be a good idea to have a rule to disallow `arr[arr.length - 1]` syntax, ass `arr.at(-1)` is imo cleaner and was made for this purpose more or less.

This would of course have to wait for 1.12.1 or whatever version fixes the issue that doesn't let ts recognize `.at`

bartlomieju:
This sounds good to me 👍

crowlKats:
performance wise though .at() is quite slower

bartlomieju:
> performance wise though .at() is quite slower

Good to know, I'm sure V8 team will work on the performance in the coming release if this API becomes prevalent.

grian32:
On the topic of performance, I've made a rudimentary JSBench here that displays the difference: https://jsbench.me/x1krswau85

I will also attempt to implement this rule and PR if I am succesful.

grian32:
> I will also attempt to implement this rule and PR if I am succesful.

I was not succesfull, if someone else would like to implement this feel free to.

subhakundu:
Is it okay, if I start working on this one? It will give me good opportunity to learn.

magurotuna:
@subhakundu 
Sure! Please do, and as always feel free to ask about anything. We're sorry that there's no comprehensive documentation about implementing a new lint rule right now (of course I wanna do though). I'd recommend looking at other rules implementation.

subhakundu:
@magurotuna is there any routine to print ASTs (from top level to lower level) so that I can identify the type of node to use for VisitAll? I have created a barebone rule for learning how to create one. Next steps are to identify type of node (which is potentially Expr I guess) and traversing down from there till we can make sure it is arr.length, binary operation '-' and '1'. It requires more AST information. I am new to SWC, so, this will be really helpful. I am now using dbg! macro. 

I will require name, code, message, hint and doc text in future, but we can discuss that later I guess. I am focusing on creating the rule now.

subhakundu:
Hi @magurotuna I require sometime to work on it. I am in some kind of personal emergency. Once that is sorted, I will start again. Is it okay? Else if someone wants to work on it, please go ahead.

magurotuna:
@subhakundu No problem, take your time 😉

subhakundu:
[...]

PR #932, opened 2021-10-06
feature: adding Lint rule in favor of arr.at(-1) (#779)

subhakundu:
I have created a pull request for the lining rule mentioned. Right now, the following things are considered.
1. a basic regex is validating if the pattern is matching or not. It can get complicated based on the scope. It will be of great help if the reviewer can help me with other use cases.
2. Some basic UTs are added. More UTs are needed to be added based on the scope of the rule.
3. A copied documentation file is added. The contents are needed to be finalized.
4. Message, code, and hint need to be finalized.
5. We need to decide if the rule is "recommended" or not at this moment.

Please consider it a basic POC. Please let me know the pointers I need to incorporate to make the rule robust. There are also some unused import. I will work on those during revision.

Testing Details:

Input:
```
const fruits = ['Apple', 'Banana']
const _x = fruits[fruits.length-1];
```

Output:
```
error[ban-array-length-minus-one]: arr[arr.length - 1] is deprecated.
 --> /Users/subhakundu/JS_Examples/array_size.ts:2:19
  |
2 | const _x = fruits[fruits.length-1];
  |                   ^^^^^^^^^^^^^^^
  |
  = help: Please consider using arr.at(-1) instead of arr[arr.length - 1]
Found 1 problem
```

### DIFF
--- a/docs/rules/ban_array_length_minus_one.md
+++ b/docs/rules/ban_array_length_minus_one.md
@@ -1,0 +1,89 @@
+Disallows assigning variables to `this`.
+
+In most cases, storing a reference to `this` in a variable could be avoided by
+using arrow functions properly, since they establish `this` based on the scope
+where the arrow function is defined.
+
+Let's take a look at a concrete example:
+
+```typescript
+const obj = {
+  count: 0,
+  doSomethingLater() {
+    setTimeout(function () { // this function executes on the global scope; `this` evalutes to `globalThis`
+      this.count++;
+      console.log(this.count);
+    }, 300);
+  },
+};
+
+obj.doSomethingLater();
+// `NaN` is printed, because the property `count` is not in the global scope.
+```
+
+In the above example, `this` in the function passed to `setTimeout` evaluates to
+`globalThis`, which results in the expected value `1` not being printed.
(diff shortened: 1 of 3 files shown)


magurotuna (commented):
I think the approach of regex matching is less robust. We should probably handle them in a purely AST-based way. 
You may want to an existing example that shows how to handle member expressions like `foo.bar`. https://github.com/denoland/deno_lint/blob/main/src/rules/no_deprecated_deno_api.rs#L166-L192 might be helpful.

Also, it'd be better to not limit this rule to `-1` only in my opinion. For example `arr[arr.length - 2]` could be replaced with `arr.at(-2)`.

petamoriken:
In my opinion, it's better to target only `arr[arr.length - <integer numeric literal>]`. This is because `Array#at` rounds the argument to an integer value, unlike the usual `[[Get]]`.

```js
const arr = [1, 2, 3];
arr.at(0.5); // arr[0]
arr.at(NaN); // arr[0]
```

If some identifier or expression is mixed in with it, the behavior will be completely different.

ry:
Closing because stale

Issue #970, opened 2021-11-09
Add lint for ensuring a JSDoc exists for public (exported) APIs


bartlomieju:
Ref https://github.com/denoland/deno_std/issues/650

It's useful to automatically ensure that all public APIs have appropriate JSDocs, I believe an additional lint rule is the way to go for this purpose.

Skillz4Killz:
As a library developer this is something really important to me. I stopped using deno lint a while ago because of the headaches i had just deno-lint-ignoring so many lines. However, having proper documentation would be worth that trouble because it helps save me time since anything undocumented means i have to explain things to users when they ask how something works.

**Ideal Functionality:**

```ts
// PASS
/** does something */

// PASS BUT NOT REQUIRED
/**
  * does something
  * @param bar string a string to run this function on
  * @returns string
*/

// FAIL
/** */

// FAIL
// does something

// FAIL SINCE NO COMMENT 
export function foo(bar: string) {
  return bar;
}
```

I would prefer`/** does something */` be considered enough because i use TS and i believe the types + clear param names are enough and extra docs on them is just redundant and lowers productivity. However, i would suggest an option for this since JS devs might want this.

kitsonk:
Do we expect this to be on by default?

This can quickly get very annoying (like when we had the lint rule about requiring a return type annotation).  For example if I write a utility function in a utility module, it isn't part of my public API, but I export it, and if it is well named and has well named parameters, JSDoc'ing that function or method is _useless_.

bartlomieju:
> Do we expect this to be on by default?
> 
> This can quickly get very annoying (like when we had the lint rule about requiring a return type annotation). For example if I write a utility function in a utility module, it isn't part of my public API, but I export it, and if it is well named and has well named parameters, JSDoc'ing that function or method is _useless_.

In the beginning definitely yes, but I would love to have it on for `deno` repo by default. Ideally we could land on a solution that doesn't involve config, but I'm afraid in this case configuration might be inevitable.

kitsonk:
I am :-1: for having it if it is on be default.  I can see it useful for `std`, but for end users, it would likely make things too annoying.

bartlomieju:
> I am 👎 for having it if it is on be default. I can see it useful for `std`, but for end users, it would likely make things too annoying.

Okay sure, let's add the rule and try it out on some real life projects before deciding one way or the other.

iuioiua:
[...]

PR #972, opened 2021-11-15
feat: add require-jsdoc rule

ah-yu:
ref #970


src/rules/require_jsdoc.rs
```diff
@@ -0,0 +1,97 @@
...
+  ) {
+    if let DefaultDecl::Fn(_) = _n.decl {
+      if !check_jsdoc_exsit(_ctx.leading_comments_at(_n.span().lo)) {
+        _ctx.add_diagnostic(_n.span(), CODE, MESSAGE);
+      }
+    }
+  }
+}
+
+fn check_jsdoc_exsit<'c>(
```
dsherret, line 67: typo: `check_jsdoc_exists`

dsherret (commented):
@ah-yu it looks good so far, but I believe we need to do it for more than just functions? Also, I think perhaps this should also look at public members of classes/interfaces?

ah-yu:
added interfaces/classes jsdoc comment checks.
`eslint-plugin-jsdoc/require-jsdoc` only checks `FunctionDeclaration` by default, should we check classes/interfaces by default? 
ref: https://github.com/gajus/eslint-plugin-jsdoc#require

bartlomieju:
@ah-yu the PR looked okay, what's the reason to close it?

bartlomieju:
@dsherret please review, should we add more feature before landing?

ah-yu:
@bartlomieju Sorry for late response, I was kind of busy those days so I closed this PR. And I'd like to finish it.

iuioiua:
Hi @ah-yu, is there any way we could progress this PR? This change could be valuable for modules that'd like to enforce documentation.

sigmaSd:
It would be definitely great to have something like rust `#![warn(missing_docs)]` hopefully this gets picked up again, its really useful for library authors (and in my opinion it should be opt in like rust's)

dsherret:
Thanks, but we ended up going with a much more powerful linting system as part of `deno doc`. For details, see this PR https://github.com/denoland/deno/pull/21032

PR #973, opened 2021-11-17
fix: ignore class setters for explicit-function-return-type rule

### LINKED ISSUE #971: explicit-function-return-type colliding with set accessor
### Lint Name

explicit-function-return-type

### Code Snippet

```ts
class Test {
    #foo: string;

    set foo(value: string) {
        this.#foo = value;
    }
}
```

### Expected Result
Linter should pass

### Actual Result
Linter complains, that set should have a return type annotation

### Additional Info
When setting a return type annotation, linter passes, but compiler does not pass ("A 'set' accessor cannot have a return type annotation").

### Version

deno 1.16.1 (release, x86_64-unknown-linux-gnu)
v8 9.7.106.2
typescript 4.4.2

dsherret:
Closes #971


bartlomieju (approved):
LGTM

Issue #975, opened 2021-11-19
Lint Rule: No boolean literal for arguments.


l-cornelius-dol:
It would be great to have a lint rule to flag when `true` or `false` is passed to a function [edit] and that argument is not the only argument.

As a solution the problem of a vague true/false value passed to a function, the "boolean trap", I have a habit of creating a self-documenting constant and using that instead. Instead of `redraw(true,true)`, I'll define `const ALL_VIEWS = true, INLINE = true;` and then use `redraw(ALL_VIEWS,INLINE);`. [edit: added another argument].

I'd like to be able to flag when I or a member of our team forgets to do this.

Thanks for considering this.

bartlomieju:
I'm open to adding such rule, but it would be an optional rule (ie. required to enable it manually via a config file). PRs are welcome!

CRBl69:
If nobody is already working on this PR, I'd like to give it a try !

bartlomieju:
@CRBl69 please do!

jmj0502:
@bartlomieju Is this issue still open? If so, I would like to give it a try (I may need some context of the code that should be affected in order to introduce the functionality tho')

jmj0502:
@lawrence-dol Question. Do you want the rule to take effect in the following cases as well?
```typescript
invoke(true, remoteServerURL, true);
write(path, true, true);
```
Basically, in cases where there is a mix between named arguments and boolean literals or only in cases that match the example you previously provided?

jmj0502:
@lawrence-dol Perfect! Many thanks for the clarification. I'm almost done so the PR will probably be submitted soon!

jmj0502:
@lawrence-dol @bartlomieju The PR is already pending for review! Look forward to your comments!

jmj0502:
@lawrence-dol What you mention actually makes sense and I have no problem implementing it. When it comes to not allowing any boolean literals, I do agree because any floating literal seems like a magic value to me (IMO). I can proceed to implement the changes in a few minutes, but it I think is important to narrow down the expected behavior. 

To recap, the rule should point out any boolean literals passed as arguments to a function, even if it is a single one.

So the following examples:
```typescript
execute(true);
redraw(true, true);
executeCommand(true, EXECUTION_MODE.MODE_ONE);
```
Will all be marked as improvement points by the linter. Right?

Also, do you think the provided hint still works with these particular changes? Or do you recommend using other hint?

Look forward to your answer c:

jmj0502:
[...]

PR #976, opened 2021-11-23
feature: add no-boolean-literal-args lint rule

CRBl69:
This PR contains the lint rule asked for in issue #975. It will not allow you to use boolean literals (`true` and `false`) when you make a function call.

It is an opt-in (as requested) lint rule.

Previously, this would be allowed :

```typescript
foo(a: boolean) {
  console.log(a);
}

foo(true);
```

But with this lint rule, you have to do the following or else you will get a warning :

```typescript
foo(a: boolean) {
  console.log(a);
}

let bar = true;

foo(bar);
```

### DIFF
--- a/src/rules/no_boolean_literal_args.rs
+++ b/src/rules/no_boolean_literal_args.rs
@@ -1,0 +1,146 @@
+// Copyright 2020-2021 the Deno authors. All rights reserved. MIT license.
+use super::{Context, LintRule};
+use crate::handler::{Handler, Traverse};
+use crate::{Program, ProgramRef};
+use deno_ast::swc::common::Spanned;
+use deno_ast::view::{Expr, ExprOrSpread, Lit};
+use derive_more::Display;
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub struct NoBooleanLiteralArgs;
+
+const CODE: &str = "no-boolean-literal-args";
+
+#[derive(Display)]
+enum NoBooleanLiteralArgsMessage {
+  #[display(fmt = "Not allowed to use true or false as an argument")]
+  Unexpected,
+}
+
+#[derive(Display)]
+enum NoBooleanLiteralArgsHint {
+  #[display(fmt = "Create a boolean to pass to the function instead")]
+  CreateBoolean,
+}
+
+impl LintRule for NoBooleanLiteralArgs {
+  fn new() -> Arc<Self> {
+    Arc::new(NoBooleanLiteralArgs)
+  }
+
+  fn tags(&self) -> &'static [&'static str] {
+    &[]
+  }
+
+  fn code(&self) -> &'static str {
+    "no-boolean-literal-args"
+  }
+
+  fn lint_program(&self, _context: &mut Context, _program: ProgramRef) {
+    unreachable!();
+  }
+
+  fn lint_program_with_ast_view(
+    &self,
+    context: &mut Context,
+    program: Program,
+  ) {
+    let mut handler = NoBooleanLiteralArgsHandler::default();
+    handler.traverse(program, context);
+  }
+
+  #[cfg(feature = "docs")]
+  fn docs(&self) -> &'static str {
+    include_str!("../../docs/rules/no_boolean_literal_args.md")
+  }
+}
+
+#[derive(Default)]
+struct NoBooleanLiteralArgsHandler;
+
+impl NoBooleanLiteralArgsHandler {
+  fn check_args<'a>(&'a mut self, args: &[&ExprOrSpread], ctx: &mut Context) {
+    for arg in args {
+      if let Expr::Lit(Lit::Bool(_)) = &arg.expr {
+        ctx.add_diagnostic_with_hint(
+          arg.expr.span(),
+          CODE,
+          NoBooleanLiteralArgsMessage::Unexpected,
+          NoBooleanLiteralArgsHint::CreateBoolean,
+        );
+      }
+    }
+  }
+}
+
+impl Handler for NoBooleanLiteralArgsHandler {
+  fn call_expr(&mut self, expr: &deno_ast::view::CallExpr, ctx: &mut Context) {
+    self.check_args(&expr.args, ctx);
+  }
+
+  fn new_expr(&mut self, expr: &deno_ast::view::NewExpr, ctx: &mut Context) {
+    if let Some(args) = &expr.args {
+      self.check_args(args, ctx);
+    }
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+
+  #[test]
+  fn no_boolean_literal_args_valid() {
+    assert_lint_ok! {
+      NoBooleanLiteralArgs,
+      "let x = func(CONST_BOOL_1, 5, 3)",
+      "let x = func(trueBool, 0)",
+      "let x = func(y, 1, z)",
+      "func(falseBool, CONST_BOOL_1)",
+      "func(CONST_BOOL_1, func2(CONST_BOOL_2))",
+    };
+  }
+
+  #[test]
+  fn no_boolean_literal_args_invalid() {
+    assert_lint_err! {
+      NoBooleanLiteralArgs,
+      "func(true, a, b)": [
+        {
+          col: 5,
+          message: NoBooleanLiteralArgsMessage::Unexpected,
+          hint: NoBooleanLiteralArgsHint::CreateBoolean,
+        }
+      ],
+      "func(a, b, true)": [
+        {
+          col: 11,
+          message: NoBooleanLiteralArgsMessage::Unexpected,
+          hint: NoBooleanLiteralArgsHint::CreateBoolean,
+        }
+      ],
+      "let x = func(true)": [
+        {
+          col: 13,
+          message: NoBooleanLiteralArgsMessage::Unexpected,
+          hint: NoBooleanLiteralArgsHint::CreateBoolean,
+        }
+      ],
+      "let x = func(func2(false))": [
+        {
+          col: 19,
+          message: NoBooleanLiteralArgsMessage::Unexpected,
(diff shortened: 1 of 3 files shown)


lawrence-dol:
How can I track when this lands in `deno lint`?

MierenManz:
There should be a speaker icon or button on this page that says subscribe. Then you'll get notifications about this thread in github.
Ps: if you don't see a subscribe button but only a "unsubscribe" then you are already subscribed and you don't have to do anything

lawrence-dol:
@MierenManz : That let's me get notifications about this issue. But I want to know when this lands in `deno`, so I can begin using it. I've not found any way on github to cleanly track a specific commit being merged into a specific branch.

src/rules/no_boolean_literal_args.rs
```diff
@@ -0,0 +1,133 @@
...
+  fn docs(&self) -> &'static str {
+    include_str!("../../docs/rules/no_boolean_literal_args.md")
+  }
+}
+
+#[derive(Default)]
+struct NoBooleanLiteralArgsHandler;
+
+impl NoBooleanLiteralArgsHandler {
+  fn check_args<'a>(&'a mut self, params: &[&ExprOrSpread], ctx: &mut Context) {
```
dsherret, line 63: It says `params` here, but it seems these are actually arguments.

src/rules/no_boolean_literal_args.rs
```diff
@@ -0,0 +1,133 @@
...
+        );
+      }
+    }
+  }
+}
+
+impl Handler for NoBooleanLiteralArgsHandler {
+  fn call_expr(&mut self, expr: &deno_ast::view::CallExpr, ctx: &mut Context) {
+    self.check_args(&expr.args, ctx);
+  }
```
dsherret, line 80: Should we add this for new expressions as well? `new SomeClass(true)`

dsherret (commented):
This rule seems quite annoying for when calling external code outside the user's control. Perhaps this rule would be more useful if it were `no-boolean-literal-params` where it checks function, constructor, and method declarations for use of a boolean parameter instead of the call expressions? cc @bartlomieju

bartlomieju:
I agree with @dsherret, this rule should only consider declarations, not invocations of functions.

nayeemrmn:
I agree this is a superfluous and overly specific rule (is this a real problem with code and why not other kinds of literals?), but applying it to declarations doesn't make sense either. Hitting it when passing args to third party functions aligns with the intent fwiw. It's supposed to make you name the argument on the calling side.

dsherret:
> Hitting it when passing args to third party functions aligns with the intent fwiw. It's supposed to make you name the argument on the calling side.

Oh, I didn't realize that.

lawrence-dol:
> This rule seems quite annoying for when calling external code outside the user's control.

The rule is intended to require that the _calling_ code name boolean arguments and not simply pass true/false. E.g. `someFunction(true,true,false,true);`. Because booleans are a particularly egregious maintenance problem.

> I agree this is a superfluous and overly specific rule

Then don't enable it in _your_ ruleset.

> (is this a real problem with code...

Yes it's a real and well-known problem with code maintenance, and this simple rule can greatly improve the _use_ of APIs that utilize boolean parameters (which is not always a bad API in and of itself).

> ... and why not other kinds of literals?)

Because boolean arguments are unique in that they _always_ benefit from naming; other literals may or may not, depending on the context and API, and it's impossible to apply a blanket rule. For example, `let cur = add(cur,10);` does not benefit from `let cur = add(cur,TEN);`, but `repaint(data,true)` does _always_ benefit from either `repaint(data,IMMED)`, or `repaint(data,DEFERRED)`.

> but applying it to declarations doesn't make sense either. Hitting it when passing args to third party functions aligns with the intent fwiw. It's supposed to make you name the argument on the calling side.

That's 100% correct.

nayeemrmn:
> or example, `let cur = add(cur,10);` does not benefit from `let cur = add(cur,TEN);`, but `repaint(data,true)` _always_ benefits from `repaint(data,IMMED)` and `repaint(data,DEFERRED)`.

That's just a wrong analogy :P you wouldn't name it `TEN`, you would name it whatever the number was supposed to mean. It's the exact same situation.

I retract slightly, unnamed positional args _in general_ can be a real problem. `rust_analyzer` shows their names inline like this: 
image
I think if this were a widely recognised problem in JS code maintenance, editors should do this for JS. I don't like the suggested fix associated with this lint rule. Passing literals as positional args shouldn't be considered a problem.

lawrence-dol:
> That's just a wrong analogy :P you wouldn't name it TEN, you would name it whatever the number was supposed to mean. It's the exact same situation.

It's not an _analogy_, it's an _example_, specifically on in which a literal integer value would be used because there's no logical abstract name for the value.

> I retract slightly, unnamed positional args in general can be a real problem.

Yes they can. But booleans are unique int that, again, they _always_ benefit from being named; as the above example shows, that's not true of other literals. Sure it _might_ be true of an integer being used as an enumeration, but one can't practically say all integers shall have names.

But I don't get the push-back. It's an optional rule that will have real, practical benefit for our team. If you don't like it, don't use it.

CRBl69:
> Perhaps this rule would be more useful if it were no-boolean-literal-params where it checks function, constructor, and method declarations for use of a boolean parameter instead of the call expressions?

> I agree with @dsherret, this rule should only consider declarations, not invocations of functions.

I'm not sure I understand this correctly, you want the rule to forbid the following ?
```typescipt
constructor(isVisible: boolean /* forbid this boolean */) {
  // ...
}
```

lawrence-dol:
No, I want it to forbid `constructor(true);`. I can't control how the function is defined, but I can control how it is used.

CRBl69:
I understood @lawrence-dol opinion (which I personally agree with), I want to understand others' too

Cre3per:
I'm neither a deno user nor a js/ts expert. I find this discussion interesting nonetheless, because it applies to all languages without keyword arguments.

IDE inlays for positional arguments, as shown by @nayeemrmn, solve the issue this lint rule is addressing, if the reader has an IDE and it the IDE understands the code.  
If a developer is working with broken code, and the IDE fails to find a function declaration matching a call, the developer will be at a loss. It's that time when the developer would most benefit from readable code.  
Personally I think inlays for arguments have great potential for worsening code readability.

There are valid uses of unnamed boolean arguments, and it's too hard for a linter to tell them apart from invalid uses

```
// valid
window.set_bordered(true);
// invalid
window.repaint(true);
```

More often than not, an unnamed boolean argument makes code harder to understand. Uses should be flagged by a linter, but without causing false positives.

[foxglove, an eslint plugin](https://github.com/foxglove/eslint-plugin/tree/main/rules#rules), has this same rule, but with an important option: 

> `allowLoneParameter: true` will not report an error if a boolean parameter is the only parameter to a function

I've never used this rule, I suppose it acts like this

```
window.set_bordered(true); // valid
window.repaint(true); // valid
window.set_border(color::red, true); // invalid
window.repaint(true, true); // invalid
```

I can't think of an example where a function with multiple parameters, one of them being a `bool`, makes sense. This is easy to detect for a linter.

Here's a good read about [boolean traps](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)

CRBl69:
Hi, I've read the blog from @Cre3per's comment and I must say that IMO, it justifies this rule as an opt-in rule. Although as @dsherret said, it should also warn on function declaration (and constructors), and suggest using an `object` instead of a `boolean`.

This :

```typescript
class Widget {
  function repaint(immediate: boolean) {
    // ...
  }
}
```

Should be replaced by this :

```typescript
class Widget {
  function repaint({immediate: boolean}) {
    // ...
  }
}
```

So in the end, the rule will warn users if they use boolean literals in function and constructor calls, and warn the use of booleans in function and constructor declaration.

What do you think ? Should I start making these modifications ?

lawrence-dol:
@CRBl69 : Incidentally, your idea could be implemented as `no-boolean-args` and address the same problem at the declaration point. Personally, I would likely leave this off, while turning `no-boolean-literal-args` on, but who knows? If JavaScript had a good syntax for defining and using enumeration types I'd likely opt to use them instead of boolean values.

lawrence-dol:
To be clear, what I am saying is that the two rules call-point and declaration-point should be separate, as they address separate developers. The one I requested in this feature has been implemented as I've suggested. An _additional_ rule could be also added to address the need at the declaration-point.

bartlomieju:
**Verdict: needs human decision** (new opinionated rule + stale branch)

Automated review focused on correctness/perf. No blocking *implementation* bugs, but some design and coverage points worth a maintainer's call before merge.

**Design / product call (not a code bug)**
- The rule flags *every* boolean literal passed to *any* call/`new` expression. This is very aggressive — idiomatic calls like `setEnabled(true)`, `el.toggle(false)`, `assertEquals(x, true)` would all warn, with no way to express "this arg is genuinely boolean". Correctly left untagged (not in `recommended`), which is the right default, but the high false-positive rate means this needs an explicit maintainer decision on whether it belongs in the ruleset at all.

**Correctness — minor**
- Spread args: `foo(...[true])` won't warn (the `ExprOrSpread.expr` is the array literal, not a `Bool`). That's arguably correct behavior, but it's untested — worth a valid-case test to pin it down.
- No handling of tagged-template or default-param positions, but those aren't call args so the current scope is self-consistent.

**Test coverage gaps**
- No valid test confirming boolean literals *outside* argument position (e.g. `const x = true; foo(x)`) are untouched.
- Only one `new_expr` case (invalid); add a valid `new` case.
- Add a spread-arg case (`foo(...[true])`) to lock in the behavior above.

**Perf:** no issues — single traversal, no extra allocations or cloning.

**Housekeeping:** branch shows `mergeable: UNKNOWN` with no CI results — it's stale and will need a rebase onto `main` (rule registration in `src/rules.rs` is a common conflict point) before it can land.

_No changes requested to logic; blocking item is the product decision + rebase._

PR #988, opened 2022-01-03
fix(require-await): do not skip visiting some ts nodes

### LINKED ISSUE #983: [1.17.0] Linter: False positive 'require-await' issue triggered
The following function gets incorrectly tagged with a 'require-await' issue:

```typescript
async function add(left: SnailNumber, right: SnailNumber): Promise<SnailNumber> {
    const added: SnailNumber = {
        left: <SnailNumber>await deepClone(left),
        right: <SnailNumber>await deepClone(right),
    }
    reduce(added)
    return added
}
```
Full code: https://github.com/Belgabor/aoc2021/blob/main/day18/task02.ts

ah-yu:
The rule  `require-await` skips visiting `TsTypeAssertion` node causes the bug in #983 

Closes #983

### DIFF
--- a/src/rules/require_await.rs
+++ b/src/rules/require_await.rs
@@ -163,8 +163,6 @@ fn is_body_empty(maybe_body: Option<&BlockStmt>) -> bool {
 }
 
 impl<'c, 'view> Visit for RequireAwaitVisitor<'c, 'view> {
-  // noop_visit_type!();
-
   fn visit_fn_decl(&mut self, fn_decl: &FnDecl) {
     let function_info = FunctionInfo {
       kind: FunctionKind::Function(Some(
--- a/src/rules/require_await.rs
+++ b/src/rules/require_await.rs
@@ -7,7 +7,7 @@ use deno_ast::swc::ast::{
   FnExpr, ForOfStmt, MethodProp, PrivateMethod,
 };
 use deno_ast::swc::common::Spanned;
-use deno_ast::swc::visit::{noop_visit_type, Visit, VisitWith};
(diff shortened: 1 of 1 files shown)


src/rules/require_await.rs
```diff
@@ -163,8 +163,6 @@ fn is_body_empty(maybe_body: Option<&BlockStmt>) -> bool {
 }
 
 impl<'c, 'view> Visit for RequireAwaitVisitor<'c, 'view> {
-  noop_visit_type!();
```
dsherret, line 166: I would prefer swc be fixed to visit TS nodes in this case rather than us removing this.

dsherret: I opened https://github.com/swc-project/swc/pull/3192 but it's unclear whether it should be merged there. Perhaps we can create our own `noop_visit_type` macro that skips over types.

bartlomieju: @dsherret actually this rule is still using "old" visitor API instead of `deno_ast::view`. I think this problem could be fixed by rewriting the rule.

ah-yu:
Closing this PR because switching to `deno_ast::view` is a better way to solve #983 .

bartlomieju:
@ah-yu do you want to try and tackle a rewrite to using `deno_ast::view`? Let me know if you need any pointers in that direction.

ah-yu:
@bartlomieju yes! I would like to give it a try. thanks!

bartlomieju:
Great, looking forward to your PR @ah-yu

PR #990, opened 2022-01-05
refactor(require-await): use `Handler` instead of `Visitor`

### LINKED ISSUE #983: [1.17.0] Linter: False positive 'require-await' issue triggered

ah-yu:
Related #818 

Close #983

### DIFF
--- a/src/rules/require_await.rs
+++ b/src/rules/require_await.rs
@@ -1,13 +1,11 @@
 // Copyright 2020-2021 the Deno authors. All rights reserved. MIT license.
 use super::{Context, LintRule};
+use crate::handler::{Handler, Traverse};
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, thank you @ah-yu

PR #991, opened 2022-01-05
fix(ban-types): ban types should not flag non-global types

dsherret:
Basically if you do:

```ts
import { Symbol } from "./something";

const t: Symbol;
```

...it should not be a lint diagnostic.


src/scopes.rs
```diff
@@ -340,9 +340,11 @@ impl Visit for Analyzer<'_> {
   }
 
   fn visit_ts_type_alias_decl(&mut self, n: &TsTypeAliasDecl) {
-    self.declare(BindingKind::TypeAlias, &n.id);
-    n.type_params.visit_with(self);
-    n.type_ann.visit_with(self);
```
dsherret, line 345: I don't think this needs to visit the type parameters and type?

bartlomieju (approved):

Issue #968, opened 2021-11-02
prefer-primordials: Force `SafeArrayIterator` to be used in deno core


petamoriken:
related: https://github.com/denoland/deno/issues/12614

### Lint Name

prefer-primordials

### Code Snippet

```ts
[1, 2, ...arr];
foo(1, 2, ...arr);
new Bar(1, 2, ...arr);
```

### Expected Result

Suggest using `SafeArrayIterator`.

```ts
const { SafeArrayIterator } = window.__bootstrap.primordials;
[1, 2, ...new SafeArrayIterator(arr)];
foo(1, 2, ...new SafeArrayIterator(arr));
new Bar(1, 2, ...new SafeArrayIterator(arr));
```

### Actual Result

Nothing.

### Additional Info

I think the following query (esquery) will help detect the problem points.

```css
:matches(ArrayExpression, CallExpression, NewExpression)
  > SpreadElement:not(
      [argument.type = NewExpression][argument.callee.name = SafeArrayIterator]
  )
```

### Version

deno 1.15.3 (release, aarch64-apple-darwin)
v8 9.5.172.19
typescript 4.4.2

petamoriken:
The `for-of` loops also need to be handled.

```js
for (const val of new SafeArrayIterator(arr)) {}
```

PR #995, opened 2022-01-12
fix(prefer-primordials): Force `SafeArrayIterator` to be used for spread syntax in deno core

petamoriken:
Related https://github.com/denoland/deno_lint/issues/968

### DIFF
--- a/src/rules/prefer_primordials.rs
+++ b/src/rules/prefer_primordials.rs
@@ -139,7 +139,7 @@ impl Handler for PreferPrimordialsHandler {
   fn expr_or_spread(
     &mut self,
     expr_or_spread: &ast_view::ExprOrSpread,
-    ctx: &mut Context
+    ctx: &mut Context,
   ) {
     if_chain! {
       if expr_or_spread.inner.spread.is_some();


petamoriken:
`spread_or_expr` handler isn't called. Does anyone who helps me exist?

petamoriken:
`spread_or_expr` handler isn't called. Does anyone who helps me exist?

petamoriken:
`spread_or_expr` handler isn't called. Does anyone who helps me exist?

bartlomieju:
> `spread_or_expr` handler isn't called. Does anyone who helps me exist?

In this rule this handler is not overridden in any way so it just defaults to its base behavior. What is the expected bahavior in this case? If you add this handler manually can you see it being called?

petamoriken:
> What is the expected bahavior in this case?

I want to detect Spread Syntax in Array Literal, Function Apply and Function Construct.

> If you add this handler manually can you see it being called?

It seems that the handler is not being called.

ia7ck (commented):
@petamoriken 
To add a lint diagnostic when the parser visits an `ExprOrSpread` node, override the [`expr_or_spread`](https://github.com/denoland/deno_lint/blob/f963e0917835f20035818e032ba853d8a69cff10/src/handler.rs#L84) handler. 👍 

```rs
impl Handler for PreferPrimordialsHandler {
  
  // ...
  
  fn expr_or_spread(&mut self, expr_or_spread: &ast_view::ExprOrSpread, ctx: &mut Context) {
    if expr_or_spread.inner.spread.is_some() {
      ctx.add_diagnostic_with_hint(expr_or_spread.span(), CODE, MESSAGE, HINT);
    }
  }
}
```

bartlomieju:
This looks great, I think it's useful. @lucacasonato any objections?

petamoriken:
For `for-of`, I will create another PR.

Issue #1004, opened 2022-01-29
Typo when getting an error for declaring a module inside a module


hironichu:
### Lint Name

prefer-namespace-keyword

### Code Snippet

```ts
declare module "test" {
	module test {
		export var x: number;
	}
}
```

### Expected Result

`module` keyword in module decl**a**ration is not allowed

### Actual Result

`module` keyword in module decl**e**ration is not allowed

### Version

deno 1.18.1 (release, x86_64-pc-windows-msvc)
v8 9.8.177.6
typescript 4.5.2

PR #1005, opened 2022-01-29
fix: typo in "prefer-namespace-keyword"

DmitryRomaniuk:
fix https://github.com/denoland/deno_lint/issues/1004


bartlomieju (approved):
Thanks

PR #1007, opened 2022-02-04
handle the 'in' operator in prefer-primordials

### LINKED ISSUE #993: prefer-primordials: Don't use `in` operator, use `ObjectHasOwn`
related: https://github.com/denoland/deno_lint/issues/986#issuecomment-1003818902

### Lint Name

prefer-primordials

### Code Snippet

```ts
"a" in b;
```

### Expected Result

Suggest using `ObjectHasOwn`.

```ts
ObjectHasOwn(b, "a");
```

### Actual Result

Nothing.

### Version

deno 1.17.1 (release, aarch64-apple-darwin)
v8 9.7.106.15
typescript 4.5.2

malectro:
closes #993

i might be missing something, but the message and hint seem pretty general here. should they be more specific, or is this fine?

### DIFF
--- a/src/rules/prefer_primordials.rs
+++ b/src/rules/prefer_primordials.rs
@@ -191,8 +191,10 @@ impl Handler for PreferPrimordialsHandler {
   }
 
   fn bin_expr(&mut self, bin_expr: &ast_view::BinExpr, ctx: &mut Context) {
-    use deno_ast::view::BinaryOp;
-    if matches!(bin_expr.op(), BinaryOp::InstanceOf) {
+    use ast_view::BinaryOp;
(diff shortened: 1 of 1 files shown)


bartlomieju:
> i might be missing something, but the message and hint seem pretty general here. should they be more specific, or is this fine?

This is fine, this rule is only used for internal Deno code, so the message doesn't need to be very specific.

bartlomieju:
> i might be missing something, but the message and hint seem pretty general here. should they be more specific, or is this fine?

This is fine, this rule is only used for internal Deno code, so the message doesn't need to be very specific.

bartlomieju (approved):
LGTM, thank you @malectro

PR #1015, opened 2022-02-17
fix(no-deprecated-deno-api): handle type annotations

### LINKED ISSUE #1014: no-deprecated-deno-api doesn't catch deprecations in types
### Lint Name

no-deprecated-deno-api

### Code Snippet

```ts
    const file: Deno.File = await Deno.open(tempFile, {
      create: true,
      write: true,
      read: true,
    });
```

### Expected Result
Should warn about `Deno.File` being deprecated

### Actual Result

No warning

### Version

Deno 1.19.0

bartlomieju:
Closes #1014

### DIFF
--- a/src/rules/no_deprecated_deno_api.rs
+++ b/src/rules/no_deprecated_deno_api.rs
@@ -202,6 +202,28 @@ impl Handler for NoDeprecatedDenoApiHandler {
       }
     }
   }
+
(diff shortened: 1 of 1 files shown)


dsherret (approved):
LGTM. I fixed it to not error when not global though.

Issue #944, opened 2021-10-20
Rule suggestion: no-external-imports (enforce deps.ts convention)


gewoonwoutje:
I'd like to use a `deps.ts` and/or import map to load external imports (i.e. non-relative imports), and would like to enforce this convention. In other words, I'd like to block using imports from `https://...` sources in all files, except import maps and a `deps.ts` dependencies file (/ files excluded with an ignore pragma).

PR #1022, opened 2022-02-27
Add no-external-imports rule

hosseind88:
#944


bartlomieju:
@hosseind88 @magurotuna I'm a bit confused by this rule. If I get this right, it disallows remote import in any file but `deps.ts`. What is the use case for that?

bartlomieju:
Okay, sounds good. I agree that it shouldn't be in the "recommended" set

hosseind88:
Hi guys, can we merge this? I see one failure in tests on ubuntu but I don't know why that is happening

hosseind88:
aa right sorry forgot that, pushed now

bartlomieju (approved):
LGTM, thank you @hosseind88

Issue #1024, opened 2022-03-15
Try out "miette" crate in "dlint" binary


bartlomieju:
Currently we use `annotate-snippets` crate to display diagnostics in `dlint` and `deno lint` has its own diagnostics display API.

We should try out https://docs.rs/miette/4.2.1/miette/ which seems to be well suited for this task and is now used by SWC:
https://github.com/swc-project/swc/pull/3946

kdy1:
Note that `Send` is required for [`SourceCode`](https://docs.rs/miette/4.2.1/miette/trait.SourceCode.html)

cc @zkat

I enabled `concurrent` feature of `swc_common` because of this bound, but deno does not use this flag and this flag is global, so I think there should be a way to use non-sync type as `SourceCode`.

Is it required because of the global hook?

dsherret:
@kdy1 we share source text and ASTs around in an `Arc` in our LSP. I'm guessing we would just implement this trait for our `ParsedSource` struct: https://github.com/denoland/deno_ast/blob/31ef71c157d93db40dac518dfb25814b122395ca/src/parsed_source.rs#L34

dsherret:
One thing is that we hold onto diagnostics for a long time and only surface them for certain actions in the language server. For example, we allow doing linting on files with diagnostics, but we don't allow formatting to happen if there's certain diagnostics so we check them then.

PR #1027, opened 2022-03-16
Use miette in dlint for diagnostic output

bartlomieju:
Closes https://github.com/denoland/deno_lint/issues/1024

Screenshot_2022-03-16_at_23 52 57

Co-Authored-By: 강동윤 <kdy1997.dev@gmail.com>


bartlomieju:
@kdy1 you did a heavy lifting in SWC to integrate miette and I reused some of the code. I'll co-author you if this PR lands.

Cargo.lock
```diff
@@ -133,6 +153,12 @@ version = "3.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8f1e260c3a9040a7c19a12468758f4c16f31a81a1fe087482be9570ec864bb6c"
 
+[[package]]
+name = "cc"
```
ry, line 157: hmm..

kdy1:
@bartlomieju Thank you!

Cargo.lock
kitsonk: hmm indeed... native C code is a necessary dependency? 😕

kitsonk: Look like it comes from [backtrace](https://crates.io/crates/backtrace), surprising given that backtrace is an "official" crate...

kitsonk: Looks like it is only used when targetting android, there is a single .c binding file.  There is this note in the [Cargo.toml](https://github.com/rust-lang/backtrace-rs/blob/master/Cargo.toml#L53-L56):

```toml
[build-dependencies]
# Only needed for Android, but cannot be target dependent
# https://github.com/rust-lang/cargo/issues/4932
cc = "1.0.67"
```

bartlomieju: Opened https://github.com/zkat/miette/issues/132

bartlomieju:
This is now ready, just pending new release of `miette`

dsherret (commented):
Almost there. It seems there is one bug.

examples/dlint/diagnostics.rs
```diff
@@ -1,200 +1,169 @@
...
+    let src_end = self
+      .source
+      .line_end(end_line_column.line_index + context_lines_after)
+      .0 as usize;
+    let src_end = std::cmp::min(src_end, line_count);
+    let src = &self.source.text_str()[src_start..src_end];
+    let name = Some(self.filename.to_string());
+    let start = miette::SourceOffset::from(src_start);
+    let len = miette::SourceOffset::from(src_end - src_start);
+    let span = miette::SourceSpan::new(start, len);
```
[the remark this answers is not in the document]
bartlomieju: Haven't found any, but I also didn't look very hard - this is mostly copy-pasted from SWC source code, so I trusted that kdy1 found the shortest way to do it.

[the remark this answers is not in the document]
kdy1: We can extract it as a method, but I didn't bother to just because there it was not required frequently.

bartlomieju:
Looks like the tests fail because the output on CI is different than the output I get locally (additional info like links are printed). I need to tweak some more settings in `miette` to enforce consistent behavior.

bartlomieju:
@dsherret please take a look

dsherret (approved):
LGTM

Issue #1013, opened 2022-02-17
Ability to automatically fix linting errors


dsherret:
It might be useful for lints to include [text changes](https://github.com/denoland/deno_ast/pull/63) in the lint diagnostic that would fix the diagnostic. For example, `no-unused-var` could return text changes that would remove unused imports.

This could then be used by a `deno lint --fix` command.

kitsonk:
We should consider how we can leverage this properly in the lsp.  I believe currently the only quick fixes are to insert comments to ignore the lint error, would be good to support more useful/intelligent fixes.

dsherret:
@kitsonk oh yeah didn't think about that. That would definitely be desired. I guess it should probably have the ability to return multiple possible solutions as a fix and they should be given a description. It could use the first one for `deno lint --fix` and then present all the possible options to someone in the editor.

kitsonk:
Good idea.  Using roughly the pattern of the [code action](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction) from the LSP protocol would make the most sense.  While it doesn't have to be exactly 1-1, there should be a method on whatever structure to be able to convert it to an LSP code action.

This would also be a good way of structuring them for something like `deno lint --check` which would give you a preview of what changes would be done with `deno lint --fix`?

PR #1034, opened 2022-03-31
feat: add no-unneeded-ternary rule

### LINKED ISSUE #1032: Rule suggestion: Remove Unnecessary Conditional Expression
Turns:
```ts
const test = "Hey";
console.log(test === "Hello World!" ? true : false);
```
Into:
```ts
const test = "Hey";
console.log(test === "Hello World!");
```
Also
```ts
const test = "Hey";
console.log(test === "Hello World!" ? false : true);
// Turns into
console.log(test !== "Hello World!");
```

ah-yu:
Closes #1032

### DIFF
--- a/docs/rules/no_unneeded_ternary.md
+++ b/docs/rules/no_unneeded_ternary.md
@@ -1,0 +1,17 @@
+Disallows ternary operators when simpler alternatives exist.
+
+It's a common mistake to use a conditional expression to select between two
+Boolean values instead of using ! to convert the test to a Boolean.
+
+### Invalid:
+
+```typescript
+const foo = condition ? true : false;
+```
+
+### Valid:
+
+```typescript
+const foo = condition ? x : y;
+const foo = condition ? x : false;
+```
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -79,6 +79,7 @@ pub mod no_this_before_super;
 pub mod no_throw_literal;
 pub mod no_top_level_await;
 pub mod no_undef;
+pub mod no_unneeded_ternary;
 pub mod no_unreachable;
 pub mod no_unsafe_finally;
 pub mod no_unsafe_negation;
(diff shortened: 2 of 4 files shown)


lucsoft:
im new to deno_lint but does this support the "quick fix" options from the LSP

ah-yu:
@lucsoft Nope. See https://github.com/denoland/deno_lint/issues/1013

docs/rules/no_unneeded_ternary.md
```diff
@@ -0,0 +1,17 @@
...
+```typescript
+const foo = condition ? true : false;
+```
+
+### Valid:
+
+```typescript
+const foo = condition ? x : y;
+const foo = condition ? x : false;
+```
```
bartlomieju, line 17: Could you add more examples? I still don't understand the purpose of this rule just from examples

bartlomieju (changes requested):

docs/rules/no_unneeded_ternary.md
ah-yu: Sure. I think I need to add some comments to make it more clear.

bartlomieju:
**Verdict: LGTM on the code — needs human decision** to accept a new rule + a rebase.

The implementation is clean and correct; no correctness or performance bugs found.

**Correctness — good**
- `cond_expr.cons.is::<Bool>() && cond_expr.alt.is::<Bool>()` faithfully mirrors ESLint's `no-unneeded-ternary` boolean-literal branch (report when *both* consequent and alternate are boolean literals). This also (correctly, per ESLint) flags `x ? true : true` / `x ? false : false`, not just the inverted pair — good.
- Valid/invalid cases match the documented semantics.

**Scope limitation (worth a maintainer note, not a bug)**
- ESLint's rule also has the `defaultAssignment` behavior (on by default) that flags `x ? x : y` → suggest `x || y`. This PR implements only the boolean-literal half. That's a reasonable MVP, but the docs link to the full ESLint rule, so either note the narrower scope in the docs or track the rest as a follow-up.

**Test coverage gaps**
- Add the same-value cases `x ? true : true` and `x ? false : false` — they're a distinguishing behavior of this rule and currently untested.
- No nested/`CondExpr`-in-branch case.

**Perf:** none — single `cond_expr` visitor, no allocations.

**Housekeeping**
- `mergeable: UNKNOWN`, no CI results → branch is stale; needs a rebase (`src/rules.rs` registration is a common conflict).
- Edits `www/public/docs.json`, which is a generated artifact — likely to conflict on rebase and may be regenerated rather than hand-edited; confirm the current source-of-truth for docs before landing.

_No logic changes requested; blockers are the product decision + rebase._

PR #1046, opened 2022-05-17
refactor: upgrade to deno ast 0.15

dsherret:
Upgrades to deno_ast 0.15


examples/dlint/color.rs
```diff
@@ -1,9 +1,11 @@
 // Copyright 2020-2021 the Deno authors. All rights reserved. MIT license.
-use crate::lexer::{lex, MediaType, TokenOrComment};
-use deno_ast::swc::parser::token::{Token, Word};
+use deno_ast::{
+  lex,
+  swc::parser::token::{Token, Word},
+  MediaType, TokenOrComment,
+};
```
dsherret, line 6: I'm not going to fix these using statements in this PR because it's too much work. We should do a format on the files with import granularity = Item once to fix all these.

PR #1047, opened 2022-05-20
fix(control-flow): analyze try statements in correct order

### LINKED ISSUE #1031: Problem in `finally` block
Not a big issue but I don't think `deno lint` should throw a `no-unreachable` problem here. I could prevent the problem by moving the `return` statement outside the block.

### Lint Name
`no-unreachable`

### Code Snippet

```ts
private async handleContext(ctx: Context<S>): Promise<Response> {
    try {
      ctx = await this.invokeHandlers(ctx, this.routes);
    } catch (error) {
      ctx.error = error;
      ctx = await this.invokeHandlers(ctx, this.catchRoutes);
    } finally {
      ctx = await this.invokeHandlers(ctx, this.finallyRoutes);
      return ctx.response;
    }
  }
```

### Expected Result

Nothing being thrown

### Version

deno 1.20.1 (release, x86_64-unknown-linux-gnu)
v8 10.0.139.6
typescript 4.6.2

ah-yu:
Closes #1031

### DIFF
--- a/src/control_flow/analyze_test.rs
+++ b/src/control_flow/analyze_test.rs
@@ -503,7 +503,7 @@ function foo() {
     false,
     Some(End::Forced {
       ret: true,
-      throw: false, // wont throw
+      throw: false,
       infinite_loop: false
     })
   ); // TryStmt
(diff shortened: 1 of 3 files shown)


src/control_flow/analyze_test.rs
```diff
@@ -585,7 +585,7 @@ function foo() {
     false,
     Some(End::Forced {
       ret: true,
-      throw: true,
+      throw: false,
```
ah-yu, line 588: The try statement in this test case won't throw.

src/control_flow/analyze_test.rs
```diff
@@ -503,7 +503,7 @@ function foo() {
...
```
ah-yu, line 506: The try statement in this test case won't throw.

bartlomieju (approved):
LGTM! Nice fix, thank you @ah-yu

PR #1058, opened 2022-07-07
feat: add no-implicit-declare-namespace-export (#1057)

### LINKED ISSUE #1057: Lint for Implicitly Exported Types in Declare Namespace
Hello, as https://github.com/denoland/deno/issues/14976 has shown, it can be hard to realize TypeScript's default behavior of exporting all types in a `declare namespace` block.  It might be beneficial to add an optional lint (something like `no_implicit_declare_namespace_export`) onto `type` and `interface` members that are implicitly exported.  This should be able to help stop further costly bugs from happening in Deno and elsewhere.

Example code:

```ts
declare namespace ResourceHandler {
  type FormatResource<T extends string> = `resource-${T}`;
}
```

Output:

```
Type 'FormatResource' is implicitly exported from the 'ResourceHandler' namespace. Try adding an 'export {};' to the top of the namespace to disable this behavior.
```

And the suggestion could be integrated into editors.  Thanks

Conaclos:
Fixes #1057

### DIFF
--- a/src/rules/no_implicit_declare_namespace_export.rs
+++ b/src/rules/no_implicit_declare_namespace_export.rs
@@ -1,0 +1,196 @@
+// Copyright 2020-2021 the Deno authors. All rights reserved. MIT license.
+use super::{Context, LintRule};
+use crate::handler::{Handler, Traverse};
+use crate::{Program, ProgramRef};
+use deno_ast::{view as ast_view, SourceRanged};
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub struct NoImplicitDeclareNamespaceExport;
+
+const CODE: &str = "no-implicit-declare-namespace-export";
+const MESSAGE: &str =
+  "Implicit exports in ambient namespaces are discouraged to
+use";
+const HINT: &str = "Try adding an `export {};` to the top of the namespace to disable this behavior";
+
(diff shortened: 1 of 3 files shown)


src/rules/no_implicit_declare_namespace_export.rs
```diff
@@ -0,0 +1,199 @@
...
+  fn lint_program(&self, _context: &mut Context, _program: ProgramRef<'_>) {
+    unreachable!();
+  }
+
+  fn lint_program_with_ast_view(
+    &self,
+    context: &mut Context,
+    program: Program<'_>,
+  ) {
+    if context.file_name().ends_with(".d.ts") {
```
ah-yu, line 39: This is different from what is written in the documentation. The document said this lint rule works for both `ts` and `d.ts` files.

Conaclos: Thanks! It is now fixed.

Conaclos:
What is the next step for approval?

bartlomieju:
Hey @Conaclos sorry for slow review, I missed previous notification.

Implementation wise it looks great to me, given that it's not on by default I think we can land it as is.

Conaclos:
@bartlomieju No problem :) Great to hear it :)

PR #1063, opened 2022-07-31
Include  `eval` in prefer-primordials lint

### LINKED ISSUE #1062: prefer-primordials: Introduce `eval`
`eval` is introduced as a primordial here: https://github.com/denoland/deno/pull/15110.
And I've seen it used in other places, such as [an FFI PR](https://github.com/denoland/deno/pull/15313).

Phosra:
Warns authors of Deno's internal runtime JavaScript to avoid using the global `eval` function.
Fixes #1062.

### DIFF
--- a/src/rules/prefer_primordials.rs
+++ b/src/rules/prefer_primordials.rs
@@ -555,7 +555,7 @@ const noop = Function.prototype;
       r#"eval("console.log('This test should fail!');");"#: [
         {
           col: 0,
-          message: MASSAGE,
+          message: MESSAGE,
           hint: HINT,
         },
       ],
--- a/src/rules/prefer_primordials.rs
+++ b/src/rules/prefer_primordials.rs
@@ -323,6 +323,10 @@ const { SafeArrayIterator } = primordials;
 for (const val of new SafeArrayIterator(arr)) {}
 for (const val of new SafeArrayIterator([1, 2, 3])) {}
       "#,
+      r#"
+const { indirectEval } = primordials;
+indirectEval("console.log('This test should pass.');");
+      "#,
     };
   }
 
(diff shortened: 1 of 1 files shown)


Phosra:
cc @bartlomieju.

magurotuna (commented):
How about adding a test case? That would be nice as we can make sure it works as expected at the moment and also because it gets more robust for future code changes.

Phosra:
I'm sorry, but it now occurs to me that this should actually require more engineering than I naïvely expected.
This should be a failing test case:
```js
((window) => {
  const { indirectEval } = primordials;
  const eval = indirectEval;
  eval("console.log('hola')");
})(...);
```
Yes, this is a binding to the original `globalThis.eval` function.
But this should be denied because it's being called via the identifier `eval`, for the reason given here: <https://github.com/denoland/deno/commit/2d3df94d6164b945e0184043b00078fe43229f7a>.
As such, I'm unsure of how to move forward here. (...I suppose that it could be followed up upon by someone more familiar with the codebase, in another PR?)

Phosra:
@magurotuna Can you confirm that these tests are satisfactory?

magurotuna:
@Phosra So sorry for the late response. That is awesome!

As for your concern, I think this is somewhat related to a technical limitation - we are not able to rely on type information currently so we have to make a compromise to some extent. As you said, the ideal logic could be left to the future. It's good to go for now 👍

magurotuna (approved):
LGTM 🚀

Phosra:
> As for your concern, I think this is somewhat related to a technical limitation - we are not able to rely on type information currently so we have to make a compromise to some extent. As you said, the ideal logic could be left to the future. It's good to go for now 👍 

Well, for what it's worth, type information is unnecessary: it's merely that the literal syntax `eval(...)` should be prohibited. Hopefully that's useful to the implementation.

Should I make an issue about this concern, linking to these comments?

magurotuna:
@Phosra Ah I get it now. That sounds doable. Making a new issue would be nice!

bartlomieju (approved):
Thanks

PR #1064, opened 2022-08-07
fix: details summary rendering

kidonng:
Before:

image

After:

image


bartlomieju (approved):
Thanks!

kidonng:
Ah hold on a minute, I think I just need to update one place.

PR #1066, opened 2022-08-15
fix: Fix markdown css import url and the starting command in README

littletof:
- The project start command was wrong, since we need to start from root, the `www/` is needed. Maybe we should add a `deno.json`
- The ignoring-rules page didn't have proper markdown formatting, since the css import url was wrong. As a hotfix, I changed to the one used in the index page, which works

### DIFF
--- a/www/README.md
+++ b/www/README.md
@@ -5,7 +5,7 @@
 Start the project (from the root of the repo):
 
 ```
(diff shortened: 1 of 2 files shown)


hashrock (approved):
LGTM!

PR #1067, opened 2022-08-15
feat: Style valid and invalid examples

littletof:
Before / After:
image

### DIFF
--- a/www/pages/index.tsx
+++ b/www/pages/index.tsx
@@ -16,7 +16,7 @@ function IndexPage(props: PageProps) {
     useData("www/public/docs.json", diskFetcher),
   );
 
-  const stlye = useData("www/public/extension.css", diskFetcher);
+  const style = useData("www/public/extension.css", diskFetcher);
 
   const rules = jsonData.map<RuleData>((rule) => ({
     code: rule.code,
@@ -46,7 +46,7 @@ function IndexPage(props: PageProps) {
           href="https://cdn.jsdelivr.net/gh/lucacasonato/manual@df7ae27/www/static/markdown.css"
           crossOrigin="anonymous"
         />
-        <style>{stlye}</style>
+        <style>{style}</style>
       </Head>
       <Header />
       <main class={tw`my-8`}>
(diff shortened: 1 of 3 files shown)


bartlomieju:
Looks good! Thanks for looking into this @littletof. @hashrock @ije can you review?

bartlomieju:
Also @littletof how does it look in dark mode?

littletof:
@bartlomieju No problem.

Right, I haven't looked into dark mode, and it looks bad. But I think It has other problems with darkmode (at least when emulating preference). 

Will look into it.

image

bartlomieju:
Thank you, much appreciated

littletof:
@bartlomieju Dark mode could look like this:

image

I will open a separate PR for the whole darkmode fix, to get merged before this.

www/pages/index.tsx
```diff
@@ -16,6 +16,8 @@ function IndexPage(props: PageProps) {
     useData("www/public/docs.json", diskFetcher),
   );
 
+  const stlye = useData("www/public/extension.css", diskFetcher);
```
hashrock, line 19: stlye -> style

www/pages/index.tsx
```diff
@@ -44,6 +46,7 @@ function IndexPage(props: PageProps) {
           href="https://cdn.jsdelivr.net/gh/lucacasonato/manual@df7ae27/www/static/markdown.css"
           crossOrigin="anonymous"
         />
+        <style>{stlye}</style>
```
hashrock, line 49: stlye -> style

hashrock (commented):
almost looks great!

hashrock:
image

Not related to this PR, but we'd like to use classes instead of ids, since there can be more than one element on the same page. let's fix it in another PR.

www/pages/index.tsx
littletof: Right, rookie mistake. Fixed.

littletof:
@hashrock Yes, first I wanted use classes in the selector, and it would be better, but these sections are generated from markdown, and to properly change them, we would need to extend the markdown -> html generator and/or add some custom markers in the markdown descriptions. (eg. replace mardown headers with HTML header tags in the markdown itself)

As a hotfix we can string replace all `id=invalid` to `class=invalild`  before we insert the HTML string, but thats not the nicest 🙂

hashrock:
I see...thanks. I think I'll leave it as is for now.

Issue #672, opened 2021-04-20
Use `MediaType` to determine if the file is JSX or not in `no-unused-vars`


magurotuna:
> This should work for a first pass, but ultimately this should use `MediaType` to determine if it is JSX or not. (remote files can be declared JSX by using `text/jsx` content-type.)

_Originally posted by @lucacasonato in https://github.com/denoland/deno_lint/pull/664#discussion_r615969757_

bartlomieju:
Once https://github.com/denoland/deno_lint/pull/904 lands this will be possible by checking `context.media_type()` value when executing a rule.

brenelz:
I think I can tackle this issue. I am learning Rust and this seems pretty straightforward

bartlomieju:
Sounds good @brenelz

PR #1074, opened 2022-08-27
fix: use media_type instead of filename to detect jsx

brenelz:
Closes https://github.com/denoland/deno_lint/issues/672

- Cargo Test Passes
image


dsherret (approved):
LGTM. Thanks @brenelz!

Issue #1077, opened 2022-09-07
Output in compact format


hadriann:
I am using the Node.js binding (@node-rs/deno-lint) and am having a hard time integrating it with vim. It would help a lot if deno-lint had a way of controlling the output, such as the "--format compact" option in eslint. So, besides the pretty multiline colorized output that it now gives, it would be great (for integrating with other tools) to also have a plain text basic output format. For example, the compact output of eslint may be something like:

```
/path/to/file.js: line 10, col 20, Error - Unnecessary semicolon. (no-extra-semi)
/path/to/file.js: line 12, col 22, Warning - Missing semicolon. (semi)
```

bartlomieju:
Makes sense, PRs are welcome

brenelz:
Took a bit of a stab at this. If you have feedback let me know

PR #1079, opened 2022-09-11
feat: add compact format

brenelz:
closes https://github.com/denoland/deno_lint/issues/1077

Added a --format compact option that outputs like the following

image

### DIFF
--- a/examples/dlint/diagnostics.rs
+++ b/examples/dlint/diagnostics.rs
@@ -11,43 +11,46 @@ pub fn display_diagnostics(
   filename: &str,
   format: Option<&str>,
 ) {
-  for diagnostic in diagnostics {
-    match format {
-      Some("compact") => print_compact(filename, diagnostic),
-      _ => print_pretty(source_file, filename, diagnostic),
-    }
+  match format {
+    Some("compact") => print_compact(diagnostics, filename),
+    Some("pretty") => print_pretty(diagnostics, source_file, filename),
+    _ => unreachable!("Invalid output format specified"),
   }
 }
 
-fn print_compact(filename: &str, diagnostic: &LintDiagnostic) {
-  eprintln!(
-    "{}: line {}, col {}, Error - {} ({})",
-    filename,
-    diagnostic.range.start.line_index + 1,
-    diagnostic.range.start.column_index + 1,
-    diagnostic.message,
-    diagnostic.code
-  )
+fn print_compact(diagnostics: &[LintDiagnostic], filename: &str) {
+  for diagnostic in diagnostics {
+    eprintln!(
+      "{}: line {}, col {}, Error - {} ({})",
+      filename,
+      diagnostic.range.start.line_index + 1,
+      diagnostic.range.start.column_index + 1,
+      diagnostic.message,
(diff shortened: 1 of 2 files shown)


examples/dlint/main.rs
```diff
@@ -50,6 +50,11 @@ fn create_cli_app<'a>() -> Command<'a> {
             .long("config")
             .help("Load config from file")
             .takes_value(true),
+        ).arg(
+          Arg::new("FORMAT")
+            .long("format")
+            .help("Configure output format")
+            .takes_value(true),
```
bartlomieju, line 57: Maybe add a validator to check if passed value is either "compat" or "pretty"?

examples/dlint/diagnostics.rs
```diff
@@ -9,22 +9,45 @@ pub fn display_diagnostics(
   diagnostics: &[LintDiagnostic],
   source_file: &SourceTextInfo,
   filename: &str,
+  format: Option<&str>,
+) {
+  for diagnostic in diagnostics {
+    match format {
+      Some("compact") => print_compact(filename, diagnostic),
+      _ => print_pretty(source_file, filename, diagnostic),
```
bartlomieju, line 17: If you apply the validator as suggested, then we should have `Some("pretty")` arm and `_ => unreachable!()`

examples/dlint/diagnostics.rs
```diff
@@ -9,22 +9,45 @@ pub fn display_diagnostics(
   diagnostics: &[LintDiagnostic],
   source_file: &SourceTextInfo,
   filename: &str,
+  format: Option<&str>,
+) {
+  for diagnostic in diagnostics {
+    match format {
```
bartlomieju, line 15: There can't be a situation where one diagnostic is formatted using "compat" and other using "pretty" formatting. This function should be refactored to forward `diagnostics` slice to either `print_compat` or `print_pretty` helpers

bartlomieju:
Thanks for looking into this @brenelz!

bartlomieju:
@brenelz also a heads up: this change will only apply to `dlint` binary (an example binary in this repo, also used to lint JS code in `denoland/deno` repo). If we wanted to have "compat" output in `deno lint` subcommand, we will also need to add such reported to https://github.com/denoland/deno/blob/98454c1eb802b91a8c77dd97888a8994c85dfa46/cli/tools/lint.rs#L54-L59

examples/dlint/diagnostics.rs
brenelz: made some updates. Hope I understood all your comments correctly

bartlomieju (approved):
LGTM, thanks!

brenelz:
> @brenelz also a heads up: this change will only apply to `dlint` binary (an example binary in this repo, also used to lint JS code in `denoland/deno` repo). If we wanted to have "compat" output in `deno lint` subcommand, we will also need to add such reported to https://github.com/denoland/deno/blob/98454c1eb802b91a8c77dd97888a8994c85dfa46/cli/tools/lint.rs#L54-L59

Did you want me to add it here as well?

PR #1082, opened 2022-10-03
chore: upgrade Rust to 1.64

magurotuna:
With this PR the following things are done:

- Upgrade to Rust 1.64
- Apply fixes according to clippy


bartlomieju (approved):
LGTM

PR #1086, opened 2022-10-06
fix: use `media_type` instead of file_name

### LINKED ISSUE #1083: Is it appropriate to use MediaType to check file types in `src/rules` ?
Is it appropriate to use MediaType instead of filename when checking file types in `src/rules` ?

https://github.com/denoland/deno_lint/blob/55a3d5b6eb6beacd5782134aa764ea89f792136f/src/rules/no_namespace.rs#L40

If checking file type using MediaType is more appropriate, I will fix this code.

Related issues: #672

hirasawayuki:
# Fixes #1083 
Fixed to use `media_type` to check file type


magurotuna (approved):
LGTM 👍  thank you!

Issue #1073, opened 2022-08-25
getter-return: does not check Reflect.defineProperty


mfulton26:
### Lint Name

getter-return

### Code Snippet

```ts
Object.defineProperty({}, "red", { get: () => {} });
Reflect.defineProperty({}, "green", { get: () => {} });

```

### Expected Result

warnings on both getters which have no return values

### Actual Result

only a warning on `Object.defineProperty` usage but not on `Reflect.defineProperty`

### Additional Info

`Reflect.defineProperty` is like `Object.defineProperty` but returns a `boolean`.

### Version

deno 1.25.0 (release, aarch64-apple-darwin)
v8 10.6.194.5
typescript 4.7.4

mfulton26:
same goes for `Object.create`; I can open a separate issue for such if that is preferred

PR #1087, opened 2022-10-07
fix: getter_return - checking the return value of the getter property in `Reflect.defineProperty()`

hirasawayuki:
# Fixes https://github.com/denoland/deno_lint/issues/1073

- Fixed to check the return value of get property in `Reflect.defineProperty`.
- Minimal test cases added because of common implementation with `Object.defineProperty`.

I would like to fix `Object.create()` separately when I have time to spare. (https://github.com/denoland/deno_lint/issues/1073#issuecomment-1227866555)


src/rules/getter_return.rs
```diff
@@ -341,6 +341,8 @@ mod tests {
          { bar: { get: function() { return true; } } });"#,
       r#"Object.defineProperties(foo,
          { bar: { get: function () { ~function() { return true; }(); return true; } } });"#,
+      r#"Refrect.defineProperty(foo, "bar", { get: function () { return true; } });"#,
+      r#"Refrect.defineProperty(foo, "bar", { get: () => { return true; } });"#,
```
magurotuna, line 345:
```suggestion
      r#"Reflect.defineProperty(foo, "bar", { get: function () { return true; } });"#,
      r#"Reflect.defineProperty(foo, "bar", { get: () => { return true; } });"#,
```

typo

magurotuna (changes requested):
Good fix. I noticed just one typo :)

src/rules/getter_return.rs
hirasawayuki: Oh... Thank you for pointing it out 🙇‍♂️

magurotuna (approved):
LGTM 👍

PR #1091, opened 2022-10-13
fix: getter_return - permit using a generator without specifying an explicit return value

### LINKED ISSUE #1072: getter-return: should permit generators
### Lint Name

getter-return

### Code Snippet

```ts
const x = Symbol();

Object.defineProperty(Number.prototype, x, {
  *get() {
    for (let n = 0; n < this; n++) yield n;
  },
});

declare global {
  interface Number {
    [x]: IterableIterator<number>;
  }
}

Array.from(5[x], (i) => i + 1);
// => [ 1, 2, 3, 4, 5 ]

```

### Expected Result

no lint warnings

### Actual Result

```warning
Expected to return a value in 'get'.
Return a value from the getter function
```

### Additional Info

A getter within the context of `Object.defineProperty`, `Object.defineProperties`, or `Reflect.defineProperty` should be able to be defined using a generator without specifying an explicit return value.

### Version

deno 1.25.0 (release, aarch64-apple-darwin)
v8 10.6.194.5
typescript 4.7.4

hirasawayuki:
# Fixes #1072 
getter_return rules permit using generator without explicit return value.

### DIFF
--- a/src/rules/getter_return.rs
+++ b/src/rules/getter_return.rs
@@ -204,6 +204,9 @@ impl<'c, 'view> GetterReturnVisitor<'c, 'view> {
             self.visit_getter_or_function(|a| {
               // function
               if let Expr::Fn(fn_expr) = &*kv_prop.value {
+                if fn_expr.function.is_generator {
+                  return;
+                }
                 a.set_getter_name(&fn_expr.ident);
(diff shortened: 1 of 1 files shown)


magurotuna (approved):
LGTM 👍

PR #1104, opened 2022-12-23
ref: Exclude rules takes precedence over include

kamilogorek:
More details and alternative solution in https://github.com/denoland/deno/issues/17161#issuecomment-1364092487

Also reworded some tests and updated assertions.

Fixes https://github.com/denoland/deno/issues/17161

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -197,15 +197,15 @@ pub fn get_filtered_rules(
         true
       };
 
-      if let Some(excludes) = &maybe_exclude {
-        if excludes.contains(&rule.code().to_owned()) {
-          passes &= false;
-        }
-      }
-
       if let Some(includes) = &maybe_include {
         if includes.contains(&rule.code().to_owned()) {
           passes |= true;
+        }
+      }
+
+      if let Some(excludes) = &maybe_exclude {
+        if excludes.contains(&rule.code().to_owned()) {
+          passes &= false;
         }
       }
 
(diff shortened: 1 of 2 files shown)


kamilogorek:
Oh, I didn't notice that for some reason linting/formatting failed when I switched to `deno_lint` thanks to rust toolchain playing games with me. When moved between repos, toolchain binaries got broken, and despite bing installed, they were bot detected, and reinstalling worked just fine.

Note to self/others on how to quickly fix:

> the 'rustfmt' binary, normally provided by the 'rustfmt' component, is not applicable

_(run from within repository root)_

```terminal
TC=$(rustup toolchain list | grep override | cut -d ' ' -f1); for c in {rustfmt,clippy}; do rustup component remove $c --toolchain $TC && rustup component add $c --toolchain $TC; done;
```

kamilogorek:
Well, that'll require a concrete decision then I guess.

From `examples/dlint/config.rs`
> // `include` has higher precedence over `exclude`

src/rules.rs
```diff
@@ -358,45 +358,64 @@ mod tests {
...
+      ]),
+      Some(vec![
+        "ban-untagged-todo".to_string(),
+        "ban-ts-comment".to_string(),
+        "no-const-assign".to_string(),
+        "no-throw-literal".to_string(),
+      ]),
     );
-    assert_eq!(rules.len(), get_recommended_rules().len());
+    assert_eq!(rules.len(), 2);
```
magurotuna, line 418: Nice comment! Though I would like to suggest adding another assertion to verify that the returned `rules` consists of `no-const-assign` and `no-throw-literal`, because that would cleary show how deno_lint treats precedence even if it's obvious.

magurotuna (commented):
Thanks for working on this issue!

Regarding the following one:

> // include has higher precedence over exclude

if I remember correctly, the precedence of "include > exclude" was arbitrarily chosen so we could change it if there's a good reason (and I believe "exclude > include" precedence would be okay given that including the same rule in the exclude and include is most likely a mistake).
One thing to keep in mind is that the current precedence is explicitly documented [here](https://deno.land/manual@v1.29.1/tools/linter). We should remember to update it.

magurotuna (approved):
LGTM 👍 Thank you!

PR #1105, opened 2022-12-24
feat: Add guard-for-in rule

### LINKED ISSUE #1103: Feature: guard-for-in
In the for-in loop, it is necessary to check whether the property is owned by the object or not. How about introducing ESLint's [guard-for-in](https://eslint.org/docs/latest/rules/guard-for-in) in dlint as well?

related: https://github.com/denoland/deno/pull/17139

petamoriken:
Closes #1103

### DIFF
--- a/src/rules/guard_for_in.rs
+++ b/src/rules/guard_for_in.rs
@@ -51,24 +51,25 @@ impl Handler for GuardForInHandler {
     match for_in_stmt.body {
       Empty(_) | If(_) => (),
       Block(block_stmt) => {
-        match block_stmt.stmts.len() {
+        match block_stmt.stmts[..] {
           // empty block
-          0 => (),
+          [] => (),
 
           // block statement with only an if statement
-          1 => {
-            let If(_) = block_stmt.stmts.first().unwrap() else {
-              ctx.add_diagnostic_with_hint(for_in_stmt.range(), CODE, MESSAGE, HINT);
-              return;
-            };
+          [stmt] => {
+            if !matches!(stmt, If(_)) {
+              ctx.add_diagnostic_with_hint(
+                for_in_stmt.range(),
+                CODE,
+                MESSAGE,
+                HINT,
+              );
+            }
           }
 
           // block statement that start with an if statement with only a continue statement
-          _ => {
-            let if_stmt = if let If(if_stmt) = block_stmt.stmts.first().unwrap()
-            {
-              if_stmt
-            } else {
+          [first, ..] => {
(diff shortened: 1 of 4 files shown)


src/rules/guard_for_in.rs
```diff
@@ -0,0 +1,175 @@
...
+                ctx.add_diagnostic_with_hint(
+                  for_in_stmt.range(),
+                  CODE,
+                  MESSAGE,
+                  HINT,
+                );
+              }
+            }
+          }
+        }
```
magurotuna, line 108: Nice working code :+1: A few things that I would like to comment:

- how about using pattern matching for slices? This will allow to remove the call to `unwrap()` on `stmts.first()`.
- To me it seems a little bit unnatural to use "let else" when we won't use a captured variable. I would simply use `matches!`.

Here's an example code where these things are applied:

```rs
        match block_stmt.stmts[..] {
          // empty block
          [] => (),

          // block statement with only an if statement
          [stmt] => {
            if !matches!(stmt, If(_)) {
              ctx.add_diagnostic_with_hint(
                for_in_stmt.range(),
                CODE,
                MESSAGE,
                HINT,
              );
            }
          }

          // block statement that start with an if statement with only a continue statement
          [first, ..] => {
            let If(if_stmt) = first else {
              ctx.add_diagnostic_with_hint(
                for_in_stmt.range(),
                CODE,
                MESSAGE,
                HINT,
              );
              return;
            };

            match if_stmt.cons {
              Continue(_) => (),
              Block(inner_block_stmt) => match inner_block_stmt.stmts[..] {
                [stmt] => {
                  if !matches!(stmt, Continue(_)) {
                    ctx.add_diagnostic_with_hint(
                      for_in_stmt.range(),
                      CODE,
                      MESSAGE,
                      HINT,
                    );
                  }
                }
                _ => {
                  ctx.add_diagnostic_with_hint(
                    for_in_stmt.range(),
                    CODE,
                    MESSAGE,
                    HINT,
                  );
                }
              },
              _ => {
                ctx.add_diagnostic_with_hint(
                  for_in_stmt.range(),
                  CODE,
                  MESSAGE,
                  HINT,
                );
              }
            }
          }
        }
```

petamoriken: I addressed it with https://github.com/denoland/deno_lint/pull/1105/commits/fe39f4ef2d13572b7389760f32b4e5a60ed4fc2c 😄 
thx!

magurotuna (approved):
LGTM. Thanks!

PR #1108, opened 2023-01-01
feat(prefer-primordials): Prevent to use array pattern assignments

### LINKED ISSUE #1106: prefer-primordials: Prevent to use array destructuring assignments
Using array destructuring assignments calls `Array.prototype @@iterator` method.

```js
const original = Array.prototype[Symbol.iterator];
Array.prototype[Symbol.iterator] = function () { console.log("bang!"); return original.call(this); };

const [a, b] = [1, 2]; // bang!
```

We should use `SafeArrayIterator` or object destructuring assignments in Deno's internal code instead:

```js
const [a, b] = new SafeArrayIterator([1, 2]);
const { 0: c, 1: d } = [1, 2];
```

petamoriken:
Closes #1106

If the array pattern assignment doesn't have a rest pattern, the object pattern assginment is recommended;
otherwise, using a SafeIterator is recommended.

### DIFF
--- a/src/rules/prefer_primordials.rs
+++ b/src/rules/prefer_primordials.rs
@@ -31,7 +31,9 @@ enum PreferPrimordialsHint {
   #[display(fmt = "Instead use the equivalent from the `primordials` object")]
   GlobalIntrinsic,
   #[display(fmt = "Wrap a SafeIterator from the `primordials` object")]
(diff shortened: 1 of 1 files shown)


magurotuna (approved):
LGTM. Good job 👍

PR #1111, opened 2023-01-06
feat(prefer-primordials): Add check for `yield*`

petamoriken:
Detects the `Symbol.iterator` method being called by `yield*`.

```js
// prototype pollution
const original = Array.prototype[Symbol.iterator];
Array.prototype[Symbol.iterator] = function () { console.log("bang!"); return original.call(this); };

function* generator() {
  yield* [1, 2, 3];
}
generator().next(); // bang!
```


magurotuna (approved):
LGTM, thanks!

PR #1114, opened 2023-01-12
feat(prefer-primordials): Add check for getter properties

### LINKED ISSUE #1112: prefer-primordials: Add check for getter properties
```js
// call ArrayBuffer.prototype.byteLength getter function
new ArrayBuffer(10).byteLength;
```

should be changed to the following:

```js
ArrayBufferPrototypeGetByteLength(new ArrayBuffer(10));
```

---

Since type information is not available, it is the only way to check if the property names are defined as getter properties in ECMAScript:

`Symbol`

* `description`

`RegExp`

* `dotAll`
* `flags`
* `global`
* `hasIndices`
* `ignoreCase`
* `multiline`
* `source`
* `sticky`
* `unicode`

`TypedArray`

* `buffer`
* `byteLength`
* `byteOffset`
* `length` 👉 better to ignore it because it cause false positives for the `Array`s

`Map`, `Set`

* `size`

`ArrayBuffer`, `SharedArrayBuffer`

* `byteLength`

`DataView`

* `buffer`
* `byteLength`
* `byteOffset`

petamoriken:
Closes #1112 

Since I reordered the functions in the second commit, it is easy to check the diffs on a per-commit base.

### DIFF
--- a/src/rules/prefer_primordials.rs
+++ b/src/rules/prefer_primordials.rs
@@ -199,57 +199,6 @@ impl Handler for PreferPrimordialsHandler {
     }
   }
 
-  fn expr_or_spread(
-    &mut self,
-    expr_or_spread: &ast_view::ExprOrSpread,
-    ctx: &mut Context,
-  ) {
-    if_chain! {
-      if expr_or_spread.inner.spread.is_some();
-      if !expr_or_spread.inner.expr.is_new();
-      then {
-        ctx.add_diagnostic_with_hint(
-          expr_or_spread.range(),
-          CODE,
-          PreferPrimordialsMessage::Iterator,
-          PreferPrimordialsHint::SafeIterator,
-        );
-      }
-    }
-  }
-
-  fn for_of_stmt(
-    &mut self,
-    for_of_stmt: &ast_view::ForOfStmt,
-    ctx: &mut Context,
-  ) {
-    if !for_of_stmt.right.is::<ast_view::NewExpr>() {
-      ctx.add_diagnostic_with_hint(
-        for_of_stmt.right.range(),
-        CODE,
-        PreferPrimordialsMessage::Iterator,
-        PreferPrimordialsHint::SafeIterator,
-      );
-    }
-  }
-
-  fn yield_expr(
-    &mut self,
-    yield_expr: &ast_view::YieldExpr,
-    ctx: &mut Context,
-  ) {
-    if yield_expr.delegate()
-      && !matches!(yield_expr.arg, Some(ast_view::Expr::New(_)))
-    {
-      ctx.add_diagnostic_with_hint(
-        yield_expr.range(),
-        CODE,
-        PreferPrimordialsMessage::Iterator,
-        PreferPrimordialsHint::SafeIterator,
-      );
-    }
-  }
-
   fn member_expr(
     &mut self,
     member_expr: &ast_view::MemberExpr,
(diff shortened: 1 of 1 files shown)


petamoriken:
@magurotuna please review

src/rules/prefer_primordials.rs
```diff
@@ -261,15 +229,29 @@ impl Handler for PreferPrimordialsHandler {
...
+      if let Expr::Ident(ident) = &member_expr.obj;
+      if GLOBAL_TARGETS.contains(&ident.sym().as_ref());
+      then {
+        ctx.add_diagnostic_with_hint(
+          member_expr.range(),
+          CODE,
+          PreferPrimordialsMessage::GlobalIntrinsic,
+          PreferPrimordialsHint::GlobalIntrinsic,
+        );
+        return;
```
petamoriken, line 245: Should I remove this `return`?

magurotuna (commented):
Maybe I'm missing something, but code like `ourOwnInstance.buffer` will be warned with this new feature, right? Is it intentional?
Although there's no type information available, just checking potentially dangerous getters by matching its name looks a bit too restrictive.

petamoriken:
> Maybe I'm missing something, but code like ourOwnInstance.buffer will be warned with this new feature, right? Is it intentional?

Yes.

> Although there's no type information available, just checking potentially dangerous getters by matching its name looks a bit too restrictive.

This way, it is hard to avoid false positives, but since most of the time `.buffer` is accessed by `TypedArray` or `DataView`, it is unlikely to cause any problems. If false positives are detected, we can still use `// deno-lint-ignore prefer-primordials` to work around them.

magurotuna:
Thanks for clarification. I get the point and am inclined towards it, but still not 100% sure if it's worth the effort and cost. (Also a bit worried if this might have adverse impacts on performance)

@bartlomieju Any thoughts? Do you think that protecting getters by matching its name us is worthwhile?

bartlomieju:
> Thanks for clarification. I get the point and am inclined towards it, but still not 100% sure if it's worth the effort and cost. (Also a bit worried if this might have adverse impacts on performance)
> 
> @bartlomieju Any thoughts? Do you think that protecting getters by matching its name us is worthwhile?

It's okay to me as long as the changes prompted by this lint don't cause performance regressions.

magurotuna:
OK thanks, so let's land it and see how the performance changes.

src/rules/prefer_primordials.rs
```diff
@@ -637,6 +670,17 @@ const noop = Function.prototype;
...
+const { ArrayBuffer } = primordials;
+new ArrayBuffer(10).byteLength;
+      "#: [
+        {
+          line: 3,
+          col: 0,
+          message: PreferPrimordialsMessage::GlobalIntrinsic,
+          hint: PreferPrimordialsHint::GlobalIntrinsic,
+        },
+      ],
```
magurotuna, line 683: Could you also add other test cases to check all the getters included in `GETTER_TARGETS` other than `byteLength`? I think it should be helpful to make sure everything is working

petamoriken: Done!

petamoriken:
I Updated to avoid false positives for setter properties.

src/rules/prefer_primordials.rs
```diff
@@ -438,6 +474,8 @@ const parseInt = () => {};
 parseInt();
       "#,
       r#"const foo = { Error: 1 };"#,
+      r#"foo.size = 1"#,
+      r#"foo.size()"#,
```
magurotuna, line 478: 👍

magurotuna (approved):
Great, LGTM

PR #1118, opened 2023-01-20
docs: add documentation for no_sparse_arrays rule

102:


magurotuna (commented):
Thanks for contributing!
The documentation itself looks good, but for this to be rendered, could you run the following command and then commit the output?

```
cargo run --example dlint --features=docs -- rules --json > www/public/docs.json
```

102:
> Thanks for contributing! The documentation itself looks good, but for this to be rendered, could you run the following command and then commit the output?
> 
> ```
> cargo run --example dlint --features=docs -- rules --json > www/public/docs.json
> ```

Done

magurotuna (approved):
LGTM, thanks!

PR #1127, opened 2023-02-03
chore: add devcontainer

tristan-f-r:
adds a devcontainer with Deno (for www) and Rust (for the repo itself) support.

(If a contributor agrees) I do also think that the perf tools should be added as well (as mentioned in README), but for now it's a bare container.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,15 @@
+# [Choice] Debian OS version: bullseye, buster
+ARG VARIANT=bullseye
+FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/base:0-${VARIANT}
+
+ENV DENO_INSTALL=/deno
+RUN mkdir -p /deno \
+    && curl -fsSL https://deno.land/x/install/install.sh | sh \
+    && chown -R vscode /deno
+
+ENV PATH=${DENO_INSTALL}/bin:${PATH} \
+    DENO_DIR=${DENO_INSTALL}/.cache/deno
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#    && apt-get -y install --no-install-recommends <your-package-list-here>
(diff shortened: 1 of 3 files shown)


magurotuna (changes requested):
Thanks for adding the useful configuration!
Could you add `.devcontainer` directory to the format target in `tools/format.ts`, so `deno fmt` will be applied to devcontainer.json?
Like this:

```diff
const p3 = await Deno.run({
  cmd: [
    "deno",
    "fmt",
    ...checkArgs,
    "tools",
    "benchmarks/benchmarks.ts",
    "www/pages",
    "docs/rules",
    "README.md",
+   ".devcontainer",
  ],
  stdin: "null",
}).status();
```

magurotuna (changes requested):
Thanks! Please also don't remember to run the formatter so the CI passes :)

bartlomieju:
**Verdict: needs changes** (likely CI format failure) — not merging.

This is a harmless chore in principle, but there's a concrete problem introduced by the PR itself.

**Correctness / CI**
- `tools/format.ts` adds `.devcontainer` to the `deno fmt` targets, but the committed `.devcontainer/devcontainer.json` is **not fmt-clean**: it uses tab indentation, has trailing whitespace (e.g. the `"settings": {` line and a blank line with spaces), and no final newline (`\ No newline at end of file`). `deno fmt` normalizes JSON to 2-space indent + trailing newline, so `deno fmt --check` in CI will fail on this file as-is. Fix: run `deno fmt .devcontainer` and commit the result.
- Additional risk: `devcontainer.json` contains `//` comments. Confirm `deno fmt` treats a `.json` file as JSONC and doesn't error on the comments — if it doesn't, the fmt step will error rather than just reformat. (VS Code reads devcontainer.json as JSONC, so the comments are valid for the tool; the question is only whether `deno fmt` accepts them for a `.json` extension.)

**Minor / non-blocking**
- Dockerfile pins nothing for the Deno install (`deno.land/x/install/install.sh`) — fine for a dev container, but a version pin would make builds reproducible.
- `FROM --platform=linux/amd64` forces amd64; on Apple Silicon this runs under emulation. Consider dropping the platform pin unless amd64 is required.

**Perf:** n/a.

**Housekeeping:** `mergeable: UNKNOWN` / no CI results — stale branch; needs a rebase and a green CI run before it can land.

_Left for a human/another agent: apply `deno fmt` to `.devcontainer`, confirm the JSONC/fmt interaction, then it's a straightforward merge._

Issue #1126, opened 2023-02-03
[Docs] No highlighting in `ignoring-rules`.


LeoDog896:
It seems that in the `ignoring-rules` page, there is no highlighting for the code present. Maybe use [`deno-gfm`](https://github.com/denoland/deno-gfm/)?

PR #1129, opened 2023-02-07
fix: Update syntax hightlighting css file and set 100vh container

kamilogorek:
Fixes https://github.com/denoland/deno_lint/issues/1126

### Post

---

### DIFF
--- a/www/pages/ignoring-rules.tsx
+++ b/www/pages/ignoring-rules.tsx
@@ -8,14 +8,14 @@ function IgnoringRulesPage() {
   const html = renderMarkdown(md);
 
   return (
-    <div class={tw`dark:bg-[#0d1117] dark:text-white py-6`}>
+    <div class={tw`dark:bg-[#0d1117] dark:text-white py-6 h-screen`}>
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
Sweet! Thanks!

not-my-profile:
Regarding the background color adding the `h-screen` class has actually had the reverse effect in Firefox. I now get a white background in Firefox (and removing the class fixes it).

magurotuna:
@not-my-profile 
What you mean by "white background" is like this? This happens on all browsers I tried, including Chrome, Firefox, and Safari
image

not-my-profile:
Yes like that. That doesn't happen on Chromium for me.

IllusionPerdu:
I think h-screen sould be:
`.h-screen {
    height: 100%;
}`
and not 100vh because when the liste is more important than the screen height it's stopped the background after the height screen.

not-my-profile:
Ah yeah thanks for pointing this out :) Opened #1133 to fix this.

PR #1132, opened 2023-02-15
Drop `LintRule::lint_program`

not-my-profile:
If you like these changes please merge them without squashing.

Note that the last of the three commits is strictly speaking a breaking API change but I'd assume that the `LintRule` API is only used internally.

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -107,30 +107,13 @@ pub trait LintRule: std::fmt::Debug + Send + Sync {
   where
     Self: Sized;
 
-  /// Executes lint on the given `Program`.
-  /// TODO(@magurotuna): remove this after all rules get to use ast_view
-  fn lint_program<'view>(
-    &self,
-    _context: &mut Context<'view>,
-    _program: ProgramRef<'view>,
-  ) {
-    unreachable!();
-  }
-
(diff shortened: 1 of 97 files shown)


not-my-profile:
Let me know if you have questions :)

magurotuna (commented):
Thanks for the cleanup! Though I'm not really sure if I understand correctly the benefit of this change. As I commented in the source code, the reason we keep `list_program` is because we need to support both `ProgramRef` (which references to swc's structs) and `Program` (which holds deno_ast's structs) while we are migrating to the latter. So having `lint_program` serves as a marker that indicates "this implementation should be replaced using deno_ast".
If you think of any other benefits this PR will bring, I'd be happy to hear :)

not-my-profile:
The change firstly simplifies the `LintRule` trait by getting rid of an unnecessary method and secondly it gets rid of  `unreachable!();`. Idiomatic Rust code does not use `unreachable!();` since it leads to a runtime panic. Looking at the code the first time I was very much confused why there was `unreachable!()` everywhere ... it's certainly not good design.

> So having `lint_program` serves as a marker

You still have that marker ... now it's in the form of the `program_ref` function.

magurotuna:
That makes sense, thanks for clarification!

magurotuna (approved):
LGTM

PR #1133, opened 2023-02-15
fix: Fix website background color

not-my-profile:
08235828f96e42af56b8656cd6cab67825b9c594 changed the CSS classes to h-screen which Tailwind CSS defines as `heigth: 100vh;`[1], which isn't what we want ... we want `h-full` for 100% height.

This fixes https://lint.deno.land/ from looking like the following when you scroll down:

[1]: https://tailwindcss.com/docs/height


bartlomieju (approved):
Thanks!

PR #1134, opened 2023-02-15
fix(www): update fresh to latest & fix the appearance issue

magurotuna:
This commit upgrade fresh to the latest version as well as fixes a background color issue that was introduced in #1133.

Now (after #1133 lands), when we apply a filter to the rules the result looks like this:
image
However, before #1133, there was another issue where we had white background when scrolling down:

To address both of these, this commit adds `background-color: #0d1117;` to the body tag. This not only resolves the issues but also makes the background color applied beyond the bottom of the page when you try to scroll down further. (See the following gif)

#### before
before this PR

#### After

after this PR


www/static/extension.css
```diff
@@ -1,3 +1,8 @@
+body {
+  background-color: #0d1117;
+  color: #ffffff;
```
not-my-profile, line 3: I think previously the dark mode was only used when the browser preferred it via a `@media (prefers-color-scheme: dark) {` media query. This appears to change it to always use dark mode?

bartlomieju (approved):
LGTM

PR #1136, opened 2023-02-21
fix(prefer-primordials): Improve getter properties checking

petamoriken:
There are two changes in this PR

## Stop checking getter properties for `RegExp`, `Map`, and `Set`

Due to my lack of consideration, I had [a lot of false positives](https://github.com/denoland/deno/pull/17715) in getting deno_lint into deno :bow:

And this can be solved by using the primordials Safe constructors (e.g. `SafeMap`); `SafeRegExp` is not in primordials yet, but I am working on getting it in https://github.com/denoland/deno/pull/17592.

## Check right sides of Assignment Expressions

This was not considered.

### DIFF
--- a/src/rules/prefer_primordials.rs
+++ b/src/rules/prefer_primordials.rs
@@ -132,18 +132,6 @@ const GETTER_TARGETS: &[&str] = &[
   "byteOffset",
   // TypedArray: avoid false positives for Array
   // "length",
-  // RegExp: no problem when using SafeRegExp
-  // "dotAll",
-  // "flags",
-  // "global",
-  // "hasIndices",
-  // "ignoreCase",
-  // "multiline",
-  // "source",
-  // "sticky",
-  // "unicode",
-  // Map, Set: no problem when using SafeMap or SafeSet
-  // "size",
 ];
 
 struct PreferPrimordialsHandler;
--- a/src/rules/prefer_primordials.rs
+++ b/src/rules/prefer_primordials.rs
@@ -463,6 +463,18 @@ parseInt();
       r#"foo.description = 1"#,
       r#"foo.description()"#,
       r#"
+const { SafeRegExp } = primordials;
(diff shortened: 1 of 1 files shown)


src/rules/prefer_primordials.rs
```diff
@@ -128,22 +129,26 @@ const GLOBAL_TARGETS: &[&str] = &[
...
+  // "flags",
+  // "global",
+  // "hasIndices",
+  // "ignoreCase",
+  // "multiline",
+  // "source",
+  // "sticky",
+  // "unicode",
+  // Map, Set: no problem when using SafeMap or SafeSet
+  // "size",
```
magurotuna, line 151: Do we have to keep having these lines?

src/rules/prefer_primordials.rs
```diff
@@ -671,69 +678,6 @@ const noop = Function.prototype;
...
-          hint: PreferPrimordialsHint::GlobalIntrinsic,
-        },
-      ],
-      r#"/aaaa/u.flags;"#: [
-        {
-          col: 0,
-          message: PreferPrimordialsMessage::GlobalIntrinsic,
-          hint: PreferPrimordialsHint::GlobalIntrinsic,
-        },
-      ],
```
magurotuna, line 687: Why not add some test cases to "valid" test function so that we can make sure these cases don't trigger this lint error anymore?

magurotuna (commented):
Generally LGTM, but left some minor comments.

src/rules/prefer_primordials.rs
petamoriken: Sounds good. It would be too redundant to leave all of it, so I will leave some of it.

petamoriken:
@magurotuna Thank you for your review. Please check again.

magurotuna (approved):
LGTM, thanks!

Issue #1145, opened 2023-03-31
Hundreads of panics when testing 


qarmin:
When running over hundreads of broken(or even sometimes valid) Javascript/Typescript files, I got multiple crashes 

### Code Snippet
Example of broken files
```ts
{
```
```ts
var x = "1";
```
```ts
/finallyimplements  Cop/yrg
```
```ts
info: String.prototype.replace (searchValue, replaceValue)
```

Example output:
```
thread 'main' panicked at 'The specified line index 26 was greater or equal to the number of lines of 26.', /home/rafal/.cargo/registry/src/github.com-1ecc6299db9ec823/dprint-swc-ext-0.8.0/src/common/text_info.rs:204:7
```
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/rules/prefer_const.rs:696:48
```
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/rules/getter_return.rs:136:8
```

Pack of broken files(973 files, many will crash with same backtrace - most is less than 1KB is size) - [Broken_973.zip](https://github.com/denoland/deno_lint/files/11122660/Broken_973.zip)

### Version
Denolint - c801e4441492ec471038ef298dc97916f1d9d8df

magurotuna:
Thanks for reporting. I'll look into it

magurotuna:
@qarmin Could you break the error you got into separate ones? I ran deno_lint on your uploaded files but I only see "unexpected eof" or "unterminated regexp" so far, which I don't think are crashes.
Can you clarify what files trigger crashes?

qarmin:
There is too much files to create issues one by one.

First I compiled dlint with `cargo build --release` and then run it via `dlint run app.js`

 

https://user-images.githubusercontent.com/41945903/230872580-7fdf3f9f-2e53-4e78-b504-ba3a30d9b84b.mp4

magurotuna:
@qarmin Thanks, I figured out what files are causing the first crash (`dprint-swc-ext-0.8.0/src/common/text_info.rs`) and the second crash (`src/rules/prefer_const.rs`). Fixes for these crashes are in #1149 and #1148.
However, I still can't reproduce the third crash, which occurs in `src/rules/getter_return.rs`. Do you have an idea of which file can cause this crash?

qarmin:
[...]

PR #1148, opened 2023-04-11
fix(prefer-const): avoid panic when scope analysis fails

magurotuna:
This commit prevents panics from happening when scope analysis fails in `prefer-const` rule.

Towards #1145


bartlomieju (approved):
LGTM

dsherret (approved):
LGTM too.

PR #1149, opened 2023-04-11
fix(dlint): avoid panic when no trailing new line is appended

magurotuna:
This commit prevents dlint from panicking when the input file doesn't contain a final newline character.

Towards #1145


dsherret (approved):
LGTM

Issue #1165, opened 2023-06-05
Remove ./tools/format.js and ./tools/lint.js


dsherret:
We should remove this script and add a deno.json instead so people can just do `deno fmt` and `deno lint`

dsherret:
Actually these files make sense to have because they do other things like running cargo clippy and dlint. In https://github.com/denoland/deno_lint/pull/1202 I've made it so that running `deno fmt` and `deno lint` work, which should be good enough.

PR #1167, opened 2023-06-06
fix(tools): print the outputs of external commands

magurotuna:
This commit lets the outputs of external commands spawned in `tools/format.ts` printed out to stdout and stderr so that we can easily figure out what is wrong with formatting.

Without this patch, the following message is shown to us from which we can identify that there is a format issue somewhere, but it's hard to figure out where the issue is.

```
$ ./tools/format.ts --check
rustfmt
deno fmt
error: Uncaught (in promise) Error: Failed: deno fmt --check benchmarks/benchmarks.ts
  throw new Error(
        ^
    at file:///Users/yusuktan/Repo/github.com/magurotuna/deno_lint/tools/format.ts:50:9
    at eventLoopTick (ext:core/01_core.js:182:11)
```

With this patch, it will become easy to spot the problematic place:

```
$ ./tools/format.ts --check
rustfmt
deno fmt

from /Users/yusuktan/Repo/github.com/magurotuna/deno_lint/README.md:
  6 | -A Rust crate for writing fast JavaScript and TypeScript linters.This crate powers [`deno lint`](https://deno.land/manual/tools/linter), but is
  7 | -not Deno specific and can be used to write linters for Node as well.
  6 | +A Rust crate for writing fast JavaScript and TypeScript linters.This crate
  7 | +powers [`deno lint`](https://deno.land/manual/tools/linter), but is not Deno
  8 | +specific and can be used to write linters for Node as well.

error: Found 1 not formatted file in 102 files
error: Uncaught (in promise) Error: Failed: deno fmt --check
  throw new Error(
        ^
    at file:///Users/yusuktan/Repo/github.com/magurotuna/deno_lint/tools/format.ts:50:9
    at eventLoopTick (ext:core/01_core.js:182:11)
```

### DIFF
--- a/tools/format.ts
+++ b/tools/format.ts
@@ -8,25 +8,25 @@ const checkArgs = check ? ["--check"] : [];
 const p1 = new Deno.Command("rustfmt", {
   args: [...checkArgs, "examples/dlint/main.rs"],
   stdin: "null",
-});
+}).spawn();
 
-const o1 = await p1.output();
+const result1 = await p1.status;
 
-if (o1.code !== 0) {
+if (!result1.success) {
   throw new Error(
-    `Failed: rustfmt ${check ? "--check" : ""} examples/dlint/main.rs`,
+    `Failed: rustfmt ${check ? "--check" : ""}`,
   );
 }
 
 const p2 = new Deno.Command("rustfmt", {
   args: [...checkArgs, "src/lib.rs"],
   stdin: "null",
-});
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, but note that @dsherret wants both scripts removed in https://github.com/denoland/deno_lint/issues/1165

magurotuna:
OK, let's do that later.

PR #1171, opened 2023-06-26
feat: add fresh-server-event-handlers rule

marvinhagemeister:
A common friction point of new users trying out fresh is that they try to install event handlers in server-only components inside `routes/`. These components are only ever rendered on the server, so attaching event handlers does nothing. Users would get confused by that.

- https://discord.com/channels/684898665143206084/991511118524715139/1121099215490256958
- https://github.com/denoland/fresh/issues/1209

This PR adds a linting rule for when a user tries to add an event handler in a server component.


src/rules/fresh_server_event_handlers.rs
```diff
@@ -0,0 +1,202 @@
...
+    };
+
+    // Check that we're dealing with a native HTML element and not a
+    // component. A component starts with an upper case letter, like
+    // <Foo />
+    if *parent_name != parent_name.to_lowercase() {
+      return;
+    }
+
+    // Check for custom elements wher we cannot make assumptions about
```
bartlomieju, line 84:
```suggestion
    // Check for custom elements where we cannot make assumptions about
```

bartlomieju (approved):
LGTM

PR #1173, opened 2023-06-27
feat: add fresh-handler-export rule

marvinhagemeister:
This adds a new lint rule which helps when fresh users accidentally export middlewares under `handlers` instead of `handler`.

See https://github.com/denoland/fresh/issues/1350


bartlomieju (approved):
LGTM

PR #1176, opened 2023-07-12
feat: add no-await-in-sync-fn rule

marvinhagemeister:
Noticed in https://github.com/denoland/fresh/pull/1454 that the default v8 warning is a bit meh. This lint rule should provide a better editor experience.


bartlomieju (commented):
I think this should be in the recommended set

bartlomieju (approved):
LGTM

Issue #1177, opened 2023-07-16
warn about sync usage inside async function


sigmaSd:
I think it would be nice to have a lint that warns like this
```ts
async function doStuff() {
   Deno.readTextFileSync("") // warning here: using a sync function inside an async function, this blocks deno event loop, you should consider changing it to an async function 
  }
```

Don't know how this stuff is implemented, but I imagine that sync function blocks tokio executor, so no other async function will proceed, here is an example program that showcases my assumption
```ts
setInterval(() => console.log(4), 1000);
//Deno.readTextFileSync("/dev/random"); // if this is uncommented, no logging will be printed
//await Deno.readTextFile("/dev/random"); // if this is uncommneted, logging will still work
```

PR #1178, opened 2023-07-16
feat: add no-sync-fn-in-async-fn rule

sigmaSd:
fix https://github.com/denoland/deno_lint/issues/1177


sigmaSd:
image

sigmaSd:
I add a suggestion for the async variant

image

sigmaSd:
seems like there is a lot of code like this https://github.com/search?q=%2Fasync.*%5Cn.*Deno.*Sync%2F+language%3ATypeScript&type=code

sigmaSd:
couple of notes
- this won't catch wrapped blocking function like `function a() {Deno.readSync..} async function b() {a()}`
- console.log is also blocking, there could be other functions ?
- rust issues https://github.com/rust-lang/rust-clippy/issues/4377 https://github.com/rust-lang/wg-async/issues/19

src/rules/no_sync_fn_in_async_fn.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    use deno_ast::view::Expr;
+    if_chain! {
+      if let Expr::Ident(obj) = &member_expr.obj;
+      if ctx.scope().is_global(&obj.inner.to_id());
+      let obj_symbol: &str = obj.sym();
+      if let Some(prop_symbol) = extract_symbol(&member_expr.prop);
+      if obj_symbol == "Deno";
+      if prop_symbol.contains("Sync");
```
magurotuna, line 97: Sync functions within the Deno namespace seem to have `Sync` at the end. So how about using `ends_with` instead of `contains`?

```suggestion
      if prop_symbol.ends_with("Sync");
```

magurotuna: Alternatively, as we'll get a function name of async equivalent, something like below might be more neat:

```suggestion
      if let Some(async_name) = prop_symbol.strip_suffix("Sync");
```

src/rules/no_sync_fn_in_async_fn.rs
```diff
@@ -0,0 +1,168 @@
...
+      const foo = function(things) {
+        Deno.readTextFileSync("");
+      }
+      "#,
+          r#"
+      class Foo {
+        foo(things) {
+        Deno.readTextFileSync("");
+        }
+      }
```
magurotuna, line 140: nit
```suggestion
      class Foo {
        foo(things) {
          Deno.readTextFileSync("");
        }
      }
```

docs/rules/no_sync_fn_in_async_fn.md
```diff
@@ -0,0 +1,37 @@
+Disallow sync function inside async function
+
+Using sync functions like `Deno.readTextFileSync` blocks the deno event loop so
+its not recommended to use it inside of an async function, because its stops
+progress of all other async tasks.
```
magurotuna, line 5:
```suggestion
Using sync functions like `Deno.readTextFileSync` blocks the deno event loop so
it's not recommended to use it inside of an async function, because it stops
progress of all other async tasks.
```

docs/rules/no_sync_fn_in_async_fn.md
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const fooFn = function foo() {
+  Deno.readTextFileSync("");
+};
+
+const fooFn = () => {
+  Deno.readTextFileSync("");
+};
+```
```
magurotuna, line 37: It would be even better to also show valid examples with `async` functions containing async version of built-in functions, because we prefer async functions in general. Something like:

```ts
async function foo() {
  await Deno.readTextFile("");
}
```

magurotuna (approved):
LGTM. Thanks so much!

PR #1181, opened 2023-07-25
Implement autofix

### LINKED ISSUE #1013: Ability to automatically fix linting errors

yacinehmito:
This PR implements autofix for a small selection of rules. 

Related PR for Deno CLI: https://github.com/denoland/deno/pull/19872

Closes #1013.

### DIFF
--- a/src/rules/no_extra_boolean_cast.rs
+++ b/src/rules/no_extra_boolean_cast.rs
@@ -226,139 +226,194 @@ mod tests {
   fn no_extra_boolean_cast_invalid() {
     assert_lint_err! {
       NoExtraBooleanCast,
-      "if (!!foo) {}": [
-        {
-          col: 4,
-          message: NoExtraBooleanCastMessage::DoubleNegation,
-          hint: NoExtraBooleanCastHint::DoubleNegation,
-        }
-      ],
-      "do {} while (!!foo)": [
-        {
-          col: 13,
-          message: NoExtraBooleanCastMessage::DoubleNegation,
-          hint: NoExtraBooleanCastHint::DoubleNegation,
-        }
-      ],
-      "while (!!foo) {}": [
-        {
-          col: 7,
-          message: NoExtraBooleanCastMessage::DoubleNegation,
-          hint: NoExtraBooleanCastHint::DoubleNegation,
-        }
-      ],
-      "!!foo ? bar : baz": [
-        {
-          col: 0,
-          message: NoExtraBooleanCastMessage::DoubleNegation,
-          hint: NoExtraBooleanCastHint::DoubleNegation,
-        }
-      ],
-      "for (; !!foo;) {}": [
-        {
-          col: 7,
-          message: NoExtraBooleanCastMessage::DoubleNegation,
-          hint: NoExtraBooleanCastHint::DoubleNegation,
-        }
-      ],
-      "!!!foo": [
-        {
-          col: 0,
-          message: NoExtraBooleanCastMessage::DoubleNegation,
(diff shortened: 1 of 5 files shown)


yacinehmito:
This is very much WIP. At the moment I only added the ability to test for the output of a fix. I am still figuring out how to do this.

@bartlomieju I see that ESLint implements fixes by letting rules provide snippets of text to insert in the source file at the location of an AST node, while Rome does so by lettings rules provide a description of AST mutations to perform. I am still unsure which path to take. The former (operate on text) seems easier to implement but fraught with correctness issues, while the later (operate on the AST) is the opposite. Any advice?

src/rules/no_extra_boolean_cast.rs
```diff
@@ -373,118 +428,166 @@ mod tests {
...
+      "!Boolean((foo, bar()));": {
+        "!(foo, bar());",
+        [
+          {
+            col: 0,
+            message: NoExtraBooleanCastMessage::BooleanCall,
+            hint: NoExtraBooleanCastHint::BooleanCall,
+          }
+        ]
+      },
```
magurotuna, line 440: This macro may be a bit harder to read, probably because there's no indicator on what `"!(foo, bar());"` means. What about the following syntax, for instance?

```rs
      "!Boolean((foo, bar()));": {
        autofixed: "!(foo, bar());",
        errors: [
          {
            col: 0,
            message: NoExtraBooleanCastMessage::BooleanCall,
            hint: NoExtraBooleanCastHint::BooleanCall,
          }
        ]
      },
```

magurotuna:
> I see that ESLint implements fixes by letting rules provide snippets of text to insert in the source file at the location of an AST node, while Rome does so by lettings rules provide a description of AST mutations to perform. I am still unsure which path to take. The former (operate on text) seems easier to implement but fraught with correctness issues, while the later (operate on the AST) is the opposite. Any advice?

That's very interesting, and I agree that text transformation is easier to implement but less robust in some cases, and AST-based way is the opposite. I would spend some time trying to figure out if the latter is something doable

src/rules/no_extra_boolean_cast.rs
bartlomieju: I agree with @magurotuna here - may I suggest to create a completely new macro so we don't overload the existing one further.

I wonder if we should have some kind of a sanity check in the macro to make sure that the proposed output actually passes linting

bartlomieju:
> > I see that ESLint implements fixes by letting rules provide snippets of text to insert in the source file at the location of an AST node, while Rome does so by lettings rules provide a description of AST mutations to perform. I am still unsure which path to take. The former (operate on text) seems easier to implement but fraught with correctness issues, while the later (operate on the AST) is the opposite. Any advice?
> 
> That's very interesting, and I agree that text transformation is easier to implement but less robust in some cases, and AST-based way is the opposite. I would spend some time trying to figure out if the latter is something doable

I agree with @magurotuna here, but it's hard to decide right now. The latter is certainly doable - SWC has infrastructure to modify AST nodes and then write them back as string. The problem I see here is that after fixing errors in certain rule, we would have to re-parse and re-lint all the files that were fixed. Not sure if that's something that can be avoided but we should keep that in mind. 

@yacinehmito thanks for starting the PR, I will give it some thought over the next week and come back with some suggestions on how to proceed.

PR #1182, opened 2023-07-30
feat(www): playground

magurotuna:
This implements a playground to lint.deno.land where we can write random source code and see the diagnostics right away.

Currently the wasm file and its glue code are generated locally and committed to the repo, which means that we need to run the manual updating process (just running `deno task wasmbuild`) every time we make changes to the linter. We could automate it by moving this process inside CI, but this isn't done here because that would require changing build settings in Deno Deploy too. Instead, we add a step to CI to check if the wasm files committed to the repo are up-to-date.

You can try it out at https://deno-lint--playground.deno.dev/playground

### Demo

https://github.com/denoland/deno_lint/assets/23649474/c926d65b-df57-4f6b-aae4-55bb35535590


www/static/deno_lint.js
```diff
@@ -0,0 +1,240 @@
+let wasm;
```
dsherret, line 1: Is there a way we could have this js file, the wasm file, and the .d.ts file built when creating the website? Then we don't need to keep updating these and they don't bloat the repository.

dsherret (commented):
This will be nice! It would be good if we could not check in the Wasm build artifacts though.

sigmaSd:
That's cool!

Couple of remarks:
- Deno apis seems to be not understood by the lsp  / same thing with top level awaits
- Can we have an option to use all rules

sigmaSd:
- the editor doesn't react to theme changes (I need to refresh the page)
[Screencast from 2023-09-01 08-14-57.webm](https://github.com/denoland/deno_lint/assets/22427111/df8125ba-344d-4db2-9326-bd6ffcfc45cc)

magurotuna:
@sigmaSd Thanks for the feedback. What do you mean by "not understood by the lsp"? Like there's no completion options showing up when you put `Deno.` in the editor part of the playground?

www/README.md
```diff
@@ -9,3 +9,40 @@ deno task start
...
+#### How to update wasm files used in playground
+
+This feature is realized using wasm and its glue code put in the `static`
+directory. To update these files, follow the steps below. (Note: These steps
+assume that you are in the root directory of this repository)
+
+1. Make sure you have `wasm-bindgen-cli` v0.2.81 installed:
+
+```shell
+cargo install -f wasm-bindgen-cli --version 0.2.81
```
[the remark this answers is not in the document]
magurotuna: Oh I forgot that one! We should definitely use that

www/static/deno_lint.js
magurotuna: Yeah I was thinking about that, and put it off for the moment to focus on the implementation part. I'll try to find a way to achieve continuous delivery.

sigmaSd:
It looks like this 
image

www/README.md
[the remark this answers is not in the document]
magurotuna: Updated to use wasmbuild. This simplified the build process significantly 👍

www/static/deno_lint.js
magurotuna: Thinking about it a little bit further, I'd prefer to have developers commit the wasm files into the repo instead of relying on CI to build them, mostly because building them in CI would require configuration changes in Deno Deploy. Of course we could do that, but I think we can do so later if it's really needed (for instance we may want to do AoT build for fresh, which can be done together with wasm build in CI)

magurotuna:
I made several updates and I think now it's ready for review again. PR description has been updated too, but here's a quick summary:

- Add language select box specifically to support JSX(TSX) syntax
- Add checkbox to enable all rules (default is recommended rules only)
- Update color theme in monaco dynamically on users' preference (see the last part of the demo movie)

www/deno.json
```diff
@@ -4,19 +4,25 @@
     "build": "deno run -A dev.ts build",
     "preview": "deno run -A main.ts"
   },
-  "lock": false,
+  "lock": true,
```
magurotuna, line 7: Note for reviewers: using libraries from jsr seems to require this

src/rules/no_external_imports.rs
```diff
@@ -55,10 +55,11 @@ struct NoExternalImportHandler;
 impl NoExternalImportHandler {
   fn check_import_path(&self, decl: &ImportDecl, ctx: &mut Context) {
     let parsed_src = ModuleSpecifier::parse(decl.src.value());
-    let maybe_file_path = ctx.specifier().to_file_path().ok();
-    let file_stem = maybe_file_path
-      .as_ref()
-      .and_then(|p| p.file_stem())
```
magurotuna, line 61: for reviewers: Url's `to_file_path` method isn't provided for wasm32-unknown-unknown. As a workaround, we extract the last segment from specifier's path part, and treat it as a file name.

bartlomieju:
**Verdict: needs human decision** (large feature, likely stale/possibly superseded) — not merging. A few concrete code notes below.

Reviewed the Rust/WASM surface in depth; skimmed the Fresh islands.

**Correctness**
- `src/wasm.rs` `MietteSourceCode::read_span` — the `end_line_index` clamp compares a *line index* against a *byte length*:
  ```rust
  let end_line_index = std::cmp::min(
    end_line_column.line_index + context_lines_after,
    self.source.text_str().len(),   // byte length, not a line bound
  );
  let src_end = self.source.line_end(std::cmp::min(end_line_index, line_count - 1));
  ```
  `text_str().len()` (bytes) is effectively always ≥ any line index, so the first `min` is dead code; the real clamp is `line_count - 1` on the next line. Not a live bug, but it should be removed or corrected to `line_count` to avoid confusion.
- `reporter.render_report(&mut s, &miette_diag).unwrap()` — an `.unwrap()` in a `#[wasm_bindgen]` path aborts the whole WASM instance on any render error. Prefer `.map_err(|e| e.to_string())?` so it surfaces as a JS error instead of a hard crash.
- `src/rules/no_external_imports.rs` — this is a **behavior change to an existing rule**, done so it works under WASM (no `to_file_path()`). Switching from `specifier().to_file_path()` to `path_segments().next_back()` changes the filename derivation: `to_file_path()` percent-*decodes*, `path_segments()` yields percent-*encoded* segments. So a specifier like `file:///my%20deps.ts` derives a different `file_stem` than before. Edge case, but it's a semantics change in a shipped rule and deserves a test + explicit sign-off.

**Performance / deps**
- `#[global_allocator] wee_alloc` — `wee_alloc` is unmaintained and has a well-known non-freeing/leak issue. For a long-lived browser playground that lints repeatedly, that can accumulate. Recommend dropping it and using the default allocator (or `dlmalloc`), especially since code size isn't critical here.
- `display_diagnostics` rebuilds `MietteSourceCode` inside the per-diagnostic loop; it only borrows the same `text_info`/`filename`, so it can be hoisted out. Negligible for playground-sized inputs — non-blocking.

**Staleness / integration (main blockers)**
- Branch is stale (`mergeable: UNKNOWN`, no CI). It predates `LintDiagnostic::severity` (added recently) and the current linter/`LintFileOptions` API, so `wasm.rs` likely won't compile against `main` without updates, and there will be conflicts.
- The new CI step `deno task wasmbuild --check` requires the checked-in `www/static/deno_lint_bg.wasm` to be regenerated after any rebase, or CI fails.
- Worth confirming with maintainers whether this `www/` playground is still the intended home — docs/site tooling has moved since this was opened, so this may be partly superseded.

_Recommend: maintainer decides scope/supersession first; if kept, needs a rebase, the two correctness fixes above, a test for the `no_external_imports` change, and a wasm rebuild._

PR #1183, opened 2023-08-07
fix(breaking): remove rule no-extra-semi

### LINKED ISSUE #1033: Remove no-extra-semi from the recommended set?
I think we should remove this rule from the recommended set and perhaps entirely. It’s a formatter concern and not a linter one.

rotu:
This rule is a formatter concern, not a linter one.
Fixes #1033


rotu:
Fixes https://github.com/denoland/deno/issues/20078

dsherret (approved):
LGTM. Thanks!

PR #1187, opened 2023-08-16
feat(camelcase): allow for non-declaration object destructuring and named import when has no alias

dsherret:
This ignores the camelcase rule for non-declaration object destructuring and named imports when the name has no local alias.

The rationale behind this change is when you are doing these actions, it indicates you have no control over the identifier.

For example:

```ts
// valid
import { op_something } from "something";
const {
  op_arraybuffer_was_detached,
  op_transfer_arraybuffer,
  op_readable_stream_resource_allocate,
  op_readable_stream_resource_get_sink,
  op_readable_stream_resource_write_error,
  op_readable_stream_resource_write_buf,
  op_readable_stream_resource_close,
  op_readable_stream_resource_await_close,
} = core.ensureFastOps();

// still invalid as this is a declaration
function myFunc({ camel_case }) {}
```

### DIFF
--- a/src/rules/camelcase.rs
+++ b/src/rules/camelcase.rs
@@ -3,7 +3,7 @@ use super::{Context, LintRule};
 use crate::handler::{Handler, Traverse};
 use crate::swc_util::StringRepr;
 
-use deno_ast::view::{NodeKind, NodeTrait};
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

PR #1189, opened 2023-08-17
feat(rules): add no console log rule

0xIchigo:
The goal of this PR is to address the [this issue](https://github.com/denoland/deno/issues/20187) presented in the main Deno GitHub repository, which proposes the implementation of a `no-console` rule similar to that of [eslint](https://eslint.org/docs/latest/rules/no-console). In order to add such a rule to `deno lint` in the main Deno repository, we need a rule defined in the linter itself.

This PR achieves that goal by adding in a `NoConsoleLog` lint rule that checks whether a `console.log()` statement is in the specified code

### DIFF
--- a/docs/rules/no_console_log.md
+++ b/docs/rules/no_console_log.md
@@ -17,10 +17,6 @@ function log() { console.log("Log"); }
 ### Valid
 
 ```typescript
-let foo = 0;
-
-const bar = 1;
-
 console.error("Error message");
 
 function log_error(message: string) {
--- a/docs/rules/no_console_log.md
+++ b/docs/rules/no_console_log.md
@@ -2,7 +2,11 @@
 
 Disallows the use of `console.log`.
 
-Oftentimes, developers are guilty of committing `console.log` statements accidentally, left in particularly after debugging. Moreover, using `console.log` in code may leak sensitive information to the output or clutter the console with unnecessary information. This rule helps maintain clean and secure code by disallowing the use of `console.log`.
+Oftentimes, developers are guilty of committing `console.log` statements
+accidentally, left in particularly after debugging. Moreover, using
+`console.log` in code may leak sensitive information to the output or clutter
+the console with unnecessary information. This rule helps maintain clean and
+secure code by disallowing the use of `console.log`.
 
 ### Invalid
 
@@ -20,6 +24,6 @@ function log() { console.log("Log"); }
 console.error("Error message");
 
 function log_error(message: string) {
-    console.warn(message);
+  console.warn(message);
 }
 ```
--- a/docs/rules/no_console_log.md
+++ b/docs/rules/no_console_log.md
@@ -2,20 +2,22 @@
 
 Disallows the use of `console.log`.
 
-Oftentimes, developers are guilty of committing `console.log` statements
-accidentally, left in particularly after debugging. Moreover, using
-`console.log` in code may leak sensitive information to the output or clutter
-the console with unnecessary information. This rule helps maintain clean and
-secure code by disallowing the use of `console.log`.
+Oftentimes, developers accidentally commit `console.log` statements, left in
+particularly after debugging. Moreover, using `console.log` in code may leak
+sensitive information to the output or clutter the console with unnecessary
+information. This rule helps maintain clean and secure code by disallowing the
+use of `console.log`.
 
 ### Invalid
 
 ```typescript
 console.log("Debug message");
 
-if debug { console.log("Debugging"); }
+if (debug) console.log("Debugging");
 
-function log() { console.log("Log"); }
+function log() {
+  console.log("Log");
+}
 ```
 
 ### Valid
--- a/docs/rules/no_console_log.md
+++ b/docs/rules/no_console_log.md
@@ -1,0 +1,29 @@
+# no_console_log
+
+Disallows the use of `console.log`.
+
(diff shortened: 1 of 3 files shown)


docs/rules/no_console_log.md
```diff
@@ -0,0 +1,29 @@
+# no_console_log
+
+Disallows the use of `console.log`.
+
+Oftentimes, developers are guilty of committing `console.log` statements accidentally, left in particularly after debugging. Moreover, using `console.log` in code may leak sensitive information to the output or clutter the console with unnecessary information. This rule helps maintain clean and secure code by disallowing the use of `console.log`.
+
+### Invalid
+
+```rust
```
lino-levan, line 9:
```suggestion
```typescript
```

docs/rules/no_console_log.md
```diff
@@ -0,0 +1,29 @@
...
+console.log("Debug message");
+
+if debug { console.log("Debugging"); }
+
+fn log() { console.log("Log"); }
```
lino-levan, line 14:
```suggestion
function log() { console.log("Log"); }
```

docs/rules/no_console_log.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+### Valid
+
+```rust
```
lino-levan, line 19:
```suggestion
```typescript
```

docs/rules/no_console_log.md
```diff
@@ -0,0 +1,29 @@
...
+let foo = 0;
+
+const bar = 1;
+
+console.error("Error message");
+
+fn log_error(message: &str) {
```
lino-levan, line 26:
```suggestion
function log_error(message: string) {
```

docs/rules/no_console_log.md
```diff
@@ -0,0 +1,29 @@
...
+    eprintln!("{}", message);
```
lino-levan, line 27:
```suggestion
    console.warn(message);
```

lino-levan (changes requested):

0xIchigo:
Damn, thank you for the suggested changes @lino-levan! My brain has been in Rust-mode for the last little while and I didn't even realize I put the code examples in Rust instead of TypeScript :tired_face:

The suggested changes have been implemented with [this commit](https://github.com/denoland/deno_lint/pull/1189/commits/1cbd2df0271c499b82920fa11f4a525b670364ea)

docs/rules/no_console_log.md
```diff
@@ -0,0 +1,29 @@
...
+
+function log() { console.log("Log"); }
+```
+
+### Valid
+
+```typescript
+let foo = 0;
+
+const bar = 1;
```
lino-levan, line 22: not sure what these are doing here

0xIchigo: Sorry, I can take those out - they were just to show that any other line of code that does not contain `console.log` is valid. Fixed with [this commit](https://github.com/denoland/deno_lint/pull/1189/commits/f21a23064e0fb5195390d33fa4307cf305c6a74f)

docs/rules/no_console_log.md
```diff
@@ -0,0 +1,25 @@
+# no_console_log
+
+Disallows the use of `console.log`.
+
+Oftentimes, developers are guilty of committing `console.log` statements accidentally, left in particularly after debugging. Moreover, using `console.log` in code may leak sensitive information to the output or clutter the console with unnecessary information. This rule helps maintain clean and secure code by disallowing the use of `console.log`.
```
marvinhagemeister: I feel like putting a negative word like "guilty" is a bit much for official docs, given that this is usually done unintentionally. Maybe we an say something like: "Oftentimes, developers accidentally commit `console.log` statements, ..."

marvinhagemeister (approved):
LGTM, I leave it to @dsherret to merge 👍

0xIchigo:
Thank you for the suggestions @marvinhagemeister! I agree with them and have fixed with [this commit](https://github.com/denoland/deno_lint/pull/1189/commits/81b4a23cd8193245dd7fd54b22da48bdef37c960). LGTM 🔥

bartlomieju:
I'm not convinced this should be in the recommended set - this means that any CLI tool that wants to output to console will spawn a plethora of errors. Maybe we should add it by default to the `fresh` set and other users can enable this rule as they see fit?

marvinhagemeister:
@bartlomieju good point. Let's just add this rule without adding it to any preset. I don't want this to enable by default for `fresh` users but I'd like to opt-in to that for the fresh codebase itself.

0xIchigo:
Gentle bump @bartlomieju - is there anything else to add in order to make this an optional rule, not added to any preset by default, or is adding the functionality to the `rules` directory enough?

lev-blit:
maybe this is for a future PR, but I find myself committing `console.debug` a lot as well (perhaps even more so than `console.log`)...

perhaps it would be useful to be able to customize the behavior of this like it's possible in eslint's [no-console rule](https://eslint.org/docs/latest/rules/no-console#options) to either disallow all `console` functions or to specify which are allowed (`error` for example).

0xIchigo:
@lev-blit yeah, that wouldn't be too hard to implement! I don't mind opening up another PR to tackle that issue once this one is merged. The nice part of how the code is structured is that in the implementation for `NoConsoleLogHandler` we have the following `if` check: `if prop_ident.sym().as_ref() == "log"` - you can simply swap out the `"log"` for `"debug"` in an entirely new rule, or modify the code to respond with different messages and errors based on whether you have a `console.log()` or a `console.debug()`

dsherret (approved):
LGTM, though I think this should align with eslint. I'm going to merge this PR as-is, then do a follow-up PR in a few minutes to convert it to be like eslint and remove it from the recommended set.

PR #1194, opened 2023-08-31
feat: change no-console-log to no-console

dsherret:
This changes the no-console-log rule to align with ESLint and instead be no-console. Developers can then opt-into using the console for specific actions via a `// deno-lint-ignore no-console` comment.

Follow-up to https://github.com/denoland/deno_lint/pull/1189 (I'm doing an swc upgrade with deno-lint upgrade so wanted to get this feature in, thus this pr)


bartlomieju (approved):

examples/dlint/testdata/issue1145_no_trailing_newline.ts
```diff
@@ -1,3 +1,3 @@
 /*---
 ---*/
-var base
\ No newline at end of file
+var base;
```
dsherret, line 3: Oh, lol. I ran deno fmt on the code.

PR #1197, opened 2023-09-09
fix(prefer-primordials): Ignore `TsTypeRef`

petamoriken:
I removed false positives in the Deno internal TypeScript code


bartlomieju (approved):
LGTM, thanks!

PR #1199, opened 2023-10-02
upgrade no-sync-fn-in-async-fn

### LINKED ISSUE #1198: upgrade no-sync-fn-in-async-fn
currently `no-sync-fn-in-async-fn` can't catch this:

```ts
async function a() {
   b() // b is blocking, but its hidden from us, we only warn if we see Deno. Sync syntax
}

function b() {
 Deno.xxxxxSync()
}
```

I thought maybe its possible to catch even this:  for example we can consider each function that have  inside it a deno Sync api a blocking function, and save it on a list and now each time we see it we also consider the parent blocking and so on, and now if we see one of these in an async context we emit a warning.

Do you think this is a reasonable idea ?

sigmaSd:
fix https://github.com/denoland/deno_lint/issues/1198

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+channel = "1.70.0"
+components = ["clippy", "rustfmt"]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,0 @@
-[toolchain]
-channel = "1.70.0"
-components = ["clippy", "rustfmt"]
--- a/src/rules/fresh_handler_export.rs
+++ b/src/rules/fresh_handler_export.rs
@@ -60,7 +60,9 @@ impl Handler for Visitor {
     let id = match export_decl.decl {
       Decl::Var(var_decl) => {
         if let Some(first) = var_decl.decls.first() {
-          let Pat::Ident(name_ident) = first.name else {return};
+          let Pat::Ident(name_ident) = first.name else {
+            return;
+          };
           name_ident.id
         } else {
           return;
--- a/src/rules/fresh_handler_export.rs
+++ b/src/rules/fresh_handler_export.rs
@@ -60,9 +60,7 @@ impl Handler for Visitor {
     let id = match export_decl.decl {
       Decl::Var(var_decl) => {
         if let Some(first) = var_decl.decls.first() {
-          let Pat::Ident(name_ident) = first.name else {
-            return;
-          };
+          let Pat::Ident(name_ident) = first.name else {return};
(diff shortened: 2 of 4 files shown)


sigmaSd:
- Currently this approach just scan identifiers as strings, can this be better ?
- Can this be made to work across the whole module instead of just the current file ?

bartlomieju:
**Verdict: WIP (draft) — needs changes before it can land.** Not merging (draft + behavior change). The interprocedural approach (mark functions that call `Deno.*Sync`, then flag calls to them inside async fns) is a reasonable direction for #1198; notes below for whoever continues it.

**Correctness (the important ones)**
- **Double traversal produces duplicate / order-dependent diagnostics.** `lint_program_with_ast_view` now calls `rule.traverse(...)` twice, but `member_expr` both *collects* `blocking_fns` **and** *emits* the direct sync-in-async diagnostic on every pass. So the existing direct case will be reported twice, and the new transitive case depends on whether the blocking fn was collected before the call site was visited. You want a clean two-phase split: pass 1 collects `blocking_fns` only (no diagnostics), pass 2 emits only. Right now both phases do both. (The author's own test comments — "should panic / TODO switch to assert_err" — confirm this isn't settled.)
- **Name-based matching via `node.text().to_string()` ignores scope/binding.** Blocking functions are matched by source text, so any identifier with the same name — a shadowing local, a different `foo` in another scope, a method — will false-positive. Consider resolving to the binding (`to_id()` + scope) as the direct-`Deno` check already does, rather than string equality.
- Only 1-hop transitivity: a function that calls a blocking function isn't itself marked blocking. Fine as a first cut, but worth stating as a known limit.

**Performance**
- `blocking_fns: Vec<String>` with `.contains()` is an O(n) scan per identifier visited, and `node.text().to_string()` allocates a `String` for *every* ident/call/var node in the file (whether or not it matches). On large files that's a lot of wasted allocation + an O(nodes × fns) product. Use a `HashSet<String>` (or set of resolved ids) and avoid the `to_string()` until you actually need to compare.

**Minor / will trip CI**
- Unused bindings: `Expr::Ident(ident) => ...` in `handle_paren_callee`, `var_declarator`, and `call_expr` bind `ident` but use `.as_node()` instead — these will raise `unused_variables` warnings (CI denies warnings). Prefix with `_` or drop the binding.
- Typo in the hint: `"consier changing"` → `"consider changing"`.

_Left as draft for the author/another agent: split collect vs. emit into two real phases, switch to binding-based matching + a `HashSet`, fix the unused vars and typo, then convert the placeholder `assert_lint_ok!` cases to `assert_lint_err!`._

PR #1201, opened 2023-10-11
fix: dlint has off-by-one reports

bartlomieju:


bartlomieju:
Closing as stale. This was an unfinished WIP — it still contains leftover `eprintln!` debug output and a TODO, and the actual CLI users run lives in the `deno` repo rather than this example harness. If there's a real off-by-one to fix, it should be done cleanly.

PR #1218, opened 2023-12-19
refactor: Linter, Context and more

bartlomieju:
This is just a cleanup in preparation for another plugin system prototype.

With this commit "Linter" is a struct that should be created only once and used
to lint many files, instead of creating a new instance for every single file.

The API was changed to "Linter::lint_file" that accepts a "LintFileOptions".


dsherret (approved):
LGTM

PR #1219, opened 2023-12-24
docs: fix link rot on `no-external-import`

nekowinston:
Just ran into a 404 here :)


magurotuna (approved):
Good catch, thank you!

PR #1228, opened 2024-01-16
feat: Update to APIs deprecated in Deno v1.40

bartlomieju:
Still WiP


iuioiua:
Note: we can use https://github.com/denoland/deno/issues/22079#issue-2098002747

Hopefully, we can land this for 1.40.

Issue #1122, opened 2023-01-29
Rule suggestion: no-self-compare


thegaryroberts:
As per: https://eslint.org/docs/latest/rules/no-self-compare

magurotuna:
That rule makes lots of sense, it would be great to have that one.

thegaryroberts:
It will help prevent nasty bugs in codebases that makes heavy use of destructuring (with shadowed variable names in closures)

AkinAguda:
### tl;dr
Are we protecting against `if (someVar === someVar)` alone or that and also `if ("stringLiteral" === "stringLiteral")`

### Full Version
While working on this, I noticed these tests in eslint:

https://github.com/eslint/eslint/blob/main/tests/lib/rules/no-self-compare.js#L42-L49

My question is, those tests are checking that the left and right of the operator are equal `string`s, not equal `Identifier`s or `MemberExpression`s.

Do we want that? Because what we are guarding against as described in the discussion above and the eslint rule description should not consider comparing if the left and right are both equal strings or am I misunderstanding something?

<img width="909" alt="Screenshot 2024-01-19 at 00 06 51" src="https://github.com/denoland/deno_lint/assets/39712415/e6928cd3-2ddc-47c8-8e9e-7d96e3521b54">

magurotuna:
Closing as it's added by #1230

PR #1230, opened 2024-01-19
feat: add no-self-compare rule

AkinAguda:
related: #1122


AkinAguda:
@magurotuna Please take a look

PR #1240, opened 2024-01-28
chore: disable lock

iuioiua:
This was also done because of Fresh's use of [esm.sh](https://esm.sh/) and [this](https://github.com/denoland/deno-gfm/issues/47#issuecomment-1460872019).


hashrock:
Sorry, I didn't notice. Thank you! This was done in #1245

PR #1242, opened 2024-01-31
docs: point to new "Deno 1.x to 2.x Migration Guide"

iuioiua:
Prerequisite https://github.com/denoland/deno-docs/pull/300


iuioiua (commented):
Couple questions:
1. Is it even worth listing out the individual deprecated APIs? Or rather, could we just say something like "see the migration guide for the full list of deprecated APIs and their migration instructions".
2. Is there a way to detect use of deprecated `.rid` properties and constructors?

bartlomieju:
@kwhinnery could you please take a look?

kwhinnery (approved):

PR #1250, opened 2024-02-14
fix: incorrect lint assumptions  for `no-empty-interface`

### LINKED ISSUE #1249: Interface with no members but with `extend` is not equivalent of the supertype
An interface that extends, but has no members isn't equivalent of the supertype when interface has been already declared.

Stated by the lint description:
> An interface declaring no members is equivalent to its supertype.
> Use the supertype instead, or add members to this interface.

It is in fact not true, since `interface ABC extends XYZ {}` extends the existing `ABC` interface by the members of `XYZ`.
And in case of the interface being defined for the first time, the same applies – it can be later extended with other `interface` declarations, where `type` cannot.

### Lint Name

`no-empty-interface`

### Code Snippet

```ts
interface ABC {
  abc: true;
}

interface XYZ {
  xyz: true;
}

interface ABC extends XYZ {}; // deno-lint(no-empty-interface)

TypeEquals<ABC, {
  abc: true;
  xyz: true;
}>;
```

```ts
type MappedType = {
  [thing in "a" | "b" | "c"]: true;
};

interface ABC extends MappedType {}; // workaround around not being able to use mapped types in interfaces. This allows `ABC` to be extended with other interface declarations.

interface ABC {
  abc: true;
}

TypeEquals<ABC, {
  abc: true;
  a: true;
  b: true;
  c: true;
}>;
```

### Expected Result
 - `interface ABC {}` should error out for being empty
 - `interface ABC extends XYZ {}` shouldn't error 

### Actual Result
 - `interface ABC extends XYZ {}` errors
 
### Additional Info

`interface ABC extends XYZ {}` can be useful in scenarios, where global namespaces are augmented to allow extending given type definitions by the module.

### Version

```
deno 1.40.4 (release, x86_64-unknown-linux-gnu)
v8 12.1.285.6
typescript 5.3.3
```

Im-Beast:
This PR removes linting error for `interface ABC extends XYZ {}` while keeping lint errors for empty interfaces with no extends (`interface ABC {}`).

Fixes #1249

### DIFF
--- a/docs/rules/no_empty_interface.md
+++ b/docs/rules/no_empty_interface.md
@@ -1,7 +1,7 @@
 Disallows the declaration of an empty interface
 
-An interface with no members serves no purpose.
-This rule will capture these situations as either unnecessary code or a mistaken empty implementation.
+An interface with no members serves no purpose. This rule will capture these
+situations as either unnecessary code or a mistaken empty implementation.
 
 ### Invalid:
 
--- a/docs/rules/no_empty_interface.md
+++ b/docs/rules/no_empty_interface.md
@@ -1,16 +1,12 @@
 Disallows the declaration of an empty interface
 
-An interface with no members serves no purpose. Either the interface extends
-another interface, in which case the supertype can be used, or it does not
-extend a supertype in which case it is the equivalent to an empty object. This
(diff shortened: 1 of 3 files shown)


dsherret (approved):
LGTM. Thanks!

PR #1255, opened 2024-02-28
feat: quick fixes

### LINKED ISSUE #1013: Ability to automatically fix linting errors

dsherret:
Closes #1181
Closes #1013

### DIFF
--- a/src/diagnostic.rs
+++ b/src/diagnostic.rs
@@ -15,15 +15,15 @@ use deno_ast::SourceRange;
 use deno_ast::SourceTextInfo;
 
 #[derive(Debug, Clone)]
-pub struct LintQuickFixChange {
+pub struct LintFixChange {
   pub new_text: Cow<'static, str>,
   pub range: SourceRange,
 }
 
 #[derive(Debug, Clone)]
-pub struct LintQuickFix {
+pub struct LintFix {
   pub description: Cow<'static, str>,
-  pub changes: Vec<LintQuickFixChange>,
+  pub changes: Vec<LintFixChange>,
 }
 
 #[derive(Clone)]
@@ -34,7 +34,7 @@ pub struct LintDiagnostic {
   pub message: String,
   pub code: String,
   pub hint: Option<String>,
-  pub quick_fixes: Vec<LintQuickFix>,
+  pub fixes: Vec<LintFix>,
 }
 
 impl Diagnostic for LintDiagnostic {
--- a/src/diagnostic.rs
+++ b/src/diagnostic.rs
@@ -14,6 +14,18 @@ use deno_ast::ModuleSpecifier;
 use deno_ast::SourceRange;
 use deno_ast::SourceTextInfo;
 
+#[derive(Debug, Clone)]
+pub struct LintQuickFixChange {
+  pub new_text: String,
+  pub range: SourceRange,
+}
+
+#[derive(Debug, Clone)]
+pub struct LintQuickFix {
+  pub description: String,
+  pub changes: Vec<LintQuickFixChange>,
+}
+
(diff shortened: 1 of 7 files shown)


dsherret:
I just noticed https://github.com/denoland/deno_lint/pull/1181 -- I think this PR is a better approach because it will also work well in the LSP.

src/diagnostic.rs
```diff
@@ -14,6 +14,18 @@ use deno_ast::ModuleSpecifier;
...
+pub struct LintQuickFixChange {
+  pub new_text: Cow<'static, str>,
+  pub range: SourceRange,
+}
+
+#[derive(Debug, Clone)]
+pub struct LintQuickFix {
+  pub description: Cow<'static, str>,
+  pub changes: Vec<LintQuickFixChange>,
+}
```
magurotuna, line 27: Correct me if I'm wrong, but my understanding is that the user of `deno_lint` crate will use these quickfix information as suggestions, and end users will decide what quickfix to apply probably via LSP action, or some other interactive way - is this correct?

What I'm imagining is that, people who work with very large code base may want to apply quickfixes automatically without any human intervention. In this scenario, they probably would like something like `deno lint --rule no-window-prefix` --autofix, but one problem with this would be that one fix needs to be picked in case there are multiple candidates for quickfixes, which would require human intervention.
I think one possible solution for this is give each quick fix a unique ID, and let end users specify a quickfix ID they want to use e.g. `deno lint --rule no-window-prefix --autofix use-globalthis`
So in any case, what do you think about giving quickfix ID to each quickfix definition?

dsherret: Yeah in the LSP they'll potentially get multiple options.

For `--fix`, what I'm doing at the moment (in my local branch of cli, didn't push yet) is only the first quick fix gets applied if there are multiple options. So for example, In the window -> globalThis or window -> self case it only will apply the window -> globalThis change.

I think adding ids would be a nice solution in the future, but that we should hold off on it until there's demand for it. I think the current PR is a good starting point to build off of in the future.

In the meantime I'll update the docs in this crate to explain this scenario.

src/rules/no_window_prefix.rs
```diff
@@ -19,6 +21,8 @@ const CODE: &str = "no-window-prefix";
 const MESSAGE: &str = "For compatibility between the Window context and the Web Workers, calling Web APIs via `window` is disallowed";
 const HINT: &str =
   "Instead, call this API via `self`, `globalThis`, or no extra prefix";
+const FIX_GLOBAL_THIS_DESC: &str = "Rename window to globalThis";
+const FIX_SELF_DESC: &str = "Rename window to self";
```
dsherret, line 25: Does it make sense to have `self` here? Maybe this should only be `globalThis`? If we should keep this, then maybe the no_window rule should also have this?

src/diagnostic.rs
magurotuna: OK, that makes sense. Let's revisit it when the need for it arises.

src/rules/no_window_prefix.rs
magurotuna: I think only `globalThis` should suffice.

bartlomieju: I agree that `globalThis` is enough here

bartlomieju (approved):
Very cool, I like how simple this solution is and the fact that it's up to the caller to actually apply the fix :+1 LGTM

PR #1263, opened 2024-04-08
feat(www): add header link to official docs

Delapouite:
Hi

This way users can learn:

- how to run the CLI with various flags
- how to configure the `deno.json` `lint` entry


bartlomieju (approved):
Thanks!

PR #1264, opened 2024-04-09
feat(www): add link to rule source on GitHub

Delapouite:
Hi

By following this link users can learn how the rule is implemented, and discover details about corner cases.

Also it eases potential contributions, if something needs to be fixed or improved about the current rule.


magurotuna (approved):
Thanks, I like this idea! LGTM

PR #1277, opened 2024-05-28
fix(no-unused-vars): enable for jsx files

lucacasonato:
This commit enables the no-unused-vars lint rule for JSX files, by wiring through `jsxFactory` and `jsxFragmentFactory`.


src/linter.rs
```diff
@@ -150,9 +156,15 @@ impl Linter {
   pub fn lint_with_ast(
     &self,
     parsed_source: &ParsedSource,
+    default_jsx_factory: Option<String>,
+    default_jsx_fragment_factory: Option<String>,
```
bartlomieju, line 160: Consider creating an option bag and putting these two there

bartlomieju (approved):

PR #1282, opened 2024-06-04
fix: use macos-13 runner

dsherret:
For https://github.com/denoland/deno/pull/24041


bartlomieju (approved):
LGTM, thanks I forgot about that one. Seems easier to switch this on to x86 than updating lint step in Deno to use ARM

PR #1285, opened 2024-06-12
fix: Make dlint report syntax/parsing errors

HasanAlrimawi:
Supported dlint reporting of syntax/parsing errors.

The commit is made to address issue [23437 from deno repo](https://github.com/denoland/deno/issues/23437)

**Code changes**

Added check for `ParsedSource.diagnostics()` and when it's not empty, an object of `ParseDiagnosticsError` holding the vector of `ParseDiagnostic` objects is returned as an `Err` object within `fn run_linter()` in _deno_lint/examples/dlint/main.rs_ file.

PS: The change within this function was preffered over the suggested one in the issue discussion to maintain the types of returned Err in order to keep the same successful flow, since the run_linter function returns AnyError not some specific one like parse_program function, and also it's much simpler (doesn't need to append the ParseDiagnostic single object that returned from `deno_ast::parse_program` into a vector before returning it as an error).

@lucacasonato

### DIFF
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -127,8 +127,7 @@ fn run_linter(
 
       let mut number_of_errors = diagnostics.len();
       if !parsed_source.diagnostics().is_empty() {
-        number_of_errors =
-          number_of_errors + parsed_source.diagnostics().to_vec().len();
+        number_of_errors += parsed_source.diagnostics().to_vec().len();
         eprintln!(
           "{}",
           ParseDiagnosticsError(parsed_source.diagnostics().to_vec())
--- a/examples/dlint/main.rs
+++ b/examples/dlint/main.rs
@@ -7,6 +7,7 @@ use clap::Command;
 use core::panic;
 use deno_ast::MediaType;
 use deno_ast::ModuleSpecifier;
+use deno_ast::ParseDiagnosticsError;
 use deno_lint::linter::LintConfig;
 use deno_lint::linter::LintFileOptions;
 use deno_lint::linter::LinterBuilder;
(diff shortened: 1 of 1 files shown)


HasanAlrimawi:
Hello @lucacasonato ,
All checks have passed, it would be great If you make a review in order to merge the changes.

magurotuna:
@HasanAlrimawi Thanks for working on it! This looks very good, however this fix only applies to `dlint` which is a standalone binary we can run via `cargo run --example dlint`. So this won't  change any behavior we get when running `deno lint` command.
I think it's totally fine to get this fix merged on main with the current code, but I also think that you may want to modify more code so that the `deno lint` command can report syntax errors. Would you like to do this?

HasanAlrimawi:
@magurotuna Thank you for the informative feedback.
- I made a change to display the parse diagnostic errors (syntax errors) along with the Lint diagnostic errors at the same time rather than just showing the parse diagnostic errors alone when returning them as an Err.
- I also made a [PR](https://github.com/denoland/deno/pull/24332) to address the same issue in deno Repo.

It would be great if you could review the changes

HasanAlrimawi:
Hello @magurotuna,
Just a kind reminder that the changes from this PR can be merged to deno_lint if you deem it appropriate.

magurotuna:
@HasanAlrimawi Sorry for the delay. The approach looks great, but I think it would be even nicer if we could display syntax errors in the same way as lint errors. For instance, the target file's content is let's say:

```ts
let a = 'fo
```

and when we run dlint against this file, the result is displayed as follows:

dlint output

So as shown in this screenshot, syntax errors are displayed with no color while lint errors are colored. It'd be awesome to add color to syntax errors too. (I guess what you need to do to achieve it is using `deno_ast::diagnostics::Diagnostic::display` of `ParsedSource`)

That said, given that the purpose of `dlint` stand-alone executable is mostly to demonstrate the linter's capability independently, displaying syntax errors in a different way than lint errors is acceptable in my opinion. So if you are interested on dealing with it, please go ahead. But otherwise I will just approve and land it as is. Which would you like?

HasanAlrimawi:
Hello @magurotuna ,
I updated the code to show parsing errors using `deno_ast::diagnostics::Diagnostic::display`.
If everything looks good, I believe it can be merged then.

magurotuna (approved):
Thank you so much, this looks great!

lint output

Issue #423, opened 2020-10-20
Globals discrepancies


jsejcksn:
#419 was just fixed by #421, adding `AggregateError` to the globals list: https://github.com/denoland/deno_lint/blob/master/src/globals.rs

However, I noticed that there are some other inconsistencies between that list and what is produced by enumerating the list in Deno:

```ts
const caseInsensitiveSort = (a: string, b: string): number => {
  const lowerA = a.toLowerCase();
  const lowerB = b.toLowerCase();
  if (lowerA < lowerB) return -1;
  if (lowerA > lowerB) return 1;
  return 0;
};

const globals = [...Object.getOwnPropertyNames(globalThis)];

console.log(globals.sort(caseInsensitiveSort).join('\n'));
```

<details>
<summary>Which produces this list in Deno <code>1.4.6</code>:</summary>

```
AbortController
AbortSignal
addEventListener
AggregateError
Array
ArrayBuffer
atob
Atomics
BigInt
BigInt64Array
BigUint64Array
Blob
Boolean
btoa
ByteLengthQueuingStrategy
clearInterval
clearTimeout
close
closed
CloseEvent
console
CountQueuingStrategy
crypto
CustomEvent
DataView
Date
decodeURI
decodeURIComponent
Deno
dispatchEvent
DOMException
encodeURI
encodeURIComponent
Error
ErrorEvent
escape
eval
EvalError
Event
EventTarget
fetch
File
FileReader
FinalizationRegistry
Float32Array
Float64Array
FormData
Function
globalThis
Headers
Infinity
Int16Array
Int32Array
Int8Array
isFinite
isNaN
JSON
Map
Math
MessageEvent
NaN
Number
Object
onload
onunload
parseFloat
parseInt
Performance
performance
PerformanceEntry
PerformanceMark
PerformanceMeasure
ProgressEvent
Promise
Proxy
queueMicrotask
RangeError
ReadableStream
ReferenceError
Reflect
RegExp
removeEventListener
Request
Response
self
Set
setInterval
setTimeout
SharedArrayBuffer
String
Symbol
SyntaxError
TextDecoder
TextEncoder
TransformStream
TypeError
Uint16Array
Uint32Array
Uint8Array
Uint8ClampedArray
undefined
unescape
URIError
URL
URLSearchParams
WeakMap
WeakRef
WeakSet
WebAssembly
WebSocket
window
Worker
WritableStream
```
</details>

[...]

PR #1289, opened 2024-06-21
fix: update globals list

egfx-notifications:
Ref https://github.com/denoland/deno_lint/issues/423

Not sure I got 100% of the changes, but at least closer to current Deno version now.


dsherret (approved):
Thanks! LGTM.

I opened https://github.com/denoland/deno/issues/24690 to remind us to merge this for 1.46. I think this could cause a bunch of lint failures and I think it might be better to update this in a minor release.

PR #1294, opened 2024-07-22
refactor: make lint rules not require `'static`

dsherret:
This allows creating lint rules in the CLI that aren't `'static`.


src/context.rs
```diff
@@ -125,11 +125,21 @@ impl<'view> Context<'view> {
...
+
   /// Stores diagnostics that are generated while linting
   pub fn diagnostics(&self) -> &[LintDiagnostic] {
     &self.diagnostics
   }
 
+  /// Parsed source of the program.
+  pub fn parsed_source(&self) -> &ParsedSource {
+    &self.parsed_source
+  }
```
dsherret, line 141: I also want this for the CLI so I can pass it to deno graph.

bartlomieju (approved):

PR #1295, opened 2024-07-22
feat: ability for rules to provide a hash of their internal state

dsherret:


src/rules.rs
```diff
@@ -139,6 +139,15 @@ pub trait LintRule: std::fmt::Debug + Send + Sync {
...
   }
+
+  /// A hash of the internal state of the lint rule. This should be a hash
+  /// of what this lint rule depends on.
+  ///
+  /// For example, if the lint rules depends on the config file, then this
+  /// might be a hash of the config file.
+  fn state_hash(&self) -> u64 {
+    0
+  }
```
dsherret, line 150: Actually, I'm going to put this in the CLI.

PR #1296, opened 2024-07-23
refactor: remove `LinterBuilder`

dsherret:
I want to add more options here and I think the builder pattern is harder to ensure everything is setup correctly.


bartlomieju (approved):
LGTM

PR #1297, opened 2024-07-24
feat: ability to provide a custom docs url

dsherret:
This is so we can link to JSR docs for certain lint rules in the CLI.


bartlomieju (approved):
LGTM, though I'm not sure about changing the `code`. Well probably need to revert that once we add plugins

dsherret:
I reverted the code change.

PR #1299, opened 2024-07-24
feat: support diagnostics without ranges

dsherret:
No lint rules in deno lint use this atm, but the CLI does.

### DIFF
--- a/src/diagnostic.rs
+++ b/src/diagnostic.rs
@@ -27,10 +27,15 @@ pub struct LintFix {
 }
 
 #[derive(Clone)]
-pub struct LintDiagnostic {
-  pub specifier: ModuleSpecifier,
-  pub range: SourceRange,
+pub struct LintDiagnosticRange {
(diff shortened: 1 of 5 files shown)


src/diagnostic.rs
```diff
@@ -44,67 +49,85 @@ pub struct LintDiagnostic {
...
+#[derive(Clone)]
+pub struct LintDiagnostic {
+  pub specifier: ModuleSpecifier,
+  /// Optional range within the file.
+  ///
+  /// Diagnostics that don't have a range mean there's something wrong with
+  /// the whole file.
+  pub range: Option<LintDiagnosticRange>,
+  pub details: LintDiagnosticDetails,
+}
```
dsherret, line 65: Split this up so a `context.add_diagnostic_details(...)` could be added that doesn't require providing a specifier.

bartlomieju (approved):
LGTM

PR #1305, opened 2024-08-05
fix(ban-types): Use `PropertyKey` type global in hint message

Malix-Labs:
Maybe a lsp "quick fix" would be nice to add too

### DIFF
--- a/src/rules/ban_types.rs
+++ b/src/rules/ban_types.rs
@@ -49,7 +49,7 @@ impl BannedType {
         r#"If you want a type meaning "any object", use `object` instead. Or if you want a type meaning "any value", you probably want `unknown` instead."#
       }
       EmptyObjectLiteral => {
-        r#"If you want a type that means "empty object", use `Record<string | number | symbol, never>` instead"#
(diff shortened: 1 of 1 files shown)


src/rules/ban_types.rs
```diff
@@ -49,7 +49,7 @@ impl BannedType {
         r#"If you want a type meaning "any object", use `object` instead. Or if you want a type meaning "any value", you probably want `unknown` instead."#
       }
       EmptyObjectLiteral => {
-        r#"If you want a type that means "empty object", use `Record<string | number | symbol, never>` instead"#
+        r#"If you want a type that means "empty object", use `Record<PropertyKey, never>` instead"#
```
marvinhagemeister, line 52: What type is `PropertyKey` referring to? In the previous message users could copy & paste the type from the message. Would this still be possible?

marvinhagemeister: Ohh nvm, just learned that this is a global in TS. Didn't know about this!

marvinhagemeister (approved):
LGTM

PR #1307, opened 2024-08-13
fix(prefer-primordials): Force to use null `[[Prototype]]` object

### LINKED ISSUE #1113: prefer-primordials: Add check for applying null `[[Prototype]]` objects to `{Object,Reflect}.defineProperty`
```js
// prototype pollution
Object.prototype.configurable = true;

const foo = Object.defineProperty({}, "foo", { value: 123 });

// { value: 123, writable: false, enumerable: false, configurable: true }
console.log(Object.getOwnPropertyDescriptor(foo, "foo"));
```

```js
// OK
Object.defineProperty({}, "foo", { __proto__: null, value: 123 });
```

How about checking for `__proto__: null` if the third argument is object literal?

petamoriken:
Fixes #1113
Refs https://github.com/denoland/deno/pull/25040

### DIFF
--- a/src/rules/prefer_primordials.rs
+++ b/src/rules/prefer_primordials.rs
@@ -21,6 +21,8 @@ enum PreferPrimordialsMessage {
   GlobalIntrinsic,
   #[display(fmt = "Don't use the unsafe intrinsic")]
   UnsafeIntrinsic,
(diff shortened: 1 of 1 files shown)


petamoriken:
@magurotuna Please take a look in your time!

magurotuna (approved):
LGTM 👍

PR #1308, opened 2024-08-13
fix(prefer-primordials): allow private identifier in `in` expression

0f-0b:


petamoriken:
I suggested banning the `in` operator to enforce proper use of `Reflect.has` and `Object.hasOwn`, but in most cases `Object.hasOwn` was not used in Deno internal. It might be better to simply allow the `in` operator for DX and performance 🙇

petamoriken:
@lucacasonato What do you think?

bartlomieju:
Thanks — this looks correct. I reproduced the false positive on `main`: `prefer-primordials` flags `#x in o` and suggests `ObjectHasOwn`/`ReflectHas`, but neither can express a private-field brand check, so `#x in o` is the only valid form. Since this rule runs on internal runtime code, the fix matters. Could you rebase on latest `main` so we can get CI green, and confirm the test covers the `#field in obj` case? Then this should be good to merge.

bartlomieju:
**Verdict: LGTM — correct fix. Recommend a maintainer merge** (rule-behavior change + merge state is `BLOCKED`, so not auto-merging).

**Correctness — good**
- The fix correctly exempts the ergonomic brand-check syntax `#brand in obj` from the `in`→`ReflectHas` suggestion. When the LHS of `in` is a `PrivateName`, the expression is *not* rewritable with a primordial (`ReflectHas(obj, #brand)` is invalid syntax), so suppressing the diagnostic is right.
- `BinaryOp::In if !bin_expr.left.is::<PrivateName>()` is the correct guard — `#x in y` is the only construct where a `PrivateName` appears as a binary LHS, so there's no risk of over-suppressing legitimate `in` uses. `instanceof` handling is unchanged.
- The if/else→`match` refactor is clean and equivalent.

**Test coverage — good**
- Adds both a valid case (the `static is()` brand check) and an invalid case (`a in A`) confirming ordinary `in` is still flagged. Solid.

**Performance:** none — same single `bin_expr` visit, one extra `is::<PrivateName>()` type check.

**Housekeeping**
- `mergeable: MERGEABLE` but `mergeStateStatus: BLOCKED` — likely just awaiting a required review/approval rather than a code problem. A maintainer should confirm CI is green and approve; the change itself is ready.

_No changes requested. Good to merge once a maintainer approves._

PR #1309, opened 2024-08-13
feat: Add rule disallowing use of NodeJS global objects

nathanwhit:
Lints things like

```ts
const foo = process.env.FOO; // process global
```

With an automated fix to correct the issue

```ts
import process from "node:process";
const foo = process.env.FOO;
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ once_cell = "1.19.0"
 derive_more = { version = "0.99.17", features = ["display"] }
 anyhow = "1.0.79"
 if_chain = "1.0.2"
+phf = { version = "0.11.2", features = ["macros"] }
 
 [dev-dependencies]
 ansi_term = "0.12.1"
--- a/src/rules/no_node_globals.rs
+++ b/src/rules/no_node_globals.rs
@@ -178,7 +178,6 @@ impl Handler for NoNodeGlobalsHandler {
       return;
     }
     if id.ctxt() == ctx.unresolved_ctxt() {
-      eprintln!("unresolved: {}", id.sym());
       self.add_diagnostic(ctx, id.range(), NODE_GLOBALS[&*id.sym()]);
     }
   }
--- a/src/rules/no_node_globals.rs
+++ b/src/rules/no_node_globals.rs
@@ -174,11 +174,11 @@ impl NoNodeGlobalsHandler {
 
 impl Handler for NoNodeGlobalsHandler {
   fn ident(&mut self, id: &ast_view::Ident, ctx: &mut Context) {
-    if !NODE_GLOBALS.contains_key(&*id.sym()) {
+    if !NODE_GLOBALS.contains_key(id.sym()) {
       return;
     }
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -34,6 +34,7 @@ once_cell = "1.19.0"
 derive_more = { version = "0.99.17", features = ["display"] }
 anyhow = "1.0.79"
 if_chain = "1.0.2"
+phf = { version = "0.11.2", features = ["macros"] }
```
dsherret, line 37: Interesting.

src/rules/no_node_globals.rs
```diff
@@ -0,0 +1,308 @@
...
+  }
+}
+
+fn program_code_start(program: Program) -> SourcePos {
+  match program_ref(program) {
+    ast_view::ProgramRef::Module(m) => {
+      m.body.first().map_or(program.start(), |node| node.start())
+    }
+    ast_view::ProgramRef::Script(s) => {
+      s.body.first().map_or(program.start(), |node| node.start())
```
dsherret, line 117: Can you switch these to `.map(...).unwrap_or(...)`? https://github.com/denoland/deno/pull/18212

src/rules/no_node_globals.rs
```diff
@@ -0,0 +1,308 @@
...
+      ],
+      "const e = clearImmediate;": [
+        {
+          col: 10,
+          line: 1,
+          message: MESSAGE,
+          hint: IMPORT_HINT,
+          fix: (
+            IMPORT_FIX_DESC,
+            "import { clearImmediate } from \"node:timers\";\nconst e = clearImmediate;"
```
dsherret, line 280: Can you add a test for when there's a copyright comment at the top of the file and no imports?

src/rules/no_node_globals.rs
```diff
@@ -0,0 +1,308 @@
...
+    if !NODE_GLOBALS.contains_key(id.sym()) {
+      return;
+    }
+    if id.ctxt() == ctx.unresolved_ctxt() {
+      self.add_diagnostic(ctx, id.range(), NODE_GLOBALS[id.sym()]);
+    }
+  }
+
+  fn import_decl(&mut self, imp: &ast_view::ImportDecl, _ctx: &mut Context) {
+    self.most_recent_import_range = Some(imp.range());
```
dsherret, line 186: Nice.

src/rules/no_node_globals.rs
```diff
@@ -0,0 +1,308 @@
...
+  None,
+}
+
+impl FixKind {
+  fn hint(&self) -> &'static str {
+    match self {
+      FixKind::Import { .. } => IMPORT_HINT,
+      FixKind::Replace(_) => REPLACE_HINT,
+    }
+  }
```
dsherret, line 89: Nitpick: Can you go through and add blank lines between the methods? I wish rustfmt would do it.

dsherret (approved):
Great!

PR #1313, opened 2024-08-17
chore: bump derive_more to v1.0

jayvdb:


bartlomieju:
Closing as stale. The tree is still on derive_more 0.99 and this 51-file v1.0 bump from 2024 will conflict heavily with everything merged since. derive_more 0.99 works fine, so this is low-value churn at this point — better redone fresh if/when we decide to upgrade. Thanks regardless!

PR #1319, opened 2024-09-03
feat: add no-legacy-type-assertion rule

lucacasonato:


dsherret (approved):
LGTM!

bartlomieju:
**Verdict: LGTM on the code — needs maintainer sign-off** (new rule added to `recommended`) + a rebase. Not auto-merging (new rule/behavior).

**Correctness — solid**
- The autofix wraps the result as `(expr as typ)`, and the always-on outer parens make it precedence-safe. Verified the tricky cases hold:
  - `5 + <number> 5` → `5 + (5 as number)` (node range covers only the assertion) ✓
  - `<number> a + b` parses as `(<number> a) + b`, so fix yields `(a as number) + b` ✓
  - `<Foo> bar.baz` → `(bar.baz as Foo)` ✓
- The media-type guard is correct: only `TypeScript | Mts | Cts`. `<Type>` assertions are ambiguous with JSX in `.tsx` and don't occur in `.d.ts`/JS, so skipping those is right.

**Minor**
- Comments inside the assertion are dropped by the fix (e.g. `<number /* x */> 5` → `(5 as number)`), since it reconstructs from `expr`/`type_ann` text. Rare and acceptable, but worth a note.
- The fix always parenthesizes even when unnecessary (`<number> 1` → `(1 as number)`). Safe; `deno fmt` won't strip the parens though, so the fixed output is slightly noisier than hand-written. Fine as a conservative choice.

**Test coverage gaps**
- Add a negative test proving the rule does **not** fire in `.tsx`/`.js`/`.d.ts` (the media-type guard is load-bearing and currently untested).
- Consider a member-expression case (`<Foo> bar.baz`) to lock in the precedence behavior.

**Perf:** none — visits only `ts_type_assertion` nodes; `text_fast` is cheap.

**Housekeeping**
- Adds to `recommended`, so it will fire across existing Deno TS using `<Type>` casts — that's the maintainer decision to confirm.
- Edits generated `www/static/docs.json`; CI has a `diff <(dlint rules --json) docs.json` check, so the hand-edited entry must match generated output exactly after rebase, or CI fails.
- `mergeable: UNKNOWN` / no CI → stale; needs a rebase.

_No logic changes requested; blockers are the `recommended` sign-off + rebase + the media-type negative test._

Issue #1322, opened 2024-09-12
Feature: add rust-style-naming rule


c-antin:
In my deno projects, I often use Rust via ffi and I was wondering if we could unify the naming conventions.

It may have been proposed already (Discord?), yet I didn't find an issue for the following proposal:
Similar to https://rust-lang.github.io/api-guidelines/naming.html a lint rule that enforces `UpperCamelCase` for types, interfaces and classes and `snake_case` for functions, methods and variables plus `SCREAMING_SNAKE_CASE` for constants (if enforcible) would be great.

pros: uniform naming across Rust and TypeScript
cons: JS legacy naming convention is `camelCase`

c-antin:
I'm working on a PR

c-antin:
See #1323

c-antin:
FYI: I've ported the rule to a TS plugin: https://jsr.io/@c-antin/lint-plugin-rust-style

bartlomieju:
See https://github.com/denoland/deno_lint/pull/1323#issuecomment-4826699467

PR #1323, opened 2024-09-15
feat(rust_style): Rust-style naming convention rule

c-antin:
This rule is useful for deno projects that call rust functions via FFI. It attempts to unify naming conventions and enforces declarations and object property names which one creates to be\
in UpperCamelCase/PascalCase for classes, types, interfaces,\
in snake_case for functions, methods, variables\
and in SCREAMING_SNAKE_CASE for static class properties and constants.

See #1322.


src/rules/rust_style.rs
```diff
@@ -0,0 +1,1922 @@
...
+                  IdentToCheck::object_pat(
+                    &key.string_repr().unwrap_or_else(|| "[KEY]".to_string()),
+                    Some(&value_ident.id.inner),
+                    true,
+                    pat_in_var_declarator(pat.into()),
+                  ),
+                );
+              }
+              _ => {
+                self.check_pat(value, ctx);
```
c-antin, line 600: ?is `ctx` valid everywhere, because it is passed to nested entities, see todo line 538?

src/rules/rust_style.rs
```diff
@@ -0,0 +1,1922 @@
...
+      ("foo_bar", false),         //snake
+      ("fooBar", false),          //camel
+      ("FooBar", true),           //upper camel
+      ("foo_bar_baz", false),     //snake
+      ("_foo_bar_baz", false),    //snake
+      ("__foo_bar_baz__", false), //snake
+      ("__fooBar_baz__", false),  //not snake
+      ("__fooBarBaz__", false),   //camel
+      ("Sha3_224", false),        //not snake
+      ("SHA3_224", true),         //screaming snake; todo: ?should this be true?
```
c-antin, line 1022: ?according to camelcase.rs `SHA3_224` is valid UpperCamelCase?

src/rules/rust_style.rs
```diff
@@ -0,0 +1,1922 @@
+// Copyright 2018-2024 the Deno authors + c-antin. All rights reserved. MIT license.
+// based on camelcase.rs
```
c-antin, line 2: compare first commit with last commit for useful diff, because `rust_style.rs` is based on `camelcase.rs`

docs/rules/rust_style.md
```diff
@@ -0,0 +1,89 @@
...
+### Valid:
+
+```typescript
+let first_name = "Ichigo";
+const FIRST_NAME = "Ichigo";
+const __my_private_variable = "Hoshimiya";
+const my_private_variable_ = "Hoshimiya";
+const obj1 = { "lastName": "Hoshimiya" }; // if an object key is wrapped in quotation mark, then it's valid
+const obj2 = { "firstName": firstName };
+const { lastName } = obj1; //valid, because one has no control over the identifier
```
c-antin, line 67: note: #1187 changed this behavior, but camelcase.md did not get updated

docs/rules/rust_style.md
```diff
@@ -0,0 +1,89 @@
...
+const { lastName: last_name } = obj;
+
+function do_something() {} // function declarations must be snake_case but...
+doSomething(); // ...camel_case function calls are allowed
+function foo({ camelCase: snake_case = "default value" }) {}
+
+class PascalCaseClass {}
+
+import { camelCased } from "external-module.js"; //valid, because one has no control over the identifier
```
c-antin, line 76: note: #1187 changed this behavior, but camelcase.md did not get updated

c-antin: See #1325

docs/rules/rust_style.md
c-antin: See #1325

src/rules/rust_style.rs
```diff
@@ -0,0 +1,2070 @@
...
+}
+
+impl Handler for RustStyleHandler {
+  fn assign_expr(&mut self, e: &AssignExpr, _ctx: &mut Context) {
+    let idents: Vec<deno_ast::swc::ast::Ident> = find_lhs_ids(&e.left);
+
+    for ident in idents {
+      self.check_ident_snake_cased_or_screaming_snake_cased(
+        &ident.range(),
+        IdentToCheck::variable_or_constant(ident.to_id().0),
```
c-antin, line 654: ?should global assign of camelCase variables be valid, because one has no control over the identifier?
if so, we should remove `fn assign_expr`.

bartlomieju:
Thanks for the PR and the thorough work here! We've decided not to take this into core — enforcing Rust-style naming (snake_case functions, etc.) is quite niche for a general JS/TS linter and overlaps/conflicts with the existing `camelcase` rule. It's a great fit as a lint plugin though, and your published plugin is perfect for anyone who wants it: https://jsr.io/@c-antin/lint-plugin-rust-style — we'd happily point users there. Closing in favor of that.

Issue #1324, opened 2024-09-15
camelcase docs don't reflect #1187


c-antin:
### Lint Name

camelcase

### Code Snippet

```ts
const { lastName } = obj1; //valid, because one has no control over the identifier
import { camelCased } from "external-module.js"; //valid, because one has no control over the identifier
```

c-antin:
See PR #1325

PR #1325, opened 2024-09-15
fix(camelcase): updated docs, see #1187

c-antin:
See #1324


bartlomieju:
**Verdict: LGTM — docs are accurate. Recommend merge after a rebase** (branch is conflicting). Docs-only, no code.

**Correctness — verified against the implementation**
Both moved examples are genuinely treated as *valid* by `src/rules/camelcase.rs`, so relabeling them Invalid→Valid is correct:
- `const { last_name } = obj1;` — in `ObjectPatProp::Assign`, the check is gated by `if !in_var_declarator { ... }`, so a shorthand destructuring binding inside a `var`/`const` declarator is intentionally not flagged (you can't rename the source key without `: alias`). ✓
- `import { not_camelCased } from "..."` — `import_named_specifier` only checks when `imported` is `Some` (i.e. an `as` rename is present). A plain named import has `imported: None` and is skipped. ✓

The added inline comments ("valid, because one has no control over the identifier") match the real rationale.

**Perf:** n/a.

**Housekeeping / blockers**
- `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` → needs a rebase before it can land.
- It also hand-edits the generated `www/static/docs.json`; CI runs `diff <(dlint rules --json) docs.json`, so after rebase the `docs.json` entry must exactly match what `--features=docs` regenerates from the `.md` (better to regenerate than hand-edit).
- Note: canonical rule docs may now live in the `denoland/docs` repo rather than `docs/rules/*.md` here — worth confirming this fix is applied wherever the site actually sources these docs, or it may not surface on the website.

_No content changes requested; just rebase + confirm the docs source-of-truth, then it's good._

PR #1328, opened 2024-09-17
fix: remove process global check

marvinhagemeister:
Deno 2 will ship with the `process` global being available by default.

See https://github.com/denoland/deno/issues/25679

### DIFF
--- a/docs/rules/no_node_globals.md
+++ b/docs/rules/no_node_globals.md
@@ -8,14 +8,14 @@ their defining modules as needed.
 
 ```typescript
 // foo.ts
-const foo = process.env.FOO; // process is not a global object in deno
+const buf = Buffer.from("foo", "utf-8"); // Buffer is not a global object in deno
 ```
 
 ### Valid:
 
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):

lucacasonato:
I think rather than removing, we should keep the lint rule and just change the message? I think it's still preferable to be explicit about the fact that you want the process global from `node:process`. It makes it much easier for tooling also to tell when something will/will not work in the browser.

dsherret:
Yeah we should keep this rule.

marvinhagemeister:
I'm not sure. We _do_ support the `process` global, so it shouldn't be an error to use it. Otherwise we wouldn't have added support for it. Do we want to encourage folks to move to explicit imports? Maybe, but then we shouldn't throw errors imo.

marvinhagemeister:
The problem is also that we don't have more granular configuration of our linter. We cannot say, error on anything but the `process` global or flag these are warnings or something.

PR #1335, opened 2024-10-22
chore: add schemas for rules and tags

nayeemrmn:
Towards https://github.com/denoland/deno/issues/26314.

We will link to these at https://github.com/denoland/deno/blob/v2.0.2/cli/schemas/config-file.v1.json#L289-L317 using `"$ref": "https://raw.githubusercontent.com/..."`. Similar to [how the extension links to this file](https://github.com/denoland/vscode_deno/blob/3.42.0/package.json#L596) in the first place.

I included the two rules implemented in CLI in the schema.

If there's some automated CI checks we should do for these files, please advise.


bartlomieju (commented):
I think you can generate this file in similar fashion that documentation site is generated. Take a look at `update-docs` task - preferably just update that task to take care of generating schema file.

dsherret (approved):
LGTM

bartlomieju (approved):
LGTM too

PR #1338, opened 2024-10-29
refactor: adjust rule code for `no-process-global` to ensure consistency

emmanuel-ferdman:
# PR Summary
The PR updates the code of the  [`no-process-global`](https://github.com/denoland/deno_lint/blob/main/src/rules/no_process_global.rs) rule to be `no-process-global` instead of `no-process-globals` in order to be consistent as other rules (code and name are the same). Other than confusion, it solves issues like accessing the rule source code in the [website](https://lint.deno.land/rules/no-process-globals) which currently leads to broken page (press on the [Rule source](https://github.com/denoland/deno_lint/blob/main/src/rules/no_process_globals.rs) button in the rule [page](https://lint.deno.land/rules/no-process-globals)).

### DIFF
--- a/schemas/rules.v1.json
+++ b/schemas/rules.v1.json
@@ -70,7 +70,7 @@
     "no-non-null-assertion",
     "no-obj-calls",
     "no-octal",
-    "no-process-globals",
+    "no-process-global",
     "no-prototype-builtins",
     "no-redeclare",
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
Thanks, good catch.

PR #1340, opened 2024-11-07
fix(no-control-regex): Only warn on control char literals

### LINKED ISSUE #1143: Rule suggestion: remove `no-control-regex` from defaults
The regex `/[\x00-\x1f\x7f]/` gives this warning:

> Unexpected control character(s) in regular expression: \x0.
> Disable the rule if the control character (\x... or \u00..) was intentional, otherwise rework your RegExp
> deno-lint(no-control-regex)

Aside from the fact the warning is wrong (it's `\x00`, not `\x0`; also it doesn't trigger if `\x7f` is the only control char in the regex), regexes such as `/[\x00-\x1f\x7f]/` are a somewhat common occurrence and a completely reasonable thing to want to write, particularly if you're doing something like parsing raw output from `stdin`/`stdout`.

Alternatively: change the rule to only trigger for control character _literals_, which impede readability due to rendering as tofu in most fonts, and are much more likely to be there eroneously, whether in regexes or elsewhere.

lionel-rowe:
Fixes https://github.com/denoland/deno_lint/issues/1143.

Old and new behavior both diverge from [ESLint](https://eslint.org/docs/latest/rules/no-control-regex) but in different ways:
* ESLint detects both control char escapes and control char literals
* Old deno_lint behavior only detected control char escapes but failed to detect control char literals
* New deno_lint behavior only detects control char literals and intentionally overlooks control char escapes (reasoning per the linked issue)

In all cases (including ESLint), the checking is done against the regex source string rather than the JS/TS source code. This gives the following behavior:

JS/TS source code | ESLint | deno_lint<br>old | deno_lint<br>new
-|:-:|:-:|:-:
`//` (literal 0x1f char) | ⚠️ | 🆗 | ⚠️
`/\x1f/` | ⚠️ | ⚠️ | 🆗
`new RegExp('')` (literal 0x1f char) | ⚠️ | 🆗 | ⚠️
`new RegExp('\x1f')` | ⚠️ | 🆗 | ⚠️
`new RegExp('\\x1f')` | ⚠️ | ⚠️ | 🆗

Also fixes the following 2 bugs noted in the linked issue:
* Formatting of the message fixed such that code point `0x00` is rendered `\x00` not `\x0`
* Added \x7f DEL to the detected control chars

### DIFF
--- a/src/rules/no_control_regex.rs
+++ b/src/rules/no_control_regex.rs
@@ -7,8 +7,6 @@ use crate::Program;
 use deno_ast::view::{CallExpr, Callee, Expr, NewExpr, Regex};
 use deno_ast::{SourceRange, SourceRanged};
 use derive_more::Display;
-use std::iter::Peekable;
-use std::str::Chars;
 
 #[derive(Debug)]
 pub struct NoControlRegex;
@@ -18,7 +16,7 @@ const CODE: &str = "no-control-regex";
 #[derive(Display)]
 enum NoControlRegexMessage {
   #[display(
-    fmt = "Unexpected control character(s) in regular expression: \\x{:x}.",
+    fmt = "Unexpected control character literal in regular expression: \\x{:0>2x}.",
     _0
   )]
   Unexpected(u64),
@@ -27,7 +25,7 @@ enum NoControlRegexMessage {
 #[derive(Display)]
 enum NoControlRegexHint {
   #[display(
-    fmt = "Disable the rule if the control character (\\x... or \\u00..) was intentional, otherwise rework your RegExp"
+    fmt = "Disable the rule if the control character literal was intentional, otherwise rework your RegExp"
   )]
   DisableOrRework,
 }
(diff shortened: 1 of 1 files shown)


bartlomieju:
**Verdict: LGTM on the implementation — needs maintainer sign-off** on the semantic change (it alters a `recommended` rule). Not auto-merging (behavior change).

I traced the data flow to confirm the new logic is internally consistent:
- `regex()` passes `regex.inner.exp` (raw regex *pattern source*), so `/\x1f/` is seen as the 4 chars `\ x 1 f` → no control char → no warning.
- `new_expr`/`call_expr` pass `extract_regex(...)`, which returns the **cooked** `Str.value`. So `new RegExp('\x1f')` (value = 0x1f) warns, while `new RegExp('\\x1f')` (value = `\x1f`) doesn't.
This precisely matches the behavior table in the PR description, and every new valid/invalid test case is consistent with `is_control_char(cp) = cp <= 0x1f || cp == 0x7f`. ✓

**Correctness — good**
- The message zero-pad fix (`\x{:0>2x}`) correctly renders `0x00` as `\x00`; the added `no_control_regex_message` test locks it in. Since `is_control_char` only matches ≤ 2 hex digits, the width-2 format is always sufficient.
- Adding `0x7f` (DEL) matches ESLint and is covered by a test.
- Literal `\t`/`\n`/`\r` (0x09/0x0a/0x0d) are intentionally included; tests distinguish `r"/\t/"` (escape → valid) from `"/\t/"` (literal tab → invalid). Correct.

**Performance — improved**
- Net simplification: removes the hand-rolled hex/peekable parsing (~119 lines) for a single `chars()` pass. Slightly faster and less allocation. Nice cleanup.

**The decision for a maintainer**
- This *inverts* what the rule detects: it now catches literal control chars (which ESLint catches and old deno_lint missed) but **intentionally stops** flagging escape sequences like `/\x1f/` and `/[\x00-\x1f]/` (which both ESLint and old deno_lint flag). So relative to ESLint it fixes one gap and opens another. The rationale is in #1143, but since this is a `recommended` rule, previously-warning code will silently stop warning — that's the trade-off to approve.

**Housekeeping:** `mergeable: UNKNOWN` / no CI → stale; needs a rebase.

_No code changes requested — the implementation is solid and well-tested. Blocker is a maintainer decision on the escape-vs-literal semantics + a rebase._

PR #1345, opened 2024-11-21
renamed no-process-global

DanMyers300:
no-process-global shows up as no-process-globals when running deno lint or using the deno lsp in neovim.
It also includes a link that doesn't work.

Screenshot from 2024-11-21 09-28-13
image


dsherret:
This really confused me. Seems like it was renamed here: https://github.com/denoland/deno_lint/pull/1338/files

I think probably it should be `no-process-global`, but at the same time, it's a breaking change (cc @bartlomieju)

DanMyers300:
> This really confused me. Seems like it was renamed here: https://github.com/denoland/deno_lint/pull/1338/files
> 
> I think probably it should be `no-process-global`, but at the same time, it's a breaking change (cc @bartlomieju)

Yeah, I’m not sure I actually fixed it with this change. My main issue is that the documentation doesn’t match the rule.

PR #1346, opened 2024-11-22
feat: add rules_of_hooks rule

marvinhagemeister:
This PR implements the `rules-of-hooks` eslint rule, see https://react.dev/reference/rules/rules-of-hooks . It ensures that hooks are only used inside components and that they are called unconditionally.


docs/rules/rules_of_hooks.md
```diff
@@ -0,0 +1,74 @@
+Ensure that (P)React hooks are only called inside component functions and not
```
bartlomieju, line 1:
```suggestion
Ensure that React/Preact hooks are only called inside component functions and not
```

docs/rules/rules_of_hooks.md
```diff
@@ -0,0 +1,74 @@
...
+called conditionally. A hook is a function that starts with `use`
```
bartlomieju, line 2:
```suggestion
called conditionally. A hook is a function that starts with `use` prefix.
```

docs/rules/rules_of_hooks.md
```diff
@@ -0,0 +1,74 @@
...
+```tsx
+function Component() {
+  const [count, setCount] = useState(0);
+  // ...
+}
+
+function useCustomHook() {
+  const [count, setCount] = useState(0);
+  // ...
+}
```
bartlomieju, line 73: Could you update this one to show all "bad" examples written in a correct form?

src/rules/rules_of_hooks.rs
```diff
@@ -0,0 +1,378 @@
...
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct RulesOfHooks;
+
+const CODE: &str = "rules-of-hooks";
+
+impl LintRule for RulesOfHooks {
+  fn tags(&self) -> &'static [&'static str] {
+    &["react", "jsx", "fresh"]
```
bartlomieju, line 16:
```suggestion
    &["recommended", "react", "jsx", "fresh"]
```

src/rules/rules_of_hooks.rs
```diff
@@ -0,0 +1,378 @@
...
+}
+
+fn is_hook_or_component_name(name: &str) -> bool {
+  if name.starts_with("use") {
+    return true;
+  }
+
+  if let Some(ch) = name.chars().next() {
+    return ch.is_uppercase();
+  }
```
bartlomieju, line 276: Where is this hit? Could you add a small unit test for this function?

docs/rules/rules_of_hooks.md
marvinhagemeister: These are the only correct ones for all bad examples.

src/rules/rules_of_hooks.rs
marvinhagemeister: This is hit in pretty much all bad variants. It tests if the enclosing function is a component.

bartlomieju (approved):

PR #1348, opened 2024-11-22
feat: add jsx-no-unescaped-entities rule

marvinhagemeister:
This rule ensure that you're not accidentally passing entities like `>` as text when you didn't intend to.

### DIFF
--- a/src/rules/jsx_no_unescaped_entities.rs
+++ b/src/rules/jsx_no_unescaped_entities.rs
@@ -1,0 +1,107 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+use super::{Context, LintRule};
+use crate::diagnostic::{LintFix, LintFixChange};
+use crate::handler::{Handler, Traverse};
+use crate::tags::{self, Tags};
+use crate::Program;
+use deno_ast::view::{JSXElement, JSXElementChild};
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct JSXNoUnescapedEntities;
+
+const CODE: &str = "jsx-no-unescaped-entities";
+
+impl LintRule for JSXNoUnescapedEntities {
+  fn tags(&self) -> Tags {
+    &[tags::RECOMMENDED, tags::REACT, tags::JSX, tags::FRESH]
+  }
+
+  fn code(&self) -> &'static str {
+    CODE
+  }
+
+  fn lint_program_with_ast_view(
+    &self,
+    context: &mut Context,
+    program: Program,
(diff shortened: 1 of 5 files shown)


src/rules/jsx_no_unescaped_entities.rs
```diff
@@ -0,0 +1,106 @@
...
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct JSXNoUnescapedEntities;
+
+const CODE: &str = "jsx-no-unescaped-entities";
+
+impl LintRule for JSXNoUnescapedEntities {
+  fn tags(&self) -> &'static [&'static str] {
+    &["react", "jsx"]
```
bartlomieju, line 17:
```suggestion
    &["recommended", "react", "jsx"]
```

src/rules/jsx_no_unescaped_entities.rs
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  #[cfg(feature = "docs")]
+  fn docs(&self) -> &'static str {
+    include_str!("../../docs/rules/jsx_no_unescaped_entities.md")
+  }
+}
+
+const MESSAGE: &str = "Found one or more unescaped entities in JSX text";
+const HINT: &str = "Escape the '\">} characters respectively";
```
bartlomieju, line 39: This is confusing - is `'` character supposed to be escaped? If so it's not listed in the list below

src/rules/jsx_no_unescaped_entities.rs
```diff
@@ -0,0 +1,106 @@
...
+      r#"<div>{">"}</div>"#,
+    };
+  }
+
+  #[test]
+  fn jsx_no_unescaped_entities_invalid() {
+    assert_lint_err! {
+      JSXNoUnescapedEntities,
+      filename: "file:///foo.jsx",
+      r#"<div>'">}</div>"#: [
```
bartlomieju, line 93: Please exercise all possible characters here

src/rules/jsx_no_unescaped_entities.rs
marvinhagemeister: It's listed here https://github.com/denoland/deno_lint/blob/a2a2bea499fa44d26ad29846866eca039741223d/src/rules/jsx_no_unescaped_entities.rs#L52

src/rules/jsx_no_unescaped_entities.rs
marvinhagemeister: They're all listed

bartlomieju (approved):

PR #1349, opened 2024-11-22
feat: add jsx-no-danger-with-children rule

marvinhagemeister:
This rule ensures that you're not using `dangerouslySetInnnerHTML` together with passing children to a JSX node. When `dangerouslySetInnerHTML` is used, any `children` are always ignored.


src/rules/jsx_no_danger_with_children.rs
```diff
@@ -0,0 +1,87 @@
...
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct JSXNoDangerWithChildren;
+
+const CODE: &str = "jsx-no-danger-with-children";
+
+impl LintRule for JSXNoDangerWithChildren {
+  fn tags(&self) -> &'static [&'static str] {
+    &["react", "jsx", "fresh"]
```
bartlomieju, line 16:
```suggestion
    &["recommended", "react", "jsx", "fresh"]
```

src/rules/jsx_no_danger_with_children.rs
```diff
@@ -0,0 +1,87 @@
...
+            ctx.add_diagnostic_with_hint(node.range(), CODE, MESSAGE, HINT);
+          }
+        }
+      }
+    }
+  }
+}
+
+// most tests are taken from ESlint, commenting those
+// requiring code path support
```
bartlomieju, line 59: What's code path?

bartlomieju (approved):

src/rules/jsx_no_danger_with_children.rs
marvinhagemeister: Lol I just copied the tests from another existing lint rule we had and looks like I copied that code comment everywhere :joy:

PR #1350, opened 2024-11-22
feat: add button_has_type rule

marvinhagemeister:
By default a `<button>` element has `type="submit"` which often is not the desired behaviour. This rule ensures that the `type` attribute is set explicitly.


src/rules/button_has_type.rs
```diff
@@ -0,0 +1,362 @@
...
+use deno_ast::{view as ast_view, SourceRanged};
+
+#[derive(Debug)]
+pub struct ButtonHasType;
+
+const CODE: &str = "button-has-type";
+
+impl LintRule for ButtonHasType {
+  fn tags(&self) -> &'static [&'static str] {
+    &["react", "jsx", "fresh"]
```
bartlomieju, line 19:
```suggestion
    &["recommended", "react", "jsx", "fresh"]
```

src/rules/button_has_type.rs
```diff
@@ -0,0 +1,362 @@
...
+    (self.message(), self.hint())
+  }
+
+  fn message(&self) -> &'static str {
+    match *self {
+      DiagnosticKind::MissingTypeAttr => {
+        "`button` elements must have a `type` attribute"
+      }
+      DiagnosticKind::MissingValue => {
+        "Missing value for button `type` attribute"
```
bartlomieju, line 58:
```suggestion
        "Missing value for button's `type` attribute"
```

src/rules/button_has_type.rs
```diff
@@ -0,0 +1,362 @@
...
+      }
+      DiagnosticKind::WrongValue => {
+        "Incorrect value for button `type` attribute"
```
bartlomieju, line 61:
```suggestion
        "Incorrect value for button's `type` attribute"
```

src/rules/button_has_type.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+  }
+
+  fn hint(&self) -> &'static str {
+    match *self {
+      DiagnosticKind::MissingTypeAttr => {
+        "Add a `type` attribute with a value of `button`, `submit` or `reset`"
+      }
+      DiagnosticKind::MissingValue => "The value of the `type` attribute must be one of `button`, `submit` or `reset`",
+      DiagnosticKind::WrongValue => "The value of the `type` attribute must be one of `button`, `submit` or `reset`",
```
bartlomieju, line 72:
```suggestion
      DiagnosticKind::MissingValue | DiagnosticKind::WrongValue => "The value of the `type` attribute must be one of `button`, `submit` or `reset`",
```

src/rules/button_has_type.rs
```diff
@@ -0,0 +1,362 @@
...
+                        }
+                      } else {
+                        ctx.add_diagnostic_with_hint(
+                          attr_value.range(),
+                          CODE,
+                          kind.message(),
+                          kind.hint(),
+                        );
+                      }
+                    }
```
bartlomieju, line 116: Ouch, the indentation here 😬 can you please use `let ... else ...` or the `if-chain` crate that we already depend on?

bartlomieju (approved):

PR #1351, opened 2024-11-22
feat: add no_danger rule

marvinhagemeister:
Disallow the usage of `dangerouslySetInnerHTML` which can lead to security vulnerabilities if misused.

### DIFF
--- a/src/rules/no_danger.rs
+++ b/src/rules/no_danger.rs
@@ -1,0 +1,96 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+use super::{Context, LintRule};
+use crate::handler::{Handler, Traverse};
+use crate::Program;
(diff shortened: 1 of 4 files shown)


src/rules/no_danger.rs
```diff
@@ -0,0 +1,96 @@
...
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct NoDanger;
+
+const CODE: &str = "no-danger";
+
+impl LintRule for NoDanger {
+  fn tags(&self) -> &'static [&'static str] {
+    &["react", "jsx"]
```
bartlomieju, line 16: Should it be recommended? Probably not?

bartlomieju (approved):

src/rules/no_danger.rs
marvinhagemeister: Good point! I'll add it to recommended.

PR #1352, opened 2024-11-22
feat: add jsx-no-comment-text-nodes rule

marvinhagemeister:
Passing code comments as JSX text nodes is a common mistake. This Lint rule checks if a code comment was accidentally passed as JSX text by checking if it starts with `//` or `/*`.

### DIFF
--- a/src/rules/jsx_no_comment_text_nodes.rs
+++ b/src/rules/jsx_no_comment_text_nodes.rs
@@ -1,0 +1,88 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+use super::{Context, LintRule};
+use crate::handler::{Handler, Traverse};
+use crate::tags::Tags;
+use crate::{tags, Program};
+use deno_ast::view::JSXText;
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct JSXNoCommentTextNodes;
+
+const CODE: &str = "jsx-no-comment-text-nodes";
+
+impl LintRule for JSXNoCommentTextNodes {
+  fn tags(&self) -> Tags {
+    &[tags::RECOMMENDED, tags::REACT, tags::JSX, tags::FRESH]
+  }
+
+  fn code(&self) -> &'static str {
+    CODE
+  }
+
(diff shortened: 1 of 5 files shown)


src/rules/jsx_no_comment_text_nodes.rs
```diff
@@ -0,0 +1,87 @@
...
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct JSXNoCommentTextNodes;
+
+const CODE: &str = "jsx-no-comment-text-nodes";
+
+impl LintRule for JSXNoCommentTextNodes {
+  fn tags(&self) -> &'static [&'static str] {
+    &["react", "jsx", "fresh"]
```
bartlomieju, line 16:
```suggestion
    &["recommended", "react", "jsx", "fresh"]
```

src/rules/jsx_no_comment_text_nodes.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const MESSAGE: &str =
+  "Comments inside children should be placed inside curly braces";
+
+struct JSXNoCommentTextNodesHandler;
+
+impl Handler for JSXNoCommentTextNodesHandler {
+  fn jsx_text(&mut self, node: &JSXText, ctx: &mut Context) {
+    //
```
bartlomieju, line 44: Remove?

src/rules/jsx_no_comment_text_nodes.rs
```diff
@@ -0,0 +1,87 @@
...
+    let value = &node.inner.value;
+    if value.starts_with("//") || value.starts_with("/*") {
```
bartlomieju, line 46: Should these be stricter checks? Eg. ensuring there are only single lines or for block comments that it ends with `*/`?

marvinhagemeister: This is too difficult to do, because of expression nodes. In practice this is imo good enough.

```jsx
a = <p>/* foo {text}</p>
```

bartlomieju (approved):

PR #1355, opened 2024-11-22
feat: add jsx-key rule

marvinhagemeister:
Ensure that the `key` prop is used when iterating over arrays. The `key` prop is crucial for fast updates in the browser.


src/rules/jsx_key.rs
```diff
@@ -0,0 +1,480 @@
...
+      DiagnosticKind::MissingKey => "Missing 'key' prop for element",
+      DiagnosticKind::MissingFragKey => "Missing 'key' prop for Fragemnt",
+    }
+  }
+
+  fn hint(&self) -> &'static str {
+    match *self {
+      DiagnosticKind::MissingKey => "Add a 'key' prop",
+      DiagnosticKind::MissingFragKey => {
+        "Use the <Fragment key=\"..\"> syntax instead"
```
bartlomieju, line 57:
```suggestion
        "Use the `<Fragment key=\"..\">` syntax instead"
```

src/rules/jsx_key.rs
```diff
@@ -0,0 +1,480 @@
...
+          check_callback(ctx, &callback.expr)
+        }
+      } else if let Expr::Member(member) = callee {
+        if let Expr::Ident(id) = member.obj {
+          if id.sym() == "Array" {
+            if let MemberProp::Ident(member_id) = member.prop {
+              if member_id.sym() == "from" {
+                if let Some(el) = node.args.get(1) {
+                  check_callback(ctx, &el.expr);
+                }
```
bartlomieju, line 106: Consider `let ... else ...` syntax here

bartlomieju (approved):

PR #1356, opened 2024-11-22
feat: add jsx-no-duplicate-props rule

marvinhagemeister:
Ensure that the same prop is not passed twice to a JSX node. This is typically a mistake.

### DIFF
--- a/src/rules/jsx_no_duplicate_props.rs
+++ b/src/rules/jsx_no_duplicate_props.rs
@@ -15,7 +15,7 @@ const CODE: &str = "jsx-no-duplicate-props";
 
 impl LintRule for JSXNoDuplicateProps {
   fn tags(&self) -> &'static [&'static str] {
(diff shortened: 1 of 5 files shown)


src/rules/jsx_no_duplicate_props.rs
```diff
@@ -0,0 +1,117 @@
...
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct JSXNoDuplicateProps;
+
+const CODE: &str = "jsx-no-duplicate-props";
+
+impl LintRule for JSXNoDuplicateProps {
+  fn tags(&self) -> &'static [&'static str] {
+    &["react", "jsx"]
```
bartlomieju, line 18:
```suggestion
    &["recommended", "react", "jsx"]
```

bartlomieju (approved):

PR #1357, opened 2024-11-22
feat: add jsx-no-useless-fragment rule

marvinhagemeister:
Fragments were introduced to allow multiple top level children on a JSX block. They're useless when used with a single child.

### DIFF
--- a/src/rules/jsx_no_useless_fragment.rs
+++ b/src/rules/jsx_no_useless_fragment.rs
@@ -1,0 +1,121 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+use super::{Context, LintRule};
+use crate::handler::{Handler, Traverse};
+use crate::Program;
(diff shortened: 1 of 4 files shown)


src/rules/jsx_no_useless_fragment.rs
```diff
@@ -0,0 +1,119 @@
...
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct JSXNoUselessFragment;
+
+const CODE: &str = "jsx-no-useless-fragment";
+
+impl LintRule for JSXNoUselessFragment {
+  fn tags(&self) -> &'static [&'static str] {
+    &["react", "jsx", "fresh"]
```
bartlomieju, line 16:
```suggestion
    &["recommended", "react", "jsx", "fresh"]
```

bartlomieju (approved):

PR #1359, opened 2024-11-22
feat: add jsx-no-children-prop rule

marvinhagemeister:
Ensure that children are passed as children instead of using the `children` prop.


src/rules/jsx_no_children_prop.rs
```diff
@@ -0,0 +1,100 @@
...
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct JSXNoChildrenProp;
+
+const CODE: &str = "jsx-no-children-prop";
+
+impl LintRule for JSXNoChildrenProp {
+  fn tags(&self) -> &'static [&'static str] {
+    &["react", "jsx", "fresh"]
```
bartlomieju, line 16:
```suggestion
    &["recommended", "react", "jsx", "fresh"]
```

bartlomieju (approved):

Issue #1238, opened 2024-01-28
Lint rule for non-bare specifier


dsherret:
Bad:

```ts
import * as path from "https://deno.land/std/path/mod.ts";
```

Good (uses import map):

```ts
import * as path from "@std/path";
```

Why: I want to ensure my dependenices are defined in the deno.json file instead of accidentally in the code (ex. via an auto-import going crazy) so that I can generate the lockfile for a project when running `deno install` (instead of also having to run/cache scripts with non-bare specifiers).

PR #1361, opened 2024-11-22
feat: add no-import-prefix rule

marvinhagemeister:
Went a bit back and forth on the name, but I feel like `no-import-prefix` might be clearer than `non-bare-import`. Don't have strong feelings about that though.

Fixes https://github.com/denoland/deno_lint/issues/1238


src/rules/no_import_prefix.rs
```diff
@@ -0,0 +1,136 @@
...
+
+const CODE: &str = "no-import-prefix";
+const MESSAGE: &str =
+  "Inline 'npm:*', 'jsr:*' or 'http:*' dependency not allowed";
+const HINT: &str = "Add it to the 'imports' section in 'deno.json' instead";
+
+impl LintRule for NoImportPrefix {
+  fn tags(&self) -> &'static [&'static str] {
+    &[]
+  }
```
bartlomieju, line 20: I'm worrying that ppl will not use this rule because it's not enabled by default. Let's think about a way we could conditionally enable it in `deno` if we discover `deno.json` or `package.json` - in such cases you're not writing a one-off script but have a project and you should centralize dependencies.

dsherret: Right now we have recommended, fresh, jsr, react, and jsx tags. I think we could also add a "workspace" tag for this, then the CLI could enable all "workspace" tag rules when in a workspace (having a deno.json or package.json)

docs/rules/no_import_prefix.md
dsherret, line 1: Moved to https://github.com/denoland/docs/pull/2496

src/rules/no_import_prefix.rs
```diff
@@ -0,0 +1,132 @@
...
+use crate::Program;
+use deno_ast::view::{CallExpr, Callee, Expr, ImportDecl, Lit};
+use deno_ast::SourceRanged;
+
+#[derive(Debug)]
+pub struct NoImportPrefix;
+
+const CODE: &str = "no-import-prefix";
+const MESSAGE: &str =
+  "Inline 'npm:', 'jsr:' or 'http:' dependency not allowed";
```
bartlomieju, line 15:
```suggestion
  "Inline 'npm:', 'jsr:' or 'http(s):' dependency not allowed";
```

src/rules/no_import_prefix.rs
```diff
@@ -0,0 +1,132 @@
...
+const HINT: &str = "Add it as a dependency in a deno.json or package.json instead and reference it here via its bare specifier";
```
bartlomieju, line 16:
```suggestion
const HINT: &str = "Add it as a dependency in a deno.json(c) or package.json instead and reference it here via its bare specifier";
```

dsherret (approved):
LGTM! Exciting.

Issue #1094, opened 2022-10-29
Lint for unversioned npm specifiers


dsherret:
Doing something like `import chalk from "npm:chalk"` is nice for quick hacking, but shouldn't be used in real apps. I think maybe this lint might be best done in the cli though because it involves resolving import maps in some cases.

JakeOShannessy:
Can this also be applied to jsr imports? This would be particularly useful as jsr itself rejects imports without version specifiers.

dsherret:
For sure. This issue was just created before jsr existed.

PR #1362, opened 2024-11-22
feat: add no-unversioned-import rule

marvinhagemeister:
Fixes https://github.com/denoland/deno_lint/issues/1094


src/rules/no_unversioned_import.rs
```diff
@@ -0,0 +1,147 @@
...
+
+fn is_unversioned(s: &str) -> bool {
+  if s.starts_with("npm:") {
+    return NPM_REG.is_match(s);
+  } else if s.starts_with("jsr:") {
+    return JSR_REG.is_match(s);
+  }
+
+  false
+}
```
bartlomieju, line 76: I think this is okay for the first pass, but maybe we should use `deno_semver` to do specifier parsing? @dsherret WDYT?

dsherret: Yeah, we should use deno_semver here.

PR #1373, opened 2024-12-07
feat: add ban-untagged-deprecation rule

### LINKED ISSUE #1266: suggestion: `ban-untagged-deprecation`
Ensures that all [`@deprecated`](https://jsdoc.app/tags-deprecated) tags have an accompanying message. I created https://github.com/denoland/deno_std/pull/4577 to compensate for this in `std`.

Fails:
```ts
/**
 * @deprecated
 */
export const foo = 42;
```

Passes:
```ts
/**
 * @deprecated This will be removed in 1.0.0.
 */
export const foo = 42;
```

takaebato:
Fixes: https://github.com/denoland/deno_lint/issues/1266

### 1. Warning Range
Using the `Comment#range` method  would apply warnings to the entire block comment, which isn’t ideal. So instead, the implementation now focuses on issuing warnings only for the specific target range.

For example, in the following file (example.ts):
```typescript:example.ts
/**
 * @param foo - The input value.
 * @deprecated   
 * @returns The computed result.
 */
export function oldFunction(foo: number): number {
    return foo + 1;
}
```
The warning highlights only the `@deprecated` tag and the trailing whitespace:

```
$ ./target/debug/examples/dlint run example.ts --rule ban-untagged-deprecation
error[ban-untagged-deprecation]: The @deprecated tag must include descriptive text
 --> /Users/taka/Documents/GitHub/deno_lint/example.ts:3:4
  | 
3 |  * @deprecated   
  |    ^^^^^^^^^^^^^^
  = hint: Provide additional context for the @deprecated tag, e.g., '@deprecated since v2.0'

  docs: https://lint.deno.land/rules/ban-untagged-deprecation


Found 1 problem
```
One concern is that while the starting position of the warning is checked in tests using `assert_lint_err`, the ending position isn’t explicitly covered. If you have ideas on improving this test coverage, or think this behavior could be better, I’d love to hear your thoughts!

### 2. Valid JSDoc (TSDoc) Tags

Since the spec doesn’t seem to clearly define the syntax, I tried testing what the TSDoc Playground recognizes as valid tags. At least, the current behavior matches the following cases. If you think there’s room for improvement here, I’d love to hear your feedback!

Recognized as Valid Tags:
```ts
/** @deprecated */
/**@deprecated */
/*** @deprecated */
/**
 * @deprecated 
*/
/**
 *@deprecated 
*/
/**
 ** @deprecated 
*/
/**
@deprecated 
*/
/**
This function is @deprecated since v2.0.
*/
```

Not Recognized as Valid Tags:
```
/***@deprecated */
/**
 **@deprecated 
*/
```

### 3. Behavior with Multiple `@deprecated` Tags
When multiple `@deprecated` tags are present, it seems that TSDoc only recognizes the last one as valid. However, designing the linter to appropriately check only the last tag would have made the logic unnecessarily complex, so the linter is implemented to check all `@deprecated` tags instead.

### DIFF
--- a/docs/rules/ban_untagged_deprecation.md
+++ b/docs/rules/ban_untagged_deprecation.md
@@ -1,0 +1,20 @@
+`@deprecated` tags must provide additional information, such as the reason for
+deprecation or suggested alternatives.
+
+### Invalid:
+
+```typescript
+/**
+ * @deprecated
+ */
+export function oldFunction(): void {}
+```
+
+### Valid:
+
+```typescript
+/**
+ * @deprecated since version 2.0. Use `newFunction` instead.
+ */
+export function oldFunction(): void {}
+```
--- a/schemas/rules.v1.json
+++ b/schemas/rules.v1.json
@@ -5,6 +5,7 @@
     "ban-ts-comment",
     "ban-types",
     "ban-unknown-rule-code",
+    "ban-untagged-deprecation",
     "ban-untagged-ignore",
     "ban-untagged-todo",
     "ban-unused-ignore",
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -12,6 +12,7 @@ pub mod adjacent_overload_signatures;
 pub mod ban_ts_comment;
 pub mod ban_types;
 pub mod ban_unknown_rule_code;
+pub mod ban_untagged_deprecation;
 pub mod ban_untagged_ignore;
 pub mod ban_untagged_todo;
 pub mod ban_unused_ignore;
@@ -255,6 +256,7 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(ban_ts_comment::BanTsComment),
     Box::new(ban_types::BanTypes),
     Box::new(ban_unknown_rule_code::BanUnknownRuleCode),
+    Box::new(ban_untagged_deprecation::BanUntaggedDeprecation),
     Box::new(ban_untagged_ignore::BanUntaggedIgnore),
     Box::new(ban_untagged_todo::BanUntaggedTodo),
     Box::new(ban_unused_ignore::BanUnusedIgnore),
--- a/src/rules/ban_untagged_deprecation.rs
+++ b/src/rules/ban_untagged_deprecation.rs
@@ -1,0 +1,131 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+use super::{Context, LintRule};
+use crate::Program;
+use deno_ast::swc::common::comments::{Comment, CommentKind};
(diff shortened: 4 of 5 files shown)


bartlomieju:
**Verdict: needs changes** (false positive on multi-line `@deprecated`) — plus a new-rule sign-off. Not merging.

New rule flagging `@deprecated` JSDoc tags with no descriptive text. Mechanics are mostly sound; I verified the tricky parts:
- The `context.all_comments()` (immutable, `'a`-scoped) + `context.add_diagnostic_with_hint()` (mutable) in the same `for_each` is fine — `all_comments()` returns `impl Iterator<Item = &'a Comment>` tied to the program lifetime, not `&self`, so no borrow conflict.
- The range math is calibrated correctly: `comment.start()` points at `/`, `comment.text` excludes `/*`, and `+ BLOCK_COMMENT_OPEN_OFFSET (2)` reproduces the expected `col: 4` for `/** @deprecated */`. ✓

**Correctness — the blocker**
- The regex `^(?:.*\s+|\s*\*\s*)(@deprecated\s*?)$` only accepts descriptive text on the **same line** as `@deprecated`. It therefore **false-positives on the standard multi-line form**, where the reason follows on subsequent lines:
  ```
  /**
   * @deprecated
   * Use `newFunction` instead.
   */
  ```
  Per TSDoc/JSDoc, a block tag's content extends to the next block tag, so this is valid and common — but this rule reports it. Recommend treating `@deprecated` as satisfied if non-whitespace, non-`@tag` content appears on the tag line **or** on following lines up to the next `@tag`/comment end. Add a test for the multi-line case.
- Related: `.*\s+@deprecated$` flags any prose line ending in the word `@deprecated` (the test `* This function is @deprecated` → invalid). That's defensible given the line-based design, but combined with the multi-line gap it means "reason on the next line" is rejected while "trailing mention" is caught — worth aligning intent.

**Minor**
- CRLF: with `\r\n`, `$` (multiline) matches before `\n`, so `\s*?` consumes the `\r` into the captured range — the diagnostic range can include a trailing `\r`. Cosmetic.

**Performance:** fine — `Lazy` single-compiled regex, linear matching, no catastrophic backtracking.

**Housekeeping**
- Edits `schemas/rules.v1.json`, but PR #1377 ("remove schema files for rules and tags") deletes those — these will conflict; coordinate ordering.
- Hand-edits generated `www/static/docs.json`; CI's `diff <(dlint rules --json) docs.json` requires it to match regenerated output.
- New rule is correctly left untagged (opt-in), but adding it is still a maintainer decision. `mergeable: UNKNOWN` → rebase needed.

_Blocker: fix the multi-line false positive (+test). Then new-rule sign-off + rebase._

PR #1380, opened 2024-12-21
refactor: replace website with a redirect to docs.deno.com

bartlomieju:
- [x] Need to double check it works properly
- [x] `docs.deno.com` should link to `docs.deno.com/lint/` instead of `lint.deno.land`


bartlomieju:
Looks like redirects work correctly, so we should be good to merge this one.

crowlKats (approved):

PR #1381, opened 2024-12-30
feat: Remove LintRule::docs()

bartlomieju:
This commit removes `LintRule::docs()` method and consequently all `.md` files
with documentation for the lint rules.

These docs now live at https://github.com/denoland/docs/tree/main/lint/rules.

Some code from `dlint` binary that prints these docs have been removed too
and instead points to relevant https://docs.deno.com/lint/rules/ pages.

This will allow to save about ~400Kb from the `deno` binary size.


dsherret (approved):
Nice!

PR #1383, opened 2025-01-02
feat: add `curly` lint rule

marvinhagemeister:
This PR adds a lint rule for consistent curly braces around control statements. It's similar to eslint's https://eslint.org/docs/latest/rules/curly rule.

Fixes https://github.com/denoland/deno/discussions/27493

Docs: https://github.com/denoland/docs/pull/1302

### DIFF
--- a/src/rules/curly.rs
+++ b/src/rules/curly.rs
@@ -5,6 +5,7 @@ use crate::diagnostic::LintFix;
 use crate::diagnostic::LintFixChange;
 use crate::handler::Handler;
 use crate::handler::Traverse;
+use crate::tags;
 use crate::tags::Tags;
 use crate::Program;
 
@@ -41,7 +42,7 @@ impl LintRule for Curly {
   }
 
   fn tags(&self) -> Tags {
-    &[]
+    &[tags::RECOMMENDED]
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/rules/curly.rs
```diff
@@ -0,0 +1,234 @@
...
+    CurlyHandler.traverse(program, context);
+  }
+
+  fn code(&self) -> &'static str {
+    CODE
+  }
+
+  fn tags(&self) -> Tags {
+    &[]
+  }
```
bartlomieju, line 45: It should be in the recommended set IMO

marvinhagemeister: done

bartlomieju (approved):

dsherret (changes requested):
Going to block the merge of this one. I'm not sure this should be a lint rule because it's something that can be enforced via code formatting. It also conflicts with what deno fmt currently does: https://dprint.dev/playground/#code/JYMwFALgTgrgpgSgAQG8kGMD2A7AzpgGzgDoDMBzMAVgQG4kBfIA/config/N4KABGBEAmCmB2B7SAuMAXATgV1iAvkA/plugin/typescript

PR #1387, opened 2025-01-24
fix: rules-of-hooks not detecting hooks in hooks

marvinhagemeister:
This fixes an error in the `rules-of-hooks` linting rule that didn't detect when a hook is inside another hook as part of a variable declaration.

```ts
// Valid, but would print a lint error
function useFoo() {
  const foo = useState(0);
}
```

Upon looking at the code further it turns out that the variable declaration handling is less involved.

### DIFF
--- a/src/rules/rules_of_hooks.rs
+++ b/src/rules/rules_of_hooks.rs
@@ -190,18 +190,8 @@ impl Handler for RulesOfHooksHandler {
       for item in self.parent_kind.iter().rev() {
         match item {
           ParentKind::Unknown => break,
-          ParentKind::Var(name) => {
-            if !is_hook_or_component_name(name) {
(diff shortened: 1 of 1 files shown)


src/rules/rules_of_hooks.rs
```diff
@@ -211,6 +201,7 @@ impl Handler for RulesOfHooksHandler {
                 DiagnosticKind::Conditionally.hint(),
               );
             } else if !is_hook_or_component_name(name) {
+              eprintln!("====");
```
bartlomieju, line 204: Remove

bartlomieju (approved):

src/rules/rules_of_hooks.rs
marvinhagemeister: oops 😬

PR #1388, opened 2025-01-24
fix: jsx_curly_braces flagging trailing space expression

marvinhagemeister:
The `jsx-curly-braces` rule should not flag string literal expression children when they're at the end of a line and a child is following after it. This is a common pattern to preserve whitespace as HTML whitespace rules would trim it otherwise.

```tsx
// Valid
<div>
  foo{" "}
  <span />
</div>
```


kitsonk:
Just found this one too, so won't open an issue.

bartlomieju (approved):

Issue #1392, opened 2025-01-26
`button-has-type` is too strict about permitted values


lucacasonato:
### Lint Name

`button-has-type`

### Code Snippet

```ts
const foo = () => 'submit';

return <button
autofocus
type={foo()}
>

</button>
```

### Expected Result

This should not be marked. Type checking can check that the return value of `foo()` is actually submit. The linter should just mark if `type` is missing entirely.

### Actual Result

```
error[button-has-type]: Incorrect value for button's `type` attribute
 --> /home/lucacasonato/projects/github.com/denoland/deno/test.tsx:5:6
  | 
5 | type={foo()}
  |      ^^^^^^^
  = hint: The value of the `type` attribute must be one of `button`, `submit` or `reset`

  docs: https://lint.deno.land/rules/button-has-type


Found 1 problem
Checked 1 file
```

### Version

2.1.7+e475749

lucacasonato:
I understand this lint, but in all honesty it is a bit flawed without type access. The following is perfectly fine, but the linter complains about it:

```tsx
function Button(props: { type: "submit" }) {
  return <button {...props}>Foo</button>;
}
```

PR #1402, opened 2025-01-27
fix(button-has-type): treat expressions as valid value

marvinhagemeister:
We don't have enough information in the plain AST to deduce if any expression results into a valid value because we don't have type information. Therefore we'll treat these cases as valid.

```jsx
// Valid
<button {...props} />
<button type={type} />
<button type={type()} />
```

Fixes https://github.com/denoland/deno_lint/issues/1392


lucacasonato (approved):
LGTM

PR #1406, opened 2025-01-28
feat: add no-empty-case

ahaoboy:
Remove empty switch-case statements

```js
switch (n) {
  case 0: { }
  case 1: { }
  default: {
    // TODO
  }
}
```
```js
switch (n) {
  case 0: 
  case 1: 
  default: {
    // TODO
  }
}
```

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -59,6 +59,7 @@ pub mod no_dupe_else_if;
 pub mod no_dupe_keys;
 pub mod no_duplicate_case;
 pub mod no_empty;
+pub mod no_empty_case;
 pub mod no_empty_character_class;
 pub mod no_empty_enum;
 pub mod no_empty_interface;
@@ -300,6 +301,7 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(no_dupe_keys::NoDupeKeys),
     Box::new(no_duplicate_case::NoDuplicateCase),
     Box::new(no_empty::NoEmpty),
+    Box::new(no_empty_case::NoEmptyCase),
     Box::new(no_empty_character_class::NoEmptyCharacterClass),
     Box::new(no_empty_enum::NoEmptyEnum),
     Box::new(no_empty_interface::NoEmptyInterface),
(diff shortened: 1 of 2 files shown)


bartlomieju:
**Verdict: needs changes** — the detection logic has a panic and a destructive false-positive. Not merging (new rule + bugs).

The handler assumes the single case statement is always a `{...}` block and slices its text: `text[1..text.len() - 1].trim().is_empty()`. That breaks when the statement isn't a block:

**Correctness — blocking**
1. **Panic** on an empty statement. `switch (n) { case 0:; }` → the single `cons` statement is `;`, so `stmt.text()` is `";"` (len 1), and `text[1..text.len()-1]` = `text[1..0]` → Rust panics (`slice index starts at 1 but ends at 0`). A lint rule must never panic on valid input.
2. **False positive + destructive autofix** on a 2-char statement. `switch (n) { case 0: a; }` → `stmt.text()` is `"a;"` (len 2), `text[1..1]` = `""` → trimmed empty → the case is flagged as "empty" and the **autofix deletes the real statement `a;`**. That's data loss on correct code.
3. **UTF-8 byte-slice panic risk**: `text[1..text.len()-1]` slices by byte offset. A statement whose text starts or ends with a multi-byte char (e.g. a non-ASCII identifier) can panic on a non-char-boundary slice.

**Root cause / fix**
Don't infer "empty block" from string trimming. Match the statement kind instead:
```rust
if let [stmt] = switch_case.cons.iter().as_slice() {
    if let deno_ast::view::Stmt::Block(block) = stmt {
        // treat as empty only if it has no statements...
    }
}
```
Note the current text approach does get one thing right that a naive `block.stmts.is_empty()` would miss: `case 1: { /* todo */ }` (comment-only block) is correctly treated as *non*-empty by the existing valid test. If you switch to an AST check, preserve that by also checking there are no comments within the block's range (otherwise you'll regress that case).

**Missing coverage**
- Add tests for `case 0:;` (must not panic) and `case 0: a;` (must NOT be flagged) — both currently fail.
- A comment-only block via the AST path (guard the regression noted above).

**Perf:** fine — per-`switch_case` visit; `stmt.text()` is cheap.

**Housekeeping:** new rule tagged `recommended` (product decision) and `mergeable: UNKNOWN` (stale → rebase).

_Blocker: fix the panic + destructive false-positive by checking the statement is actually an empty block, then add the two regression tests._

PR #1408, opened 2025-01-28
dlint: add --fix

ahaoboy:
Add `--fix` option to dlint, same as `deno lint --fix` command. This can be used with git to easily observe how a new change affects the code

apply_lint_fixes seems to be reusable? I simply copied from deno code

### DIFF
--- a/examples/dlint/diagnostics.rs
+++ b/examples/dlint/diagnostics.rs
@@ -1,8 +1,8 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
-use std::collections::HashSet;
 use deno_ast::{diagnostics::Diagnostic, SourceTextInfo};
 use deno_lint::diagnostic::LintDiagnostic;
+use std::collections::HashSet;
 
 pub fn display_diagnostics(
   diagnostics: &[LintDiagnostic],
--- a/examples/dlint/diagnostics.rs
+++ b/examples/dlint/diagnostics.rs
@@ -1,6 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
-use deno_ast::diagnostics::Diagnostic;
+use std::collections::HashSet;
+use deno_ast::{diagnostics::Diagnostic, SourceTextInfo};
 use deno_lint::diagnostic::LintDiagnostic;
 
 pub fn display_diagnostics(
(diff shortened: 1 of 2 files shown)


bartlomieju:
**Verdict: needs changes** — `--fix` panics on the common case. Not merging (feature + bug, CI unknown).

Nice feature (mirrors `deno lint --fix` for the `dlint` example), but the fix path crashes in normal use.

**Correctness — blocking**
- **Guaranteed panic on clean/unfixable files.** In `main.rs` the diagnostics map is populated for *every* linted file (`lock.insert(file_path, diagnostics)` is unconditional — even files with zero diagnostics). The new fix loop then does:
  ```rust
  let s = apply_lint_fixes(&info, d).expect("failed to apply_lint_fixes");
  ```
  But `apply_lint_fixes` returns `None` when `diagnostics.is_empty()` **or** when no diagnostic carries a fix. So the moment `dlint --fix` hits a clean file — or any file whose diagnostics aren't auto-fixable (the majority of rules) — `.expect(...)` panics. In practice `--fix` will almost always crash.
  Fix: `if let Some(s) = apply_lint_fixes(&info, d) { std::fs::write(file_path, s)?; }` and drop the `.expect`.
- **Single-pass only, despite the comment.** The copied comment says "we'll circle back for another pass to fix the remaining," but there's no retry loop — overlapping fixes are dropped and never re-applied. `deno lint --fix` wraps this in an iterate-until-stable loop. As-is, overlapping fixes silently require the user to re-run `--fix` N times. Either add the loop or adjust the comment/expectations.

**Performance — minor**
- `source_code: source_code.clone()` and storing `source_code` in the map now happen unconditionally, so every run (even without `--fix`) pays an extra full-source clone per file, plus another `source_code.clone().into()` in the fix branch. Gate the retention/clone on `fix` to avoid the cost on the common no-fix path.

**Style**
- `std::fs::write` runs even when the fixed text equals the original; skip the write when unchanged to avoid needless disk churn / mtime bumps.
- The `import_fixes` dedup hack is deno-specific (author notes it was copied); fine to keep, but a comment on why dlint needs it would help.

**Housekeeping:** `examples/dlint` only (no library/rule impact), `mergeable: UNKNOWN` → rebase.

_Blocker: replace `.expect` with an `if let Some(..)` so `--fix` doesn't panic on clean/unfixable files; then decide on single- vs multi-pass._

PR #1409, opened 2025-01-29
fix: some bug for verbatim module syntax

dsherret:
I had this in my working tree from many months ago. Opening a PR so it's not forgotten.

### DIFF
--- a/src/rules/verbatim_module_syntax.rs
+++ b/src/rules/verbatim_module_syntax.rs
@@ -41,6 +41,10 @@ enum Hint {
   ChangeImportToImportType,
   #[display(fmt = "Change `export` to `export type`")]
   ChangeExportToExportType,
+  #[display(fmt = "Extract default import to type-only import declaration")]
+  ExtractDefaultImportType,
+  #[display(fmt = "Change `import` to `import type` and extract out separate `import type` for default import")]
+  ChangeImportToImportTypes,
   #[display(fmt = "Add a `type` keyword before the identifier")]
   AddTypeKeyword,
 }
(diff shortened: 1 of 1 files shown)


bartlomieju:
**Verdict: WIP (draft) — not merging.** Noted as a placeholder ("so it's not forgotten"); a couple of concrete things for whoever picks it up.

**Observations (draft)**
- The diff adds two new `Hint` variants (`ExtractDefaultImportType`, `ChangeImportToImportTypes`) and two new `invalid` test cases, but **no handler/fixer implementation** to actually emit those hints or produce those fixes. As-is these tests would fail — the behavior they describe isn't wired up yet. That's expected for a stash dump, just flagging that the implementation half is missing.
- The second new test's expected `fix` looks like a **copy-paste error**: for input `import a, { b } from 'react';\nexport type Test = a | b;`, the expected output is
  `"import type a from 'react'\nimport type { b } from 'react';\nuseState(true);\nexport type Test = React;"`
  — the `useState(true);` and `export type Test = React;` lines are carried over from the *previous* test case and don't correspond to this input. It also mixes `message: AllExportIdentsUsedInTypes` with `hint: ChangeImportToImportTypes`; worth confirming that pairing is intended.
- Minor consistency: the first case's fix emits `import type React from 'react'` (no semicolon) right next to `import { useState } from 'react';` (with semicolon). Probably want the trailing `;` for consistency.

**Perf:** n/a (tests + enum only).

_Left as draft. To finish: implement the default-import extraction fixer for the mixed `import Default, { named }` case, then correct the second test's expected output._

PR #1410, opened 2025-01-29
feat: add non-generic-typed-array rule

dsherret:
Not sure about the name.

This is an optional like rule that forces making types like `Uint8Array` generic when not used in parameters. We'll use it in the CLI and deno/std.


dsherret:
Not doing this.

PR #1416, opened 2025-02-14
fix(rule-of-hooks): handle var decl with destructuring

kt3k:
This PR fixes a case of rule-of-hooks. Currently the example below doesn't produce lint diagnostics because of the handling of var declaration:

```js
function Foo() {
  if (cond) {
    const [state, setState] = useState(0);
  }
}
```

This PR fixes the above case.


marvinhagemeister (approved):
LGTM, nice catch

PR #1418, opened 2025-02-19
fix(react-rules-of-hooks): don't error on hook following non-terminal if statement

marvinhagemeister:
If an if-statement is placed before a hook it doesn't necessarily mean that the hook was called conditionally. It only is, if the if-statement contains a return statement.

```ts
// This should error
function Foo() {
  if (condition) {
   return
  }

  useState(0)
}


// ...but this is ok
function Foo() {
  let count = 0;
  if (condition) {
   count++
  }

  useState(0)
}
```


dsherret (approved):
LGTM

bartlomieju (approved):

PR #1420, opened 2025-02-21
fix(jsx-curly-braces): ignore values that require escaping

marvinhagemeister:
Don't turn expression into a string when the JSX child would require escaping or it would clash with `jsx-no-unescaped-entities` rule.

Fixes https://github.com/denoland/deno/issues/28225


dsherret (approved):
LGTM

PR #1424, opened 2025-02-25
add "semi" rule

ry:


dsherret:
Isn't this a formatting concern and not linting?

bartlomieju:
Closing as out of scope. deno_lint intentionally carries no purely stylistic/formatting rules — semicolon insertion is fully handled by `deno fmt`, which is the canonical place for formatting in the Deno toolchain. A `semi` lint rule would duplicate the formatter. Thanks for the contribution!

PR #1425, opened 2025-02-26
feat: add quotes rule

ry:


dsherret:
Same here. This is a formatting concern and shouldn't be a lint rule I think.

bartlomieju:
Closing as out of scope, same rationale as #1424: quote style is handled by `deno fmt`, and deno_lint deliberately avoids stylistic/formatting rules that the formatter already covers. Thanks for the contribution!

PR #1434, opened 2025-05-08
chore: fix rust version on README.md

unvalley:
minor fix: the Rust version in the README was outdated, so this pr updates it.


dsherret (approved):
Thanks!

PR #1443, opened 2025-06-18
fix(no-invalid-regexp): allow v-mode regexes containing -- after single token in a char class

### LINKED ISSUE #1442: `no-invalid-regexp`: some `v`-mode regexes are marked invalid when actually valid, e.g. `/[\p{N}--0]/gv`
Some `v`-mode regexes are marked invalid when actually valid, such as a character class subtracted from a bare Unicode property character class.

### Lint Name

`no-invalid-regexp`

### Code Snippet

```ts
// Lint error "Invalid RegExp literal: Rework regular expression to be a valid deno-lint(no-invalid-regexp)"
const re = /[\p{N}--0]/gv;
// works fine at runtime, logging ['1', '2', '3'] as expected
console.log('1203'.match(re));
// identical semantics and runtime behavior, redundant brackets around unicode property eliminate the lint error
const re2 = /[[\p{N}]--0]/gv;
```

### Expected Result

No lint error for regexes like `/[\p{N}--0]/gv`

### Actual Result

Lint error

### Additional Info

N/A

### Version

deno 2.3.3 (stable, release, x86_64-unknown-linux-gnu)
v8 13.7.152.6-rusty
typescript 5.8.3

lionel-rowe:
Fixes https://github.com/denoland/deno_lint/issues/1442

The large diff in src/js_regex/mod.rs is mostly just ctrl+h test changes due to the validator methods now accepting/returning `UnicodeMode` enums (`u`/`v`/none) instead of booleans (`u`/none).

### DIFF
--- a/src/js_regex/mod.rs
+++ b/src/js_regex/mod.rs
@@ -3,7 +3,6 @@
 mod reader;
 mod unicode;
 mod validator;
-
 use std::fmt;
 
 pub use validator::{EcmaRegexValidator, EcmaVersion};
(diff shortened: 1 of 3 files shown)


bartlomieju:
**Verdict: LGTM on the implementation — needs maintainer sign-off** (user-visible behavior change to a `recommended` rule) + a rebase. Not auto-merging (behavior change + large).

Most of the +2980/-1029 is mechanical test churn from the `bool` → `UnicodeMode` refactor, as the author notes. I reviewed the real logic in `validator.rs` and `no_invalid_regexp.rs`.

**Correctness — good**
- `validate_flags` now returns `UnicodeMode` (`None`/`Unicode`/`UnicodeSets`) and correctly rejects `u`+`v` together and gates `v` behind ES2024 (previously wrongly allowed at ES2022). The `[true,true] => Err` / `[true,_]` / `[_,true]` match is correct.
- The rule's `check_regex` refactor replaces a genuinely buggy hack — the old code validated the pattern with `u=true` *and* `u=false` and only flagged when *both* failed — with a single, mode-correct `validate_flags()? → validate_pattern(mode)`. Cleaner and more correct (empty flags now validate as non-unicode instead of the double-check heuristic).
- Invalid-test coverage is **improved**, not lost: each case now asserts the specific error (`Invalid RegExp literal: Unterminated character class`, etc.), plus new cases for `/(?:)/uv` and u-mode `/[\p{L}--0]/u`. Nice.
- Typo fix in `HINT` ("be a valid" → "be valid").

**Intentional gap worth a maintainer's eye**
- The core fix (issue #1442) skips character-class range validation entirely when `v_flag` is set:
  ```rust
  // v flag has different semantics around `--` ... we just skip validation for now.
  if self.v_flag { continue; }
  ```
  This stops the false positives on valid v-mode set subtraction (`/[\p{N}--0]/v`), which is the right direction for a lint rule (prefer false-negatives over false-positives). The trade-off: genuinely-invalid v-mode character classes will no longer be caught. That's acceptable but should be a conscious decision, ideally with a tracking issue for full v-mode set-operation validation.

**Performance — improved**
- Now validates each pattern once (was up to twice per regex). Net win.

**Housekeeping**
- User-visible: diagnostic messages now include the specific error suffix, and `v` requires ES2024 — behavior changes on a `recommended` rule, so they need sign-off.
- `mergeable: UNKNOWN` → stale; needs a rebase.

_No code changes requested. Blockers are the maintainer sign-off on the message/semantic changes + a rebase; consider filing a follow-up for real v-mode `--` validation._

PR #1445, opened 2025-06-18
fix: nested switch empty fallthrough

### LINKED ISSUE #1444: bug: false positive of no-fallthrough for empty fallthrough in nested switch
### Lint Name

no-fallthrough

### Code Snippet

```ts
function booleanAnd(b1: boolean, b2: boolean) {
  switch (b1) {
    case true:        // false positive lint here
      switch (b2) {
        case true:
        default:
          return true;
      }
    case false:
      switch (b2) {
        case true:
          return true;
        case false:
          return false;
      }
  }
}
```

### Expected Result

No linting errors.

### Actual Result

```bash
❯ deno lint foo.ts 
 --> /home/xubaiw/foo.ts:3:5
  | 
3 |     case true:
  |     ^^^^^^^^^^
  | 
4 |       switch (b2) {
  | ^^^^^^^^^^^^^^^^^^^
  | 
5 |         case true:
  | ^^^^^^^^^^^^^^^^^^
  | 
6 |         default:
  | ^^^^^^^^^^^^^^^^
  | 
7 |           return true;
  | ^^^^^^^^^^^^^^^^^^^^^^
  | 
8 |       }
  | ^^^^^^^
  = hint: Add `break` or comment `/* falls through */` to your case statement

  docs: https://docs.deno.com/lint/rules/no-fallthrough


Found 1 problem
Checked 1 file
```

### Additional Info

This issue is different from #1331 , which is involved with `End::Break`, while this issue is about `End::Continue`.

Also reported in https://github.com/denoland/deno/issues/29805 (closed now).

This problem may arise from here:

https://github.com/denoland/deno_lint/blob/4a56a4939012b533ecd775f3bc2c36812af235c6/src/control_flow/mod.rs#L480-L488

The fallthrough case is considered `End::Continue`, and fail to merge in `merge_forced`. 

### Version

deno 2.3.5 (stable, release, x86_64-unknown-linux-gnu)
v8 13.7.152.6-rusty
typescript 5.8.3

xbwwj:
Close #1444 

This PR fixes empty fallthrough case `End` merge logic by tracking `End::Continue` state of each case.

Also add a test case of boolean and operation using nested switch.

### DIFF
--- a/src/control_flow/mod.rs
+++ b/src/control_flow/mod.rs
@@ -479,13 +479,25 @@ impl Visit for Analyzer<'_> {
         .iter()
         .filter_map(|case| self.get_end_reason(case.start()))
         .try_fold(
-          End::Forced {
-            ret: false,
-            throw: false,
-            infinite_loop: false,
+          (
+            End::Forced {
+              ret: false,
+              throw: false,
+              infinite_loop: false,
+            },
+            // record if previous is continue
+            false,
+          ),
+          |acc, cur| {
+            if matches!(cur, End::Continue) {
+              Some((acc.0, true))
+            } else {
+              acc.0.merge_forced(cur).map(|acc| (acc, false))
+            }
           },
-          |acc, cur| acc.merge_forced(cur),
-        );
+        )
+        // None if last is continue
+        .and_then(|(acc, last_continue)| (!last_continue).then_some(acc));
 
       match forced_end {
         Some(e) if has_default => e,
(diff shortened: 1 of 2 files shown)


bartlomieju:
Closing as superseded. The exact issue this fixes (#1444) is already resolved on `main` by the merged #1497, which addresses the same nested-switch `End::Continue` empty-fallthrough case (and adds the identical `booleanAnd` test). Issue #1444 is now closed. Thanks for the fix and the precise root-cause analysis — it helped.

PR #1466, opened 2025-11-26
feat: add `no-const-assign` to the `recommended` set

### LINKED ISSUE #1465: linter handles `prefer-as-const`, but not `no-const-assign`
The linter often bugs me to use `const`. But it never tells me that I accidentally assigned to a `const`, that I have to find out at runtime. Why doesn't the linter support `no-const-assign`?

tobias47n9e:
Fixes #1465


bartlomieju (approved):
Thanks!

PR #1470, opened 2025-12-12
fix: explicit-function-return-type now also works on arrow functions

jradam:
This is a fix for https://github.com/denoland/deno_lint/issues/1334

The `explicit-function-return-type` rule was not implemented for arrow functions, so I added it.

~~This fix aligns the deno_lint implementation of this rule with the default [typescript-eslint behaviour](https://typescript-eslint.io/rules/explicit-function-return-type/).~~

This fix gets deno_lint closer to the default [typescript-eslint behaviour](https://typescript-eslint.io/rules/explicit-function-return-type/), but also makes deno_lint somewhat stricter - as default typescript-eslint allows skipping the return type on arrow functions in a few cases.

Happy to extend this PR to include those cases - though I suppose it's up to a maintainer to decide how strict this should be.

### DIFF
--- a/src/rules/explicit_function_return_type.rs
+++ b/src/rules/explicit_function_return_type.rs
@@ -62,6 +62,17 @@ impl Handler for ExplicitFunctionReturnTypeHandler {
       );
     }
   }
+
+  fn arrow_expr(&mut self, arrow: &ast_view::ArrowExpr, context: &mut Context) {
+    if arrow.return_type.is_none() {
+      context.add_diagnostic_with_hint(
+        arrow.range(),
+        CODE,
+        ExplicitFunctionReturnTypeMessage::MissingRetType,
+        ExplicitFunctionReturnTypeHint::AddRetType,
+      );
+    }
+  }
 }
 
 #[cfg(test)]
@@ -73,10 +84,12 @@ mod tests {
     assert_lint_ok! {
       ExplicitFunctionReturnType,
       "function fooTyped(): void { }",
-      "const bar = (a: string) => { }",
       "const barTyped = (a: string): Promise<void> => { }",
       "class Test { set test(value: string) {} }",
       "const obj = { set test(value: string) {} };",
+      "const arrow = (a: string): void => { }",
+      "const arrow = (): number => 0",
+      "const arrow = async (): Promise<void> => {}",
     };
 
     assert_lint_ok! {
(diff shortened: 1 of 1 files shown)


bartlomieju:
**Verdict: needs human decision** — the implementation is correct, but it's stricter than the typescript-eslint *default* it claims to match, which makes it noisy on idiomatic callbacks. Not merging (behavior change to an existing rule).

**Correctness — the code itself is fine**
- `arrow_expr` flags any `ArrowExpr` with `return_type.is_none()`; range/col are right (tests assert `col: 14`), it's consistent with the existing `function` handler, and the valid/invalid tests were updated correctly (the previously-valid `const bar = (a: string) => {}` is now invalid — the intended change). No panics, no perf concerns (one visit per arrow).

**But the "aligns with typescript-eslint default" claim isn't accurate**
The typescript-eslint default enables several allowances this PR doesn't implement, so it will flag code their default leaves alone:
- `allowTypedFunctionExpressions: true` (default) → typescript-eslint does **not** flag a contextually-typed arrow, e.g. `const f: () => void = () => {}` or `setTimeout(() => {}, 100)` or `arr.map((x) => x)` where the callback type is known. **This PR flags all of those.** Since arrows are overwhelmingly used as callbacks, that's a large false-positive surface vs. the stated baseline.
- `allowHigherOrderFunctions: true` (default) → `const f = () => () => {}` isn't flagged on the outer arrow by typescript-eslint; here it would be.
- `allowExpressions` / `allowDirectConstAssertionInArrowFunctions` similarly unhandled.

None of this is a bug per se — deno_lint's existing rule is already a simplified (stricter) variant and is **opt-in** (no `recommended` tag), which limits blast radius. But two things for a maintainer to decide:
1. Reword the PR/claim: this is *stricter than* the typescript-eslint default, not aligned with it.
2. Decide whether to implement the `allowTypedFunctionExpressions`/`allowHigherOrderFunctions` exceptions (now feasible as rule options given the new per-rule config support) before enabling arrow checks, or intentionally ship the stricter form.

**Housekeeping:** `mergeable: UNKNOWN` → rebase needed.

_Recommend: a maintainer decides strict-vs-eslint-default; if strict is intended, at least fix the alignment claim. Otherwise add the typed-expression/higher-order allowances (+ tests) so common callbacks aren't flagged._

jradam:
> Recommend: a maintainer decides strict-vs-eslint-default; if strict is intended, at least fix the alignment claim. Otherwise add the typed-expression/higher-order allowances (+ tests) so common callbacks aren't flagged.

Thanks for looking into this @bartlomieju - fair points. I've updated the PR summary, and will wait for further input on how strict this should be. Happy to make any changes required.

PR #1476, opened 2026-02-27
feat: add no-runtime-private methods rule

tobias47n9e:
This optional rule discourages the usage of methods with `#`-prefix in favor of `private method`. This encourages users to trust typescript's `private` keyword and does not prevent debuggers from accessing the method.

Note: I would like to make a similar rule for `#`-prefixed instance properties. Those are also not accessible by Deno debug, and I think it would be good if the linter could discourage using them.

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -93,6 +93,7 @@ pub mod no_process_global;
 pub mod no_prototype_builtins;
 pub mod no_redeclare;
 pub mod no_regex_spaces;
+pub mod no_runtime_private_methods;
 pub mod no_self_assign;
 pub mod no_self_compare;
 pub mod no_setter_return;
@@ -340,6 +341,7 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(no_prototype_builtins::NoPrototypeBuiltins),
     Box::new(no_redeclare::NoRedeclare),
     Box::new(no_regex_spaces::NoRegexSpaces),
+    Box::new(no_runtime_private_methods::NoRuntimePrivateMethods),
     Box::new(no_self_assign::NoSelfAssign),
     Box::new(no_self_compare::NoSelfCompare),
     Box::new(no_setter_return::NoSetterReturn),
--- a/src/rules/no_runtime_private_methods.rs
+++ b/src/rules/no_runtime_private_methods.rs
@@ -1,0 +1,114 @@
+use super::program_ref;
+use super::{Context, LintRule};
+use crate::tags::Tags;
+use crate::Program;
+use crate::ProgramRef;
+use deno_ast::swc::ast::PrivateMethod;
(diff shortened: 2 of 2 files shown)


0f-0b:
At least V8 provides APIs for debuggers to access private elements. This is what I get using VS Code's built-in debugger with Deno:

tobias47n9e:
@0f-0b For me it is also about the watch functionality of the debugger. I cannot easily check on nested methods and fields that have the `#`-prefix at a breakpoint.

``` ts
this.#database.#somePrivateMethod()
```

``` ts
this.#database.#privateField
```

It could be out of scope for deno lint, but I just wanted to check if something like this would be considered :smile:

bartlomieju:
**Verdict: needs changes** (missed-nested false negative + missing copyright header) — plus new-rule sign-off. Not merging.

New opt-in rule flagging `#method()` private methods in favor of `private method()`.

**Correctness**
- **False negative: nested private methods are missed.** `visit_private_method` overrides the default `Visit` behavior but never calls `method.visit_children_with(self)`, so traversal stops descending once it hits a private method. Anything inside a private method's body is skipped:
  ```ts
  class Outer {
    #outer() {
      class Inner { #inner() {} }   // NOT flagged
    }
  }
  ```
  Fix: end `visit_private_method` with `method.visit_children_with(self);`. (Switching to the project's `ast_view` `Handler`/`Traverse` pattern — as nearly all other rules use — would auto-recurse and avoid this class of bug entirely; see below.)
- The second invalid test input `private #keywordAndRuntimePrivate() {}` is **not valid TypeScript** — you can't combine an accessibility modifier with a `#` private name ("An accessibility modifier cannot be used with a private identifier"). swc may parse it leniently, but it's a misleading fixture; use a valid example.

**Consistency / housekeeping**
- **Missing copyright header.** The file starts directly with `use super::program_ref;`; every other rule file begins with `// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.`. The repo's license check will likely fail CI without it.
- No `docs/rules/no_runtime_private_methods.md` or `#[cfg(feature = "docs")] fn docs()`. New rules conventionally ship docs, and the docs.json CI diff expects them.
- This rule uses the low-level `deno_ast::swc::ecma_visit::Visit` directly instead of the `impl Handler` + `Traverse` pattern used by essentially every other rule. Aligning it would fix the recursion bug for free and keep the codebase consistent.

**Perf:** fine (single visitor pass) — though the missing recursion makes it "cheap" precisely because it skips subtrees it shouldn't.

**Product decision:** correctly untagged (opt-in), but whether to discourage `#` private methods (true runtime privacy) in favor of TS `private` is an opinionated call for a maintainer. `mergeable: UNKNOWN` → rebase.

_Blockers: add `visit_children_with` (or port to `Handler`), add the copyright header, fix the invalid test input, add docs. Then maintainer sign-off._

PR #1477, opened 2026-02-28
feat: add no-runtime-private-fields rule

tobias47n9e:
See also: https://github.com/denoland/deno_lint/pull/1476

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -93,6 +93,7 @@ pub mod no_process_global;
 pub mod no_prototype_builtins;
 pub mod no_redeclare;
 pub mod no_regex_spaces;
+pub mod no_runtime_private_fields;
 pub mod no_self_assign;
 pub mod no_self_compare;
 pub mod no_setter_return;
@@ -340,6 +341,7 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(no_prototype_builtins::NoPrototypeBuiltins),
     Box::new(no_redeclare::NoRedeclare),
     Box::new(no_regex_spaces::NoRegexSpaces),
+    Box::new(no_runtime_private_fields::NoRuntimePrivateFields),
     Box::new(no_self_assign::NoSelfAssign),
     Box::new(no_self_compare::NoSelfCompare),
     Box::new(no_setter_return::NoSetterReturn),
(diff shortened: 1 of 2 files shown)


bartlomieju:
**Verdict: needs changes** (same false-negative + missing header as #1476, plus a rule-overlap and a misleading hint) — and new-rule sign-off. Not merging.

New opt-in rule flagging `#field` private props. It shares the exact issues I flagged on its sibling #1476, so those apply here too:

**Correctness**
- **Nested private members missed (false negative).** Neither `visit_private_method` nor `visit_private_prop` calls `visit_children_with(self)`, so traversal stops descending into the flagged node. A private prop whose initializer contains a nested class is skipped:
  ```ts
  class Outer {
    #x = class { #y = 1; };   // #y NOT flagged
  }
  ```
  Fix: call `prop.visit_children_with(self)` / `method.visit_children_with(self)` (or port to the `ast_view` `Handler`/`Traverse` pattern used by other rules, which recurses correctly).
- **Overlap with #1476.** This rule *also* implements `visit_private_method`, so `#method()` is flagged by **both** `no-runtime-private-fields` and `no-runtime-private-methods`. If a user enables both, they get duplicate diagnostics, and the "fields" rule reporting methods is surprising. Decide: either drop method handling here (rely on #1476), or merge the two PRs into a single `no-runtime-private` rule.
- **Misleading `Static` hint.** For `static #FOO`, the hint is "Use `static FOO`" — but that makes the field *public*, changing semantics. The privacy-preserving equivalent is `private static FOO`. Fix the hint text.
- Invalid TS fixture again: `private #keywordAndRuntimePrivate = 3;` isn't valid TS (accessibility modifier + `#` name can't combine). Use a valid example.

**Housekeeping**
- **Missing copyright header** — file starts with `use super::program_ref;`; every other rule begins with the `// Copyright ...` line. Likely fails the CI license check.
- No `docs/rules/*.md` / `#[cfg(feature = "docs")] fn docs()`; docs.json CI diff expects them.

**Perf:** fine (single visitor pass), modulo the missing recursion.

**Product decision:** untagged/opt-in (good), but this + #1476 should be designed together — one combined rule vs. two overlapping ones is a maintainer call. `mergeable: UNKNOWN` → rebase.

_Blockers: add recursion, fix the copyright header + static hint + invalid fixture, and resolve the method-overlap with #1476 (ideally coordinate the two into one rule)._

PR #1377, opened 2024-12-11
chore: remove schema files for rules and tags

bartlomieju:
This commit removes the schema files for rules and tags, that
were used by Deno <2.2.

This is now done in the Deno repo with https://github.com/denoland/deno/pull/27324
landed.

Closes https://github.com/denoland/deno_lint/issues/1376

PR #1478, opened 2026-03-19
feat: add prefer-object-shorthand rule

### LINKED ISSUE #1184: Rule suggestion: prefer-object-shorthand
as in eslint: https://eslint.org/docs/latest/rules/object-shorthand

scarf005:
## Summary
- add `prefer-object-shorthand` for `{ foo: foo }` and `{ foo: function() {} }`
- provide quick fixes for property, method, async/generator, and computed-key shorthand cases
- add the new rule to the rule registry and schema

Closes #1184

### DIFF
--- a/schemas/rules.v1.json
+++ b/schemas/rules.v1.json
@@ -116,6 +116,7 @@
     "prefer-ascii",
     "prefer-const",
     "prefer-namespace-keyword",
+    "prefer-object-shorthand",
     "prefer-primordials",
     "require-await",
     "require-yield",
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -119,6 +119,7 @@ pub mod prefer_as_const;
 pub mod prefer_ascii;
 pub mod prefer_const;
 pub mod prefer_namespace_keyword;
+pub mod prefer_object_shorthand;
 pub mod prefer_primordials;
 pub mod react_no_danger;
 pub mod react_no_danger_with_children;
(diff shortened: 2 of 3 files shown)


bartlomieju:
**Verdict: WIP (draft) — looking solid; not merging yet.** New opt-in `prefer-object-shorthand` rule with quick-fixes. The logic checks out well; notes below for finishing it.

**Correctness — I traced the fixer paths and they hold**
- Property shorthand: gated on `PropName::Ident` key + `Expr::Ident` value with equal `sym()`, replacing `foo: foo` → `foo`. Correctly skips `'foo': foo` (string key) and `[foo]: foo` (computed). ✓
- Method shorthand: early-returns on named function expressions (`fn_expr.ident.is_some()`), so `{ foo: function foo() {} }` stays valid. The `FUNCTION_EXPR_REGEX` reconstruction correctly handles anonymous, `async`, `function*`, `async function*`, string keys, and computed keys — all covered by tests, and it preserves `this`/`arguments` semantics since it stays a non-arrow method. ✓
- TS-friendly: because `tail` captures everything after `function[*]`, type params / return types survive (`function(): void {}` → `foo(): void {}`). Nice.
- No panic risk: the `let Some(captures) = ... else return` guards the regex miss.

**Minor**
- Whitespace between `function` and `(` (e.g. `function () {}`) is carried into `tail`, yielding `foo () {}`. Valid JS, but slightly ugly; `deno fmt` would normalize it. Optional to trim.
- Consider a test with a comment between `function` and params to lock in behavior.

**To finish (draft)**
- No `docs/rules/prefer_object_shorthand.md` / `#[cfg(feature="docs")] fn docs()` yet — needed, and the docs.json CI diff will expect it.
- Edits `schemas/rules.v1.json`, which PR #1377 deletes — these will conflict; coordinate ordering (and drop this schema edit if #1377 lands first).
- New rule is untagged (opt-in) — good default; still a maintainer product decision to accept. `mergeable: UNKNOWN` → rebase.

**Perf:** fine — per-`key_value_prop` visit; one `Lazy` regex compiled once.

_Close to ready. Remaining: docs, resolve the schema-file conflict, and maintainer sign-off._

PR #1479, opened 2026-03-20
feat: add in-process GritQL support to dlint

scarf005:
## Summary
- compile and execute GritQL patterns in-process in `examples/dlint`
- add rewrite-only `--fix` support with JS, JSON, and JSONC coverage
- replace mock-binary coverage with real-engine fixtures and config wiring


bartlomieju:
**Verdict: WIP (draft) — well-engineered; needs a maintainer direction call.** Not merging (draft). Confined to `examples/dlint` (no library/rule impact), so lower risk.

Reviewed `grit.rs` (the core) closely. The design is solid for a first cut.

**Correctness — good defensive handling**
- `apply_fixes` reads original text, runs each applicable pattern sequentially on the evolving `current_text`, and only records a change if the text actually differs — clean.
- `extract_supported_rewrite` guards the tricky cases well: it bails on cross-file rewrites, on rewrites computed against **stale** source (`original_content != current_text`), and on conflicting rewrites from one pattern. That's exactly the class of GritQL foot-guns worth rejecting rather than silently mis-applying.
- `diagnostics_from_results` maps `Match`/`Rewrite`/`CreateFile`/`RemoveFile` to diagnostics and correctly ignores the non-actionable `PatternInfo`/`AllDone`/`InputFile`/`DoneFile`/`AnalysisLog` variants.

**Things to decide / finish (draft)**
- **Dependency weight:** this pulls `marzano_core`/`marzano_language`/`marzano_util` into the build (big Cargo.lock churn). In-process GritQL is a meaningful architectural commitment for the example — worth an explicit maintainer decision vs. the prior mock-binary approach before investing further.
- **Single-pass fixing:** each pattern runs once; there's no fixpoint loop, so a pattern that becomes applicable *after* another pattern's rewrite won't re-fire. Fine as a first cut — just document the limitation (as you did for "rewrite-only `--fix`"; `CreateFile`/`RemoveFile` are diagnostics-only).
- **Byte ranges:** results use `start_byte`/`end_byte` as `usize` into `create_diagnostic`. Confirm those map to `deno_ast` byte-based `SourceRange` for multi-byte UTF-8 sources (add a non-ASCII fixture to be safe).
- Non-test `.unwrap()`s: the ones I saw are in tests (temp-dir setup) which is fine; keep production paths on `?`/`with_context` as they currently are.

**Perf:** `applicable_patterns`/`collect_diagnostics` recompute `detect_file_language` per call — negligible.

_Strong foundation. Main blocker is the maintainer direction call on bundling the Grit engine; then docs + the fixpoint/utf-8 items._

PR #1480, opened 2026-03-20
feat: add warning-aware import diagnostics

scarf005:
## Summary
- add warning severity plumbing to `deno_lint` diagnostics and example reporting
- make `no-import-prefix` and `no-unversioned-import` warnings with lockfile-aware `deno.json(c)` autofixes
- avoid guessed versions and handle overlapping fixes when both warnings trigger

Related: denoland/deno#30681

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -348,6 +348,7 @@ dependencies = [
  "env_logger",
  "globwalk",
  "if_chain",
+ "jsonc-parser",
  "log",
  "once_cell",
  "os_pipe",
@@ -796,6 +797,15 @@ name = "itoa"
 version = "1.0.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4a5f13b858c8d314ee3e8f639011f7ccefe71f97f96e50151fb991f267928e2c"
+
+[[package]]
+name = "jsonc-parser"
+version = "0.31.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "45157f76d8fbe134e0c97403d69ac48989a1e0249a058b20f9f5eeff11bb09f0"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "libc"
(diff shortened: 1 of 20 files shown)


bartlomieju:
**Verdict: WIP (draft) — needs a rebase onto the merged severity work before it can land.** Not merging. Confined mostly to new rules + `examples/dlint` plus diagnostic plumbing.

**Biggest item: the severity plumbing is now duplicated/superseded**
This PR re-adds its own `LintDiagnosticSeverity { Error, Warning }` enum plus `diagnostic_level()`/`highlight_style()`/`severity()` in `src/diagnostic.rs`. But the target branch **already merged** `LintDiagnostic::severity` (#1487): `src/diagnostic.rs` now defines `LintDiagnosticSeverity` (Error/Warning) with `as_str()`, serde round-trip, the `severity` field defaulting to `Error`, and the same `DiagnosticLevel`/highlight-style mapping — all with tests. As-is this PR will hard-conflict and produce duplicate definitions. The plumbing half should be dropped and rebased to build on the existing enum + `severity` field, keeping only what's genuinely new (the `add_warning_with_fixes`-style ergonomics if the merged API doesn't already provide them). The wide churn across `jsx_*`, `no_window*`, `no_process_global`, etc. will also need reconciling against the merged changes.

**The genuinely new, valuable part: the two import rules**
- `no-unversioned-import` + `no-import-prefix` as *warnings* with lockfile-aware `deno.json(c)` autofixes via `fix_with_deno_config_package`. The "avoid guessed versions / handle overlapping fixes when both trigger" design is the right instinct — overlapping fixes are exactly where autofixers corrupt code, so good that it's handled.
- Integration tests spin up real temp `deno.json`/`deno.lock` projects (unique dirs via pid + `AtomicUsize`) — solid coverage. Minor: I don't see temp-dir cleanup, so runs may leave `deno_lint_no_unversioned_import_*` dirs in `$TMPDIR`; add teardown (or use a scope-guard) to avoid litter.

**To finish (draft)**
- Rebase onto #1487's severity API (the main blocker).
- New rules → maintainer product decision; add `docs/rules/*.md` for them.
- Confirm `add_warning_with_fixes` aligns with whatever the merged per-rule config/severity work (#1487/#1534) exposes, rather than a parallel path.

**Perf:** nothing concerning in the rule paths (per-import checks).

_Foundation is reasonable; the work now is mostly reconciling with the severity infrastructure that landed after this was opened._

PR #1482, opened 2026-04-01
feat: add aarch64 linux build support

lu-zero:
Add ubuntu-24.04-arm to CI matrix to produce native aarch64 binaries. This follows the same pattern as deno's CI setup.

Generated by Mistral Vibe.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        os: [macos-15-intel, ubuntu-latest, windows-latest]
+        os: [macos-15-intel, ubuntu-latest, windows-latest, ubuntu-24.04-arm]
 
     env:
       CARGO_INCREMENTAL: 0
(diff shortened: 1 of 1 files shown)


bartlomieju:
**Verdict: LGTM — clean CI change. Recommend merge once the new ARM job is confirmed green** (not auto-merging: CI status is `UNKNOWN`, and a workflow change should be validated by an actual run).

**Correctness — looks right**
- Adds `ubuntu-24.04-arm` (a valid GitHub-hosted ARM runner) to the matrix and builds natively, so no cross-compile target install is needed — correct approach.
- Nicely guards the existing x86_64 packaging step with `&& !contains(matrix.os, 'arm')` so the arm runner doesn't also produce the x86_64 zip, and adds a dedicated `Pre-release (linux-aarch64)` step emitting `dlint-aarch64-unknown-linux-gnu.zip`.
- Adds that artifact to the release `files:` list, consistent with the existing per-job upload pattern (each job uploads only the file it produced).

**Minor (non-blocking)**
- Other steps gated on `contains(matrix.os, 'ubuntu')` — the wasm up-to-date check, the `dlint rules --json` vs `docs.json` diff, and the benchmarks — will now *also* run on the ARM job (since its label contains "ubuntu"). That's harmless duplicate work, and arguably useful coverage, but if you want to avoid the extra runtime you could scope those to `!contains(matrix.os, 'arm')` too.

**Verify before merge**
- It's a CI workflow change (and AI-generated per the description), so the meaningful validation is the ARM job actually passing on this PR. `mergeStateStatus: UNKNOWN` means there's no confirmed green run yet — a maintainer should trigger/check CI, then it's a straightforward merge.

**Perf:** adds one matrix job (more CI minutes) — expected for a new target.

_No changes requested; just confirm the `ubuntu-24.04-arm` job goes green, then merge._

Issue #1274, opened 2024-05-12
Add LintDiagnostic::severity


bartlomieju:
Other popular linters like ESLint allow to customize the severity of diagnostics - either "warning" or "error". Currently `deno_lint` treats all diagnostics all "error".

To allow for severity customization there are two major parts that need to happen:
1. Add `severity` field to `LintDiagnostic` struct so it can be properly formatted during printing
2. Add configuration options that would allow to customize the severity of each lint rule.

To handle point 1 following things need to happen:
1. Add `LintDiagnosticSeverity` enum like so:
```rust
#[derive(Debug, Default, Clone, Copy)]
pub enum LintDiagnosticSeverity {
  #[default]
  Error,
  Warning,
}
```
2. Update `LintDiagnostic` struct and add `severity: LintDiagnosticSeverity` field to it
3. Update necessary APIs to accept `severity: LintDiagnosticSeverity` argument in following order:
- `Context::create_diagnostic`
- `Context::add_diagnostic_with_fixes`
- `Context::add_diagnostic_with_hint`
- `Context::add_diagnostic`
4. Update `LintDiagnostic::level()` to convert `LintDiagnosticSeverity` into `deno_ast::DiagnosticLevel`.

These changes should be enough to be able to apply different formatting based on severity and can be first applied in `examples/dlint/diagnostics.rs` in `print_compact` and `print_pretty`.

Handling point 2 is more involved and requires further discussion as the current lint configuration is rather bare-bones - it allows to specify a list of lints to run, but not customize them one-by-one.

If someone picks up this issue I can provide more details on how to handle that, but mind that it require changes across several repositories - `deno_lint`, `deno_config` and `deno` itself.

eryue0220:
Could I have a try for this one if no one take it?

bartlomieju:
@eryue0220 sorry for a slow response. This is still up for grabs if you want to tackle it.

PR #1487, opened 2026-04-25
feat: add LintDiagnostic::severity

MukundaKatta:
## Summary
Refs #1274. Adds a `severity` field on `LintDiagnostic` plus a `LintDiagnosticSeverity` enum (`Error` default, `Warning`).

## Files
- `src/diagnostic.rs` — added `LintDiagnosticSeverity` enum, `severity` field on `LintDiagnostic`, severity-aware `Diagnostic::level()` and snippet highlight style, plus 5 unit tests.
- `src/context.rs` — `create_diagnostic` populates `severity: Default::default()` (Error) for backwards compat; new `add_diagnostic_details_with_severity` opt-in API.
- `examples/dlint/diagnostics.rs` — compact formatter prints actual severity label instead of hardcoded "Error".

## Backwards compatibility
Default severity preserves the existing "all errors" behavior. New field added without removing or renaming any API. Existing rules unchanged.

## Test plan
- `cargo test` — 373 lib + 3 example pass, 0 fail.
- New tests cover default-is-Error, JSON round-trip (lowercase), `level()` mapping, and label strings.
- `cargo clippy --all-targets -- -D warnings` clean.

## Out of scope
The issue's "point 2" (per-rule config customization) crosses repos and was left for further discussion per the issue author.

### DIFF
--- a/examples/dlint/diagnostics.rs
+++ b/examples/dlint/diagnostics.rs
@@ -16,15 +16,26 @@ pub fn display_diagnostics(
 
 fn print_compact(diagnostics: &[LintDiagnostic]) {
   for diagnostic in diagnostics {
+    // Capitalize the severity label ("error"/"warning" -> "Error"/"Warning")
+    // so the compact output keeps its existing leading-cap style.
+    let severity_label = {
+      let s = diagnostic.severity.as_str();
+      let mut chars = s.chars();
+      match chars.next() {
+        Some(c) => c.to_uppercase().collect::<String>() + chars.as_str(),
+        None => String::new(),
+      }
+    };
     match &diagnostic.range {
       Some(range) => {
         let display_index =
           range.text_info.line_and_column_display(range.range.start);
         eprintln!(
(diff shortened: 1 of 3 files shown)


bartlomieju:
Thanks for this @MukundaKatta — picking this as the implementation for #1274 over the parallel #1496 (now closed), since this one is more complete (it also maps the snippet highlight style to `Warning` and updates the dlint compact formatter).

I pushed one small follow-up commit folding in the cleanups from #1496 (`maintainerCanModify` was on):

- routed construction through a single `create_diagnostic_with_severity` so `add_diagnostic_details_with_severity` no longer builds-then-mutates;
- expanded the `LintDiagnosticSeverity` doc to explain why severity lives on the diagnostic instance rather than on the rule's details.

Everything else is unchanged. `cargo test` (lib + examples), `cargo clippy --all-targets`, and `cargo fmt` are all clean locally. Looks good to land once CI re-runs.

bartlomieju (approved):
Thanks

Issue #1449, opened 2025-07-26
Rule `no-unused-vars` doesn’t detect unused JSDoc `@import` types


jaydenseric:
### Lint Name

`no-unused-vars`

### Code Snippet

```js
// @ts-check

/** @import { ReactNode } from "react" */
```

### Expected Result

There should be a lint error that the type import `ReactNode` is unused.

### Actual Result

No lint error.

### Additional Info

In a `.ts` module, this:

```ts
import type { ReactNode } from "react";
```

Results in this lint error:

```
`ReactNode` is never used
If this is intentional, alias it with an underscore like `ReactNode as _ReactNode` deno-lint(no-unused-vars)
```

So it has already been established that `no-unused-vars` is supposed to enforce unused type imports.

Setting the TypeScript config `compilerOptions.noUnusedLocals` to `true` causes a TypeScript error about unused JSDoc `@import` types, but I would like to not have to set that and be able to rely on Deno lint.

### Version

deno 2.4.2 (stable, release, aarch64-apple-darwin)
v8 13.7.152.14-rusty
typescript 5.8.3

PR #1492, opened 2026-05-30
no-unused-vars: also collect references in doc comments

### LINKED ISSUE #1472: bug: false positive of `no-unused-vars` for symbols used in TSDoc comments
### Lint Name

`no-unused-vars`

### Code Snippet

```ts
import type { Foo } from "./file.ts";

/**
 * Related to {@link Foo.bar | this property}
 */
const baz = 42;
```

### Expected Result

The symbol `Foo` is needed in order to make the TSDoc comment of `baz` valid, so it is used. I expect `deno_lint` to see this, and to exit with status code 0.

### Actual Result

`no-unused-vars` is reported

### Version

deno 2.6.5 (stable, release, x86_64-unknown-linux-gnu)
v8 14.2.231.17-rusty
typescript 5.9.2

71:
This also introduces a second variant of the diagnostic specifically for things referred to from doc comments, but from nowhere else, which makes the logic a bit more complex, but IMO is worth it, as TypeScript won't suggest changing imports to `type` imports if it doesn't detect any use for a symbol.

Admittedly this doesn't really belong in `no-unused-vars`, but it just felt unfortunate to now accept to have non-`type` imports for things only used in doc comments. Still, I'd be fine removing this from this PR.

Fixes #1472, #1449, https://github.com/denoland/deno/issues/27583.

I also tried to keep the code fairly similar to what was already there without introducing new allocations. For instance, instead of using `SyntaxContext::empty()` in `Id` I could have made the `HashSet<Id>` a `HashMap<Atom, HashSet<Option<SyntaxContext>>>` or `HashMap<(Atom, Option<SyntaxContext>)>`, but this felt unnecessarily different.

### DIFF
--- a/src/rules/no_unused_vars.rs
+++ b/src/rules/no_unused_vars.rs
@@ -2,6 +2,7 @@
 
 use super::program_ref;
 use super::{Context, LintRule};
+use crate::diagnostic::{LintFix, LintFixChange};
 use crate::tags::{self, Tags};
 use crate::Program;
 use crate::ProgramRef;
@@ -18,6 +19,7 @@ use deno_ast::swc::ast::{
   TsTypeQueryExpr, TsTypeRef, VarDecl, VarDeclarator,
 };
 use deno_ast::swc::ast::{Id, SimpleAssignTarget};
+use deno_ast::swc::common::SyntaxContext;
 use deno_ast::swc::ecma_visit::{Visit, VisitWith};
 use deno_ast::swc::utils::find_pat_ids;
 use deno_ast::view::AssignOp;
@@ -27,6 +29,7 @@ use if_chain::if_chain;
 use std::collections::HashSet;
 use std::iter;
 use std::rc::Rc;
+use std::sync::LazyLock;
 
 #[derive(Debug)]
 pub struct NoUnusedVars;
@@ -37,6 +40,8 @@ const CODE: &str = "no-unused-vars";
 enum NoUnusedVarsMessage {
   #[display(fmt = "`{}` is never used", _0)]
   NeverUsed(String),
+  #[display(fmt = "`{}` is never used outside of doc comments", _0)]
+  NeverUsedOutsideDocComments(String),
 }
 
 #[derive(Display)]
@@ -80,6 +85,18 @@ impl LintRule for NoUnusedVars {
     match program {
       ProgramRef::Module(m) => m.visit_with(&mut collector),
       ProgramRef::Script(s) => s.visit_with(&mut collector),
+    }
+
+    // Collect symbols referenced in doc comments. Instead of going through all leading comments we
+    // could use `context.leading_comments_at(node.start())` for all declaration nodes in
+    // `Collector`, but that would require us to add handling for many node types, so instead we go
+    // through all comments in the program at once.
+    //
+    // The downside to this approach is that we don't have a `SyntaxContext` for each symbol, but
+    // in any case not all nodes that can have leading comments have a `SyntaxContext`, so we'd need
+    // to handle empty-`SyntaxContext` references anyway.
+    for comment in context.comments().leading_map().values().flatten() {
+      collect_if_doc_comment(&mut collector.used_types, comment);
     }
 
     let mut visitor = NoUnusedVarVisitor::new(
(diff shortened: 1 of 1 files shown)


bartlomieju:
Took a closer look — the approach is sound and the core fix works: an `import type { Foo }` referenced only via `{@link Foo}` was previously a false positive, and the new tests confirm it's now correctly suppressed. The hand-rolled brace matcher looks safe (offset strictly increases so the loop terminates, and byte-indexing only happens on ASCII branch chars so no mid-UTF8 slicing). Tests pass and clippy is clean.

A few things worth raising before merge:

- **Default and namespace imports aren't covered.** Only `visit_import_named_specifier` was rewired to consult doc references (via `handle_named_import`). `visit_import_default_specifier` and `visit_import_star_as_specifier` still check `used_types.contains(&local.to_id())` with a real `SyntaxContext`, which never matches the empty-ctxt doc entries. So `import Foo from "./x"` used only in `{@link Foo}` is still falsely flagged. Worth either handling or documenting as a known limitation.

- **Doc references are file-global with an empty `SyntaxContext`**, so any `{@link Bar}` suppresses the unused check for a named import `Bar` regardless of scope. Fine for module-scoped named imports, and over-suppression is the safe direction, but worth a note (the PR body already explains the tradeoff).

- **`NeverUsedOutsideDocComments` is opinionated** — as you noted yourself. It's gated correctly (`is_type_only` / `in_type_only_context` / `media_type().is_typed()`) and isn't a regression (these were already flagged as plain `NeverUsed`), but it's a maintainer call whether the value→`type` nudge belongs in `no-unused-vars`.

Nothing blocking from a correctness standpoint — the main thing I'd want resolved is the default/namespace-import gap.

bartlomieju:
**Verdict: needs changes** — the doc-comment link parser has a panic. Not merging (rule-behavior change + bug). Nice feature otherwise, and the effort to avoid extra allocations is appreciated.

**Correctness — blocking panic**
In `collect_doc_comment_link`, the word branch:
```rust
let end = text[offset..]
  .find(|ch: char| !ch.is_alphanumeric())
  .unwrap_or(text.len());          // BUG
...
let word = &text[offset..offset + end];
```
`str::find` returns an index **relative to `text[offset..]`**, but the `None` fallback uses `text.len()` (the full length). When the identifier runs to the end of the comment text (no trailing delimiter), `end == text.len()` and `text[offset..offset + end]` slices past the end → **panic (byte index out of bounds)**.

Concrete repro — `/** {@linkcode Foo*/`: `comment.text` is `* {@linkcode Foo` (len 16), the match ends at offset 13 (`Foo`), `find` returns `None`, so `end = 16` and the code slices `text[13..29]` on a 16-byte string → panic. A lint rule must never panic on valid (even if malformed-JSDoc) input.
Fix: `.unwrap_or(text.len() - offset)` (or `.unwrap_or(text[offset..].len())`). Add a test with a link identifier immediately before `*/`.

**Correctness — otherwise looks sound**
- UTF-8 safety is OK elsewhere: offset only advances by 1 on matched **ASCII** bytes and by char-boundary word lengths, and the word slice uses a char-predicate `find`, so boundaries hold (once the fallback bug above is fixed).
- The `find_at` loop can't spin: `collect_doc_comment_link` returns an offset past `m.end()`, so the next search starts beyond the current match.
- `handle_named_import` logic (used-in-type vs used-only-in-doc-comment via the `SyntaxContext::empty()` sentinel, then suggesting `type` import only in typed files / non-type-only imports) reads correctly, and the `in_type_only_context` save/restore around `visit_import_decl` is right.

**Design (maintainer call)**
- As you note, doc-comment awareness is a bit outside `no-unused-vars`' remit, and it introduces a second diagnostic variant. That's a product decision — flagging it so a maintainer weighs keeping it here vs. splitting it out.
- Name-only matching means two different symbols sharing a name can mask each other (inherent to link syntax). Acceptable, but worth a doc note.

**Perf:** good — single pass over leading comments, linear parser, no new allocations in the hot path. `LazyLock` regex compiled once (confirm MSRV ≥ 1.80 for `std::sync::LazyLock`).

**Housekeeping:** `mergeable: UNKNOWN` → rebase.

_Blocker: fix the `unwrap_or(text.len())` slice bug (+ regression test). Then maintainer sign-off on scope._

bartlomieju:
Follow-up to the review above: here's a ready-to-apply patch for the out-of-bounds **panic** in `collect_doc_comment_link`. `str::find` returns an index relative to `text[offset..]`, but the `None` fallback used `text.len()` (the full length), so when a `{@link}` identifier runs to the end of the comment text the slice `text[offset..offset + end]` goes past the end and panics. Repro: `/** {@linkcode Foo*/`.

```diff
--- a/src/rules/no_unused_vars.rs
+++ b/src/rules/no_unused_vars.rs
@@ fn collect_doc_comment_link(
         // Try to parse a word.
         let end = text[offset..]
           .find(|ch: char| !ch.is_alphanumeric())
-          .unwrap_or(text.len());
+          .unwrap_or(text.len() - offset);
```

Plus a regression test (add to the `no_unused_vars_valid` `assert_lint_ok!` block):

```rust
// Regression test: a link identifier that runs to the very end of the
// comment text (no trailing delimiter before `*/`) must not panic.
r#"
import type { Foo } from "./dummy.ts";
/** {@linkcode Foo*/
export class Baz {}
      "#,
```

`cargo test --lib no_unused_vars` passes with this change. (I couldn't push directly since this fork has "Allow edits from maintainers" disabled — feel free to apply it, or enable that setting and it can be pushed for you.)

KnorpelSenf:
@71 are you still interested in fixing this up, or should someone attempt to fix this?

PR #1493, opened 2026-06-01
fix(no-undef): recognize DOM globals

### LINKED ISSUE #27379: lint(no-undef): says `document` is undefined
Version: Deno 2.1.4

The lint rule `no-undef` https://docs.deno.com/lint/rules/no-undef/ says that it disallows undeclared variables, but I wouldn't expect this to apply to globally available variables like `document`.

### deno.json
```json
{
  "compilerOptions": {
    "lib": ["dom", "dom.iterable", "dom.asynciterable", "deno.ns"]
  },
  "lint": {
    "rules": {
      "include": ["no-undef"]
    }
  }
}
```
### main.ts
```ts
console.log(document);
```
### terminal
```
error[no-undef]: document is not defined
 --> /Users/soul/Projects/lint/main.ts:1:13
  |
1 | console.log(document);
  |             ^^^^^^^^

  docs: https://lint.deno.land/rules/no-undef


Found 1 problem
Checked 1 file
```

Tatlatat:
## Summary

The `no-undef` rule reports browser/DOM globals such as `document` and `HTMLElement` as `… is not defined`, even when a project's `compilerOptions.lib` includes `"dom"`. This is because the rule only checks the `GLOBALS` list, which contains the Deno runtime globals but no DOM-specific ones.

Repro (from denoland/deno#27379):

```jsonc
// deno.json
{
  "compilerOptions": { "lib": ["dom", "dom.iterable", "dom.asynciterable", "deno.ns"] },
  "lint": { "rules": { "include": ["no-undef"] } }
}
```

```ts
// main.ts
console.log(document); // error[no-undef]: document is not defined  ← should not error
```

## Changes

- Add a separate `DOM_GLOBALS` list in `src/globals.rs`, kept distinct from the Deno-runtime `GLOBALS` so the two sets stay clearly grouped.
- Consult `DOM_GLOBALS` in `no_undef`'s `check()` (via `GLOBALS.iter().chain(DOM_GLOBALS)`).
- Overly generic identifiers (`Text`, `Range`, `Comment`, `Option`, `Image`, `Selection`, `Attr`, `Node`) are intentionally **excluded** to avoid masking genuine typos in non-browser code.
- Add positive tests (`document`, `HTMLElement`, `MutationObserver`, `requestAnimationFrame`) and negative regression tests (`documentt`, `HTMLElementt` still error).

## Notes / scope

Deriving the active globals from `compilerOptions.lib` would be a more principled long-term solution, but the existing `GLOBALS` are already matched unconditionally, so this PR follows the same approach as an incremental fix and leaves lib-based detection as future work.

`cargo test`, `cargo clippy`, and `cargo fmt --check` all pass.

Fixes denoland/deno#27379


bartlomieju:
**Verdict: LGTM on the code — needs maintainer sign-off** (behavior change to `no-undef`) + a small perf suggestion. Not auto-merging (rule-behavior change).

**Correctness — good**
- `GLOBALS.iter().chain(DOM_GLOBALS).any(...)` is correct, and the bool field is (rightly) ignored by `no-undef`. Positive tests (`document`, `HTMLElement`, `MutationObserver`, `requestAnimationFrame`) and negative regression tests (`documentt`, `HTMLElementt` still error) cover both directions well.
- Excluding generic names (`Text`, `Range`, `Comment`, `Option`, `Node`, `Image`, `Selection`, `Attr`) is good judgment — those would mask real typos in non-DOM code.

**Design trade-offs for a maintainer to weigh**
- DOM globals are matched **unconditionally**, so a non-DOM Deno project will now also *not* flag `document`/`HTMLElement` (false negatives). That's the flip side of fixing the reported false positive; you correctly note the existing `GLOBALS` are already unconditional, so it's consistent — but it does mean `no-undef` no longer catches stray DOM refs in server-only code. The lib-based derivation you mention as future work is the principled fix.
- The generic-name exclusions mean the fix is deliberately incomplete: a DOM project using `Node`, `Text`, or `Range` will still get false positives. Reasonable, but worth documenting the known gap so it's not reported again.

**Performance — worth addressing while touching this**
- `check()` runs `GLOBALS.iter().chain(DOM_GLOBALS).any(...)` **per identifier reference**, i.e. a linear scan over ~340 entries (up from ~250) for every ident in the file → O(idents x 340) string comparisons. The linear pattern pre-existed, but the list just grew ~35%. Consider building a single `HashSet<&'static str>` (or a perfect-hash via `phf`) once and doing O(1) lookups; that removes the hot-path cost for both `GLOBALS` and `DOM_GLOBALS`.

**Housekeeping**
- `DOM_GLOBALS` all use `false` for the second field — fine since `no-undef` ignores it, but if `DOM_GLOBALS` is ever consumed by `no-global-assign`, double-check that value is intended (DOM globals aren't reassignable, so `false` looks right).
- `mergeable: UNKNOWN` → rebase.

_No code changes required to land it; blockers are the maintainer decision on unconditional-vs-lib-based globals + a rebase. The `HashSet` swap is an optional but timely perf win._

PR #1494, opened 2026-06-02
refactor: port to oxc

nathanwhit:
based on https://github.com/denoland/deno_ast/pull/331


bartlomieju:
**Verdict: WIP (draft) — large, well-structured port; blocked on an upstream dependency.** Not merging. This needs dedicated human review per rule-cluster with a green test suite; it's not a squash-and-go, and an automated pass can't meaningfully line-review 84 rule ports. Notes at the architecture level:

**Hard blocker: unpublished dependency**
- `Cargo.toml` points `deno_ast` at a **git branch of a personal fork** (`git = "https://github.com/nathanwhit/deno_ast", branch = "oxc-port"`, based on deno_ast#331). This can't land until that deno_ast work merges and ships a real crates.io release, then this PR pins that version. Treat as a stacked PR gated on deno_ast#331. (`mergeStateStatus: BLOCKED`.)
- Also bumps `rust-toolchain` 1.90 → 1.96 — fine, just flag it as a required-toolchain bump for contributors/CI.

**Scope / approach — looks sound**
- Feature set changes from `["scopes","transforms","utils","view","react"]` to `["scopes","semantic","visit"]`: the entire `ast_view::Handler`/`NodeTrait` traversal layer is replaced by oxc's `Visit`. `handler.rs` includes a clear SWC→oxc node-rename map, which is the right way to make 84 rule ports reviewable.

**Where reviewers/CI should focus (highest regression risk)**
- **Behavior parity is the whole ballgame.** The rule test corpus (valid/invalid + exact `col`/range assertions) is the safety net — the merge gate should be the full suite green, with spot-checks that diagnostic *ranges* still match, since swc byte-offsets vs oxc spans can differ subtly.
- **Unified node kinds**: `ClassDecl`/`ClassExpr` → `Class` and `FnDecl`/`FnExpr` → `Function`. Rules that branched on declaration-vs-expression (e.g. hoisting/name checks) are the most likely to regress; worth targeted review.
- **Scope/semantic swap** (`scopes` → `semantic`): binding-resolution-dependent rules (`no-undef`, `no-unused-vars`, `prefer-const`, `no-global-assign`, `prefer-primordials`) can shift behavior if global/scope resolution differs.
- **Comment handling** (JSDoc/ignore-directive rules) if oxc surfaces comments/trivia differently.

**Perf:** oxc is generally faster, but validate with the existing `benchmarks/` to confirm no regression on large files.

_Excellent direction and cleanly organized. Blockers: land+publish deno_ast#331, then rebase/pin; after that, review rule-cluster by rule-cluster against a green test suite rather than as one diff._

PR #1495, opened 2026-06-17
fix: accept both 'handler' and 'handlers' exports in Fresh routes

d-jch:
## Problem

The `fresh-handler-export` linter rule incorrectly flags `handlers` (plural) as an error in Fresh route files. However, Fresh framework actually supports both `handler` and `handlers` exports.

## Evidence

From Fresh source code ([fs_routes.ts](https://github.com/denoland/fresh/blob/main/packages/fresh/src/fs_routes.ts)):

```
Route files must export a default component, a "handler" or "handlers" export, or a "config" export.
```

## Fix

This PR removes the check that flags `handlers` as invalid, allowing both conventions as Fresh intended.

## Changes

- `src/rules/fresh_handler_export.rs`: Remove the diagnostic for `handlers` export
- Update tests to show both `handler` and `handlers` are valid in `routes/`

## Testing

The existing tests have been updated to reflect that both exports are valid.

### DIFF
--- a/src/rules/fresh_handler_export.rs
+++ b/src/rules/fresh_handler_export.rs
@@ -4,16 +4,12 @@ use super::{Context, LintRule};
 use crate::handler::{Handler, Traverse};
 use crate::tags::{self, Tags};
 
-use deno_ast::view::{Decl, Pat, Program};
-use deno_ast::SourceRanged;
+use deno_ast::view::Program;
 
 #[derive(Debug)]
 pub struct FreshHandlerExport;
 
 const CODE: &str = "fresh-handler-export";
-const MESSAGE: &str =
-  "Fresh middlewares must be exported as \"handler\" but got \"handlers\" instead.";
-const HINT: &str = "Did you mean \"handler\"?";
 
 impl LintRule for FreshHandlerExport {
   fn tags(&self) -> Tags {
@@ -38,33 +34,9 @@ struct Visitor;
 impl Handler for Visitor {
   fn export_decl(
     &mut self,
-    export_decl: &deno_ast::view::ExportDecl,
-    ctx: &mut Context,
+    _export_decl: &deno_ast::view::ExportDecl,
+    _ctx: &mut Context,
   ) {
-    // Fresh only considers components in the routes/ folder to be
-    // server components.
-    let Some(mut path_segments) = ctx.specifier().path_segments() else {
-      return;
-    };
-    if !path_segments.any(|part| part == "routes") {
-      return;
-    }
-
-    let id = match export_decl.decl {
-      Decl::Var(var_decl) => {
-        if let Some(first) = var_decl.decls.first() {
-          let Pat::Ident(name_ident) = first.name else {
-            return;
-          };
-          name_ident.id
-        } else {
-          return;
-        }
-      }
-      Decl::Fn(fn_decl) => fn_decl.ident,
-      _ => return,
-    };
-
     // Fresh accepts both "handler" and "handlers" exports
     // No diagnostic needed - both are valid
   }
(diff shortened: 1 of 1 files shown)


bartlomieju:
Thanks for digging into this! The motivation is correct, but I don't think the implementation should land in its current shape.

**The premise checks out for Fresh 2.x.** I confirmed against Fresh's source:
- `main` (2.x): `const handlers = mod.handlers ?? mod.handler ?? null;` — both exports are accepted and treated identically, so flagging `handlers` is a genuine false positive now.
- Fresh 1.x only reads `route.handler` (singular), which is exactly the footgun this rule was added (#1173) to catch. So the rule *was* correct for 1.x; Fresh 2 changed the contract.

So dropping the check is reasonable. The problem is **how** it's dropped — the PR turns the rule into a registered-but-dead no-op:

- The rule still exists in `get_all_rules_raw()` and carries the `[fresh]` tag, but `export_decl` is now empty, so it can never emit a diagnostic. `Visitor` still traverses the whole AST to do nothing.
- The new tests assert nothing: a no-op rule passes `assert_lint_ok!` for *any* input, so the three new cases would pass even if the rule were deleted (or fed garbage). They no longer protect any behavior.
- The docs are now wrong and untouched — `fresh-handler-export.md` in `denoland/docs` still lists `export const handlers = {}` as **Invalid**.

**Suggested shape:** if `handlers` is valid, the rule has no remaining purpose, so remove it entirely rather than stub it:
1. Delete `src/rules/fresh_handler_export.rs`
2. Remove the `pub mod fresh_handler_export;` and `Box::new(...)` registration in `src/rules.rs`
3. Update/remove the docs page in `denoland/docs`

One thing worth confirming before deleting: existing `deno.json` configs may reference `fresh-handler-export` in `lint.rules.include`/`exclude` or in `deno-lint-ignore` directives. If unknown rule codes aren't handled gracefully, a deprecation path may be preferable to a hard delete. A maintainer should weigh in on delete-vs-deprecate.

d-jch:
Thanks for the thorough review @bartlomieju! Agree the stub approach is wrong — the right fix is either a full removal or a proper deprecation path, which is a maintainer-level call.

I'll close this PR for now. Happy to help with the implementation once the path is decided.

d-jch:
One more thought before this goes away: instead of deleting the rule entirely, what about flipping the check to a whitelist?

The current rule was designed for Fresh 1.x where only `handler` (singular) was valid, so it flagged `handlers` (plural) as the footgun. But Fresh 2 accepts **both** `handler` and `handlers` — and importantly, it still rejects everything else (`const handlers = mod.handlers ?? mod.handler ?? null`).

So the rule still has value, just with updated logic:

| Export name | Old rule (1.x) | No-op stub (this PR) | Whitelist approach |
|---|---|---|---|
| `handler` | ✅ | ✅ | ✅ |
| `handlers` | ❌ (false positive) | ✅ | ✅ |
| `foo` / anything else | ✅ (missed) | ✅ (missed) | ❌ diagnostic |

The check becomes: if the export is a `handler`-like declaration in `routes/` and the name is **neither** `handler` nor `handlers`, emit a diagnostic.

Roughly:
```rust
if !id.sym().eq("handler") && !id.sym().eq("handlers") {
    ctx.add_diagnostic_with_hint(id.range(), CODE, MESSAGE, HINT);
}
```

The message/hint would need updating too — something like `"Fresh route handlers must be exported as \"handler\" or \"handlers\""`.

This way Fresh 2 users get correct validation, and the rule stays useful rather than being deleted. @bartlomieju what do you think?

d-jch:
Replaced by #1516 — same goal (stop false-positives on `handlers`), but with a whitelist approach instead of a no-op stub, so that unrecognized export names are still flagged.

PR #1496, opened 2026-06-28
feat: add severity field to LintDiagnostic

bartlomieju:
Adds a per-diagnostic severity concept to the crate. A new `LintDiagnosticSeverity` enum (`Error`, `Warning`) and a `severity` field on `LintDiagnostic` replace the previously hardcoded `DiagnosticLevel::Error` in the `Diagnostic::level()` implementation, which now maps the severity onto `deno_ast`'s `DiagnosticLevel`.

The change is backward compatible: every existing diagnostic construction path defaults to `Error`, so current behavior is unchanged. Rules and external consumers can now emit a warning-level diagnostic via the new `Context::add_diagnostic_details_with_severity` method or by setting the `severity` field on a `LintDiagnostic` directly.

This only exposes the capability in the crate. Wiring user-facing configuration (so `deno.json` can mark a rule as a warning) lives in the deno CLI and is intentionally out of scope here.

Towards #1274.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,8 +2,8 @@
 
 use crate::control_flow::ControlFlow;
 use crate::diagnostic::{
-  LintDiagnostic, LintDiagnosticDetails, LintDiagnosticRange, LintDocsUrl,
-  LintFix,
+  LintDiagnostic, LintDiagnosticDetails, LintDiagnosticRange,
+  LintDiagnosticSeverity, LintDocsUrl, LintFix,
 };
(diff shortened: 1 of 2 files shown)


bartlomieju:
Closing in favor of #1487, which implements the same feature and is a bit more complete (it also maps the snippet highlight style to Warning and updates the dlint compact formatter). I'll fold the small cleanups from this PR into #1487. Thanks @MukundaKatta for getting there first.

PR #1500, opened 2026-06-28
feat: add jsx-a11y accessibility lint rules (content/attribute cluster)

bartlomieju:
First cluster of accessibility rules ported from oxlint / eslint-plugin-jsx-a11y, implemented in Rust against `deno_ast::view` using the existing `Handler`/`Traverse` pattern.

Four rules are added:

- `jsx-a11y-anchor-has-content` — an `<a>` anchor must have content accessible to screen readers (text, a non-hidden child, or `dangerouslySetInnerHTML` / `title` / `aria-label`).
- `jsx-a11y-heading-has-content` — heading elements `<h1>`–`<h6>` must have content.
- `jsx-a11y-html-has-lang` — the `<html>` element must have a non-empty `lang` attribute.
- `jsx-a11y-iframe-has-title` — `<iframe>` must have a non-empty `title` attribute.

A new `a11y` tag is introduced. These rules are **opt-in**: they are tagged `a11y`, `jsx`, `react`, `fresh` but intentionally **not** `recommended`, so they only run when a user enables them. Individual rules can be promoted to `recommended` later.

The configurable `components` option (custom-component name mapping) and the autofixers from the reference implementations are intentionally deferred for now; only built-in DOM element names are handled. Test fixtures are ported from the oxlint reference (pass cases → `assert_lint_ok!`, fail cases → `assert_lint_err!`).

This is the start of the jsx-a11y porting track (35 rules total); see the project notes for the full plan.

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -26,6 +26,10 @@ pub mod fresh_handler_export;
 pub mod fresh_server_event_handlers;
 pub mod getter_return;
 pub mod guard_for_in;
+pub mod jsx_a11y_anchor_has_content;
+pub mod jsx_a11y_heading_has_content;
+pub mod jsx_a11y_html_has_lang;
+pub mod jsx_a11y_iframe_has_title;
 pub mod jsx_boolean_value;
 pub mod jsx_button_has_type;
 pub mod jsx_curly_braces;
@@ -269,6 +273,10 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(fresh_server_event_handlers::FreshServerEventHandlers),
     Box::new(getter_return::GetterReturn),
     Box::new(guard_for_in::GuardForIn),
+    Box::new(jsx_a11y_anchor_has_content::JSXA11yAnchorHasContent),
+    Box::new(jsx_a11y_heading_has_content::JSXA11yHeadingHasContent),
+    Box::new(jsx_a11y_html_has_lang::JSXA11yHtmlHasLang),
+    Box::new(jsx_a11y_iframe_has_title::JSXA11yIframeHasTitle),
     Box::new(jsx_boolean_value::JSXBooleanValue),
     Box::new(jsx_button_has_type::JSXButtonHasType),
     Box::new(jsx_curly_braces::JSXCurlyBraces),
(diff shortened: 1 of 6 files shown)


bartlomieju:
**Verdict: LGTM on the code — needs maintainer sign-off** (new rule cluster) + a structural cleanup suggestion. Not auto-merging (new rules).

Clean port of 4 jsx-a11y rules using the standard `Handler`/`Traverse` pattern, correctly **opt-in** (`a11y`/`jsx`/`react`/`fresh` tags, not `recommended`). Logic and tests read as faithful to the oxlint/eslint-plugin-jsx-a11y reference.

**Correctness — verified**
- `anchor-has-content`: the child/attr/aria-hidden logic matches the reference. I checked the whitespace edge (`<a>\n</a>` → `!text.value().is_empty()` treats whitespace text as content, so it's *not* flagged) — this **matches eslint-plugin-jsx-a11y's `Boolean(value)` behavior**, so it's faithful, not a bug. If you ever want stricter behavior, `.trim().is_empty()` would flag whitespace-only anchors, but keeping parity with the reference is the right call here.
- `expr_to_boolean` / `is_hidden_from_screen_reader` correctly handle `aria-hidden` with bare/string/expr values and `input type="hidden"`. Good test coverage ported from oxlint (pass→ok, fail→err).

**Structural — worth doing now, before the track grows**
- **Helper duplication.** `get_attr_ignore_case`, `is_hidden_from_screen_reader`, `expr_to_boolean`, and `object_has_accessible_child` are copy-pasted verbatim between `jsx_a11y_anchor_has_content.rs` and `jsx_a11y_heading_has_content.rs` (and the attr helpers recur in the html/iframe rules). Since this is rule 1–4 of a stated 35-rule port, extract these into a shared `jsx_a11y/util.rs` (or `swc_util`) module now — otherwise the duplication multiplies across 35 files and drift between copies becomes a real risk.

**Minor**
- Expression-container children like `{''}`, `{0}`, `{false}` fall through to `_ => true` and count as "content". That also mirrors the reference's leniency; just noting it in case a stricter interpretation is wanted later.

**Perf:** fine — per-`jsx_element` visits with linear attr scans; no concerns.

**Housekeeping:** new `a11y` tag + 4 opt-in rules is a product decision (your own PR, so presumably intended). `mergeable: UNKNOWN` → rebase.

_No behavioral changes requested; recommend extracting the shared helpers before landing so the rest of the a11y track builds on a common util._

PR #1505, opened 2026-06-28
feat: add jsx-no-conflicting-pragmas rule

### LINKED ISSUE #969: Lint proper use of jsx pragma
```js
/** @jsxImportSource https://esm.sh/preact */
/** @jsx h */
/** @jsxFragment Fragment */
```

The above code works fine (it uses the `@jsxImportSource`) as expected. It is not obvious to users what should happen here though, so we should add a `deno lint` to disallow this. One or the other, not both.

_Originally posted by @lucacasonato in https://github.com/denoland/deno/issues/12631#issuecomment-962707286_

bartlomieju:
Adds a rule that flags mixing the automatic-runtime pragma
`@jsxImportSource` with the classic-runtime pragmas `@jsx` /
`@jsxFragment` (or `@jsxFrag`) in the same file:

```js
/** @jsxImportSource https://esm.sh/preact */
/** @jsx h */
/** @jsxFragment Fragment */
```

When `@jsxImportSource` is set the file uses the automatic runtime, so the
`@jsx` and `@jsxFragment` pragmas are silently ignored. Having both is a
footgun — you should use one runtime or the other. The rule reports each
classic pragma when an `@jsxImportSource` pragma is also present, and only
looks at the file's leading comments, where pragmas actually take effect.

The rule is tagged `react`/`jsx` but intentionally left out of
`recommended`, so it must be enabled explicitly. Happy to promote it to
`recommended` if preferred — flagging the decision rather than making it.

Closes #969.

### DIFF
--- a/schemas/rules.v1.json
+++ b/schemas/rules.v1.json
@@ -25,6 +25,7 @@
     "jsx-key",
     "jsx-no-children-prop",
     "jsx-no-comment-text-nodes",
+    "jsx-no-conflicting-pragmas",
     "jsx-no-danger-with-children",
     "jsx-no-duplicate-props",
     "jsx-no-unescaped-entities",
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -32,6 +32,7 @@ pub mod jsx_curly_braces;
 pub mod jsx_key;
 pub mod jsx_no_children_prop;
 pub mod jsx_no_comment_text_nodes;
+pub mod jsx_no_conflicting_pragmas;
 pub mod jsx_no_duplicate_props;
 pub mod jsx_no_unescaped_entities;
 pub mod jsx_no_useless_fragment;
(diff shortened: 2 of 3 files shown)


bartlomieju:
**Verdict: LGTM — correct and well-tested. Recommend merge** after resolving the schema conflict + rebase (new rule → maintainer product decision, which you've explicitly flagged). Not auto-merging (new rule).

**Correctness — verified the regex boundaries (the crux of this rule)**
I simulated both patterns against the tricky inputs; all behave exactly as documented:
- `@jsxImportSource\b` matches `@jsxImportSource` but **not** `@jsxImportSourceTypes` (the `T` after `Source` blocks the `\b`) — so a types-only directive doesn't switch the runtime. ✓
- `@jsx(Frag(ment)?)?\b` matches `@jsx`, `@jsxFrag`, `@jsxFragment`, but **not** `@jsxImportSource`, `@jsxRuntime`, or `@jsxFragmentFoo` (word-boundary correctly rejects longer identifiers). ✓
- Logic is order-independent and only inspects comments before the first statement (where pragmas actually take effect); comment-only files fall back to scanning all comments. All of this is covered by the valid/invalid tests, including the multi-pragma and reversed-order cases.

**Perf:** fine — two `Lazy`-compiled regex `is_match` calls per leading comment; leading comments are few.

**Minor**
- A single comment holding both `@jsxImportSource` and a classic pragma gets flagged as a whole (range = full comment). Reasonable; just noting the range granularity.
- No `docs/rules/jsx_no_conflicting_pragmas.md` / `docs()` in the diff — new rules conventionally ship docs, and the docs.json CI diff expects them.

**Housekeeping / blockers**
- Edits `schemas/rules.v1.json`, which PR #1377 deletes — these will conflict; coordinate ordering and drop the schema edit if #1377 lands first.
- `mergeable: UNKNOWN` → rebase.
- The `recommended` promotion is yours to call — you've flagged it; opt-in is a fine default.

_No code changes requested. Solid rule; just docs + the schema-file conflict + rebase before landing._

Issue #1244, opened 2024-02-02
Tag All Rules


barziahmed:
Hello, 

I want to tag and use all rules (excluding Fresh) for my project, not just the recommend rules. How do I do that? I don't want to write each rule's name in "include" configuration. 

This is my current deno linter .json configuraiton:
```
	"lint": {
		"rules": {
			"tags": ["recommended"]
		}
	},
...
```
Can I do something like this? 	"tags": ["all"]?

Great linter and great project by the way.

bartlomieju:
I can recognize the need here, but I'm not sure about the exact semantics. `"all"` would imply all rules, so how would we exclude just `"fresh"` rules by default?

Would something like this work for you?

```
{
  "lint": {
     "rules": {
       "tags": ["all", "-fresh"]
     }
}
```

l-cornelius-dol:
I have a specific need for this. I want all linting options enabled with tailored excludes to match one of five linting levels for our code: "all", "strict", "normal", "lax", "none" (none will have only minimal specific includes, and all will have minimal specific excludes; both are for very exceptional use cases). Our team has lots of code created over three decades, and the goal is to keep EOL projects from getting worse, improve good projects, and be able to ensure new projects have the highest code quality.

I very much want to turn all rules on by default, so that I encounter new rules as they are added. Then I can decided at what level that new rule belongs.

Currently I am periodically, manually, copying the list from the Rust source file, and dropping it in my linting configs. It's quite tedious.

bartlomieju:
@l-cornelius-dol thanks for the comment - so would the solution from https://github.com/denoland/deno_lint/issues/1244#issuecomment-2103633950 work for you?

l-cornelius-dol:
> @l-cornelius-dol thanks for the comment - so would the solution from [#1244 (comment)](https://github.com/denoland/deno_lint/issues/1244#issuecomment-2103633950) work for you?

Yes, it would. In fact "all" was the intuitive thing I tried; when that failed I searched the code and concluded that it only supports "recommended".

NB: I can't stand Node and it's dependency/supply-chain hellscape, so I have gone all-in on Deno for our build-support system. Therefore, in our case at least, linting is not about what's right or even recommended for Deno code, but what's the right balance for our code-base, both current and up to two decades' old. I sure do appreciate the work you-all do over there.

temeddix:
It would be very nice to be able to tag all rules. Adding rules one by one manually in `lint.rules.include` of `deno.json` is not a very clean solution.

PR #1514, opened 2026-06-28
feat: support the `all` tag to enable every rule

bartlomieju:
Refs #1244.

Adds a special `all` tag to `filtered_rules`: when the requested tag set
contains `all`, every rule matches regardless of its own tags. This lets a
configuration opt into the complete rule set with

```jsonc
{ "lint": { "rules": { "tags": ["all"] } } }
```

instead of enumerating every rule under `include`. It composes with the
existing `exclude` list to drop individual rules, e.g. `tags: ["all"]` +
`exclude: ["no-explicit-any"]`.

This is the deno_lint-side primitive only. Two related ideas from the issue
are intentionally out of scope here and can build on top of this:

- tag-level negation (`tags: ["all", "-fresh"]`) — needs the config layer in
  the Deno CLI to map the negated tag onto excludes;
- user-defined custom tags.

Adds tests for `all` alone (returns every rule) and `all` + `exclude`.

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -185,7 +185,8 @@ pub fn recommended_rules(
 ///
 /// - if `maybe_tags` is `None` then all defined rules are returned, otherwise
 ///   only rules matching at least one tag will be returned; if provided list
-///   is empty then all rules will be excluded by default
+///   is empty then all rules will be excluded by default. The special tag
+///   `all` matches every rule regardless of the rule's own tags.
 ///
 /// - if `maybe_exclude` is `Some`, all rules with matching codes will
 ///   be filtered out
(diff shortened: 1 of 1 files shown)


bartlomieju:
**Verdict: LGTM — correct and tested. Recommend merge** (public-behavior change, so flagging rather than auto-merging; needs green CI / rebase).

**Correctness — good**
- `tags_set.contains("all") || rule.tags().iter().any(...)` short-circuits to include every rule when `all` is present, and composes correctly with `exclude` (applied downstream) — both paths are covered by the added tests (`all` → `get_all_rules().len()`, and `all` + exclude → `len() - 1` with the excluded code absent).
- `contains("all")` on the `HashSet<String>` resolves via `Borrow<str>`, so no allocation for the check. The `|| any(...)` ordering also means the common `all` case avoids the per-tag scan.

**Minor**
- `all` is now effectively a reserved tag name. No current rule uses it, but worth a one-line note in the doc comment that `all` is reserved, so a future rule doesn't accidentally define a real `all` tag.
- Scope is correctly limited to the deno_lint primitive; tag-negation (`-fresh`) and custom tags are appropriately deferred to the CLI config layer, as you note.

**Perf:** neutral/slightly better (short-circuit on `all`).

**Housekeeping:** public API/behavior addition → maintainer sign-off (your own PR). `mergeable: UNKNOWN` → rebase.

_No changes requested. Good to merge once CI is green._

PR #1516, opened 2026-06-29
fix(fresh-handler-export): whitelist handler and handlers instead of blacklisting handlers

d-jch:
## Summary

Updates the `fresh-handler-export` rule from a blacklist approach (flagging only `handlers`) to a whitelist approach (flagging anything that is NOT `handler` or `handlers`).

Closes #1495

## Motivation

Fresh 2.x accepts **both** `handler` and `handlers` as valid route handler export names. From the Fresh source:

```
const handlers = mod.handlers ?? mod.handler ?? null;
```

The old rule was designed for Fresh 1.x where only `handler` (singular) was recognized — exporting `handlers` (plural) was a silent footgun. Fresh 2 changed the contract, so the previous behavior of flagging `handlers` is now a **false positive**.

However, simply deleting the rule or turning it into a no-op would miss the fact that any _other_ export name is still silently ignored by Fresh. The rule retains value by checking that the export name is one of the two recognized names.

## Changes

| Export name | Old behavior | New behavior |
|---|---|---|
| `handler` | ✅ | ✅ |
| `handlers` | ❌ false positive | ✅ |
| `foo` / anything else | ✅ missed | ❌ diagnostic |

- **Logic**: Changed from `if name == "handlers"` → error to `if name != "handler" && name != "handlers"` → error
- **Message**: Updated to `"Fresh route handlers must be exported as \"handler\" or \"handlers\"."`
- **Hint**: Updated to `"Expected \"handler\" or \"handlers\"."`
- **Tests**: Added `assert_lint_ok!` for `handlers` in `routes/`, added `assert_lint_err!` for unknown names (`foo`, `bar`, `baz`)

### DIFF
--- a/src/rules/fresh_handler_export.rs
+++ b/src/rules/fresh_handler_export.rs
@@ -12,8 +12,8 @@ pub struct FreshHandlerExport;
 
 const CODE: &str = "fresh-handler-export";
 const MESSAGE: &str =
-  "Fresh middlewares must be exported as \"handler\" but got \"handlers\" instead.";
-const HINT: &str = "Did you mean \"handler\"?";
+  "Fresh route handlers must be exported as \"handler\" or \"handlers\".";
+const HINT: &str = "Expected \"handler\" or \"handlers\".";
 
 impl LintRule for FreshHandlerExport {
   fn tags(&self) -> Tags {
@@ -65,8 +65,8 @@ impl Handler for Visitor {
       _ => return,
     };
 
-    // Fresh middleware handler must be exported as "handler" not "handlers"
-    if id.sym().eq("handlers") {
+    // Fresh only accepts "handler" or "handlers" as route handler exports
+    if !id.sym().eq("handler") && !id.sym().eq("handlers") {
       ctx.add_diagnostic_with_hint(id.range(), CODE, MESSAGE, HINT);
     }
   }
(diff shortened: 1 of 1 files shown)


src/rules/fresh_handler_export.rs
```diff
@@ -65,8 +65,8 @@ impl Handler for Visitor {
       _ => return,
     };
 
-    // Fresh middleware handler must be exported as "handler" not "handlers"
-    if id.sym().eq("handlers") {
+    // Fresh only accepts "handler" or "handlers" as route handler exports
+    if !id.sym().eq("handler") && !id.sym().eq("handlers") {
```
bartlomieju, line 69: This is the crux of the issue. Fresh route files legitimately export `config` (`RouteConfig`) in addition to `handler`/`handlers` — it's in the same contract this PR cites. With this whitelist, `export const config = {}` (a `Decl::Var`) now gets flagged as an error, which is a false positive in valid Fresh routes.

Beyond `config`, this also flags *any* helper export from a route module (e.g. `export const MAX = 50;` or `export function format() {}`), which Fresh silently ignores rather than treating as a footgun. At minimum `config` needs to be in the allowed set:

```rust
const ALLOWED: &[&str] = &["handler", "handlers", "config"];
// ...
let sym = id.sym();
if !ALLOWED.iter().any(|name| sym.eq(*name)) {
  ctx.add_diagnostic_with_hint(id.range(), CODE, MESSAGE, HINT);
}
```

src/rules/fresh_handler_export.rs
```diff
@@ -120,19 +120,37 @@ mod tests {
...
+      r#"export function handlers() {}"#,
+    );
+    assert_lint_ok!(
+      FreshHandlerExport,
+      filename: "file:///routes/index.tsx",
+      r#"export async function handlers() {}"#,
+    );
+
+    // Unknown export names should be flagged
+    assert_lint_err!(FreshHandlerExport, filename: "file:///routes/index.tsx",  r#"export const foo = {}"#: [
```
bartlomieju, line 141: These new cases are a real improvement over the no-op stub from #1495 (they actually exercise the logic), and the column offsets look correct.

Missing, though, is a guard against the regression noted above — there's no `assert_lint_ok!` proving a valid Fresh route export like `config` is *not* flagged. Please add:

```rust
assert_lint_ok!(
  FreshHandlerExport,
  filename: "file:///routes/index.tsx",
  r#"export const config = {}"#,
);
```

This would currently fail and demonstrates the false positive.

src/rules/fresh_handler_export.rs
```diff
@@ -12,8 +12,8 @@ pub struct FreshHandlerExport;
 
 const CODE: &str = "fresh-handler-export";
 const MESSAGE: &str =
-  "Fresh middlewares must be exported as \"handler\" but got \"handlers\" instead.";
-const HINT: &str = "Did you mean \"handler\"?";
+  "Fresh route handlers must be exported as \"handler\" or \"handlers\".";
```
bartlomieju, line 15: Message/hint wording is clearer than before. Minor: since the rule now fires on arbitrary unrecognized exports (not just a `handler`→`handlers` near-miss), the phrasing "must be exported as handler or handlers" can be confusing when it lands on, say, `export const config`. If the scope stays this broad, consider wording that conveys "this export name isn't recognized by Fresh" rather than implying the user intended a handler.

bartlomieju (commented):
Thanks for the follow-up to #1495 — the core premise is right: Fresh 2.x accepts both `handler` and `handlers` (`mod.handlers ?? mod.handler`), so flagging `handlers` was a genuine false positive.

My main concern is that the whitelist as written trades one false positive for another. Fresh's route contract (quoted in #1495) is: *"a default component, a `handler` or `handlers` export, **or a `config` export**."* The whitelist only allows the two handler names, so `export const config = {}` — valid in every Fresh app — now errors. More broadly, a whitelist turns this rule into "routes/ may not contain any unrecognized named `const`/`function` export," which also flags legitimate helper exports that Fresh simply ignores. That's a meaningfully more opinionated rule than the targeted typo-check this used to be.

Left inline notes. Also note the predecessor PR was closed pending a maintainer call on whitelist-vs-delete/deprecate — worth settling that direction before iterating further. Whichever way it lands, the `denoland/docs` page for `fresh-handler-export` still lists `export const handlers = {}` as Invalid and needs updating.

d-jch:
Thanks for the detailed review — I appreciate the insight. You're right that the whitelist approach is more opinionated than the targeted typo-check this rule was designed as, and `config` plus helper exports like `MAX` / `format` aren't footguns worth flagging.

I'll step back and let you decide the direction here. Feel free to close this PR if the whitelist approach doesn't fit, or keep it open if you'd rather iterate further.

bartlomieju:
Thanks for digging into this, @d-jch! I've opened #1519 which builds on your whitelist approach and addresses the review: it adds `config` (`RouteConfig`) to the recognized set so valid routes aren't flagged, and rewords the message/hint to convey "Fresh doesn't recognize this export." You're credited as co-author. Closing this in favor of #1519.

Issue #622, opened 2021-02-23
Globals should be derived from parsing lib files...


kitsonk:
Instead of maintaining a hand crafted list of global variables, deno_lint should take either a list generated externally when the API is called, or parse a set of provided API files when the API is called.

With the generated list, the Deno CLI (or other consumer) would generate a list of global symbol names and pass in based on the ambient configuration.

The other option would be that Deno CLI (or other consumer) would supply a list of `lib.*.d.ts` files that are being applied to the program/files to be linted, which should be parsed so that their global symbols are extracted.

Issue #590, opened 2021-01-15
Document is not defined


ebebbington:
Possible to use any tsconfig with `deno lint`, or any other way to bypass these errors?

```
$ cat my_file.ts

...
const prevBody = document.body.children.length;
...

$ deno lint --unstable my_file.ts

(no-undef) document is not defined
    const prevBody = document.body.children.length;

```

magurotuna:
As far as I know, there's no way of configuring what variables are defined globally.
Currently the list of global variables is constant, being managed in [this file](https://github.com/denoland/deno_lint/blob/master/src/globals.rs). This list doesn't contain `document`, so `no-undef` error was triggered in your example.

I'm not really sure `document` should be listed here, but I guess it should.
@bartlomieju what do you think about it?

ebebbington:
Thanks for the response :) I was curious if there's a way about it, currently i am just slapping `// deno-lint-ignore no-undef` on, and if it helps, my use case is I have code that talks to the DOM (not client side code im trying to lint),and this is exposed via the modules API

So linting is a big part of the project,  i guess it isn't 'breaking', but it would be really nice to have :) Though saying that, it seems like `document` is something that shouldn't be a thing thats enabled by default so a. `--config` flag. could work well maybe?

But the above is just my opinion

bartlomieju:
I'm leaning towards accepting more configuration in Rust API for the linter so we can synchronize list of globals based on the `lib` value in `tsconfig.json` users are using. I guess we need to use more of the available sets from https://www.npmjs.com/package/globals and add proper API to the `Linter`; then we can handle that in `deno` to sniff out what `lib`s user wishes to use.

kitsonk:
We need to find a way to integrate the linter with TypeScript in the long term so that tsc can be queried and get type information. Maintaining lists of global variables is intenrsincly a bad idea, as any number of things can be changed which impact the global variables.

Issue #555, opened 2020-12-04
Triple Slash Directives ignored in Linter


jcc10:
The following code triggers a linter error:
```
/// <reference lib="es2020" />
/// <reference lib="dom" />
/// <reference no-default-lib="true"/>
console.log(document);
```
outputs
```
$ deno --unstable lint test.ts
(no-undef) document is not defined
console.log(document);
            ^^^^^^^^
    at /home/digitalkitsune/github/immersive-recording/livestream-subview-controller/test.ts:4:12

Found 1 problem
Checked 1 file
```
This indicates that the linter is not working.

I should note that there is no issues with the bundling
```
$ deno bundle ./test.ts
Bundle file:///home/digitalkitsune/github/immersive-recording/livestream-subview-controller/test.ts
Check file:///home/digitalkitsune/github/immersive-recording/livestream-subview-controller/test.ts
console.log(document);
```
However, removing the 1 or more of the triple slash directives does cause the bundling to fail.

PR #1517, opened 2026-06-29
feat(no-undef, no-global-assign): support host-supplied globals

bartlomieju:
The no-undef and no-global-assign rules consult a single hand-maintained
GLOBALS list, so DOM globals such as document and HTMLElement are reported
as "not defined" even when a project's compilerOptions.lib includes "dom",
and there is no way to make either rule reflect the environment the project
actually configured.

This adds an optional globals field to LintConfig: a name -> writable map
that, when present, fully replaces the built-in GLOBALS list for both rules.
The intent is that the host (deno) derives this set from the resolved
TypeScript lib and passes it in, so deno_lint itself carries no hardcoded
per-lib tables and never drifts against a TypeScript version. When the field
is absent the built-in list is used, so default behavior is unchanged. Both
rules now resolve globals through Context::global_with_writable /
Context::is_global, which preserves the writable flag that no-global-assign
relies on (onmessage may be reassigned, Object may not).

A follow-up commit converts the built-in GLOBALS list from a linear-scanned
slice to a compile-time phf::Map, so both the built-in and host-supplied
lookups are O(1) per identifier instead of a scan over ~200 entries. The list
is module-private, so this is not a public API change. It also drops a
duplicate TransformStream entry that was harmless for the linear scan but
that phf rejects as a duplicate key.

This is the deno_lint half of the change; a deno-side follow-up will compute
the set from compilerOptions.lib and wire it through LintConfig.globals. It
supersedes #1493, which instead adds a second, unconditional DOM_GLOBALS
list -- that approach weakens no-undef for every project regardless of its
lib configuration, whereas this keeps the recognized globals tied to what
the project configured.

Refs #622, #590, #555, #1287, denoland/deno#27379


bartlomieju:
**Verdict: LGTM — correct, well-tested, good design + a perf win. Recommend merge** (public API + behavior change → maintainer sign-off; `mergeStateStatus: BLOCKED`, so not auto-merging).

This is the principled version of the DOM-globals problem: the host derives the recognized set from `compilerOptions.lib` and passes it via `LintConfig::globals`, instead of hardcoding an unconditional DOM list. It correctly supersedes #1493 (which I reviewed earlier and flagged for exactly the false-negative-in-non-DOM-projects issue and the linear-scan perf — both addressed here). **Recommend closing #1493 as superseded.**

**Correctness — verified**
- `Context::global_with_writable` resolves host globals when present, else built-in `GLOBALS`, preserving the writability bit that `no-global-assign` depends on. `is_global` = `.is_some()`. Clean.
- `no_global_assign` faithfully keeps the semantics: `if let Some(writable) = ...; if !writable { flag }` — read-only globals can't be reassigned, writable ones can. The new `no_global_assign_configured_globals` test covers read-only (`document`), writable (`onmessage`), and full-replacement (`Object` not in the supplied set → allowed).
- The `GLOBALS` → `phf::Map` conversion is behavior-preserving; dropping the duplicate `TransformStream` key is correct (harmless under linear scan, required by phf). Module-private, so no public API change from that part.

**Performance — improvement**
- O(1) phf lookup per identifier replaces the ~200-entry linear scan for both built-in and host-supplied paths. This is the hot-path win I suggested on #1493 — nice to see it here for real.

**One caution for the deno-side follow-up (out of scope here)**
- Since a host-supplied set **fully replaces** the built-in list, the deno side must include core JS globals (`Object`, `Array`, `globalThis`, …) in addition to lib-derived ones, or `no-undef` will false-positive on them. Intentional per the design, just worth guarding in the CLI wiring.

**Housekeeping:** public `LintConfig::globals` addition + behavior change → sign-off. `mergeable: MERGEABLE` but `BLOCKED` (likely awaiting review/CI).

_No changes requested. Solid; recommend merging (and closing #1493)._

PR #1518, opened 2026-06-29
feat(ignore-directives): support block-scoped deno-lint-ignore

### LINKED ISSUE #476: Ignore directives should affect the entire next AST node
As in, the following should be valid:
```ts
// deno-lint-ignore no-explicit-any
{
  let a: any;
  let b: any;
  let c: any;
}
```

This is taken from Rust, `#[rustfmt::skip]` and `#[allow(warnings)]` work this way. `// deno-fmt-ignore` already works this way also. It's much a much nicer way of handling ignore ranges than `// deno-lint-ignore-start` etc.

bartlomieju:
A `// deno-lint-ignore` directive only suppressed diagnostics on the
single line that follows it, so silencing a rule across a multi-line
span meant repeating the directive on every line or reaching for the
much coarser `deno-lint-ignore-file`.

A directive placed directly above a `{ ... }` block now suppresses
diagnostics anywhere inside that block. Coverage is intentionally
limited to bare, explicit blocks: the directive has to be a leading
comment of a `BlockStmt`. A directive placed before a function, class
or other statement attaches to that node rather than to a block, so it
keeps the ordinary next-line behavior and does not silently swallow
deeply nested diagnostics, which was the main concern raised when this
was discussed.

Closes #476

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -247,6 +247,8 @@ impl<'a> Context<'a> {
       };
 
       let diagnostic_line = range.text_info.line_index(range.range.start);
+
+      // Ordinary directive: `// deno-lint-ignore` on the line directly above.
       if diagnostic_line > 0 {
         if let Some(l) =
           self.line_ignore_directives.get_mut(&(diagnostic_line - 1))
@@ -255,6 +257,25 @@ impl<'a> Context<'a> {
             continue;
           }
         }
+      }
+
+      // Block-scoped directive: `// deno-lint-ignore` before a `{ ... }` block
+      // suppresses diagnostics anywhere inside that block.
+      let mut ignored_by_block = false;
+      for (comment_line, directive) in self.line_ignore_directives.iter_mut() {
+        if let Some(block_end_line) = directive.block_end_line() {
+          let block_start_line = comment_line + 1;
+          if block_start_line <= diagnostic_line
+            && diagnostic_line <= block_end_line
+            && directive.check_used(&diagnostic.details.code)
+          {
+            ignored_by_block = true;
+            break;
+          }
+        }
+      }
+      if ignored_by_block {
+        continue;
       }
 
       filtered.push(diagnostic);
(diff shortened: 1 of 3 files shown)


bartlomieju:
**Verdict: LGTM — correct and well-tested, including the key safety guard. Recommend merge** (ignore-directive behavior change → maintainer sign-off; needs green CI / rebase).

**Correctness — verified the main risk**
- The over-suppression concern (a directive before `if (x) {`/`function f() {` swallowing the whole body) is handled correctly *and tested*: because comments attach to the nearest following token, `leading_comments(block_range.start)` only returns a directive that sits directly above a **bare** `{`. The test `function foo(): any { let a: any }` yields `block_end_line() == None`, and the behavioral test confirms `debugger` inside a function body is still reported. So block scope only applies to explicit bare blocks — exactly the intent.
- The containment check (`block_start_line <= diagnostic_line <= block_end_line` + `check_used`) is right, and it runs only after the ordinary next-line path, so existing behavior is unchanged when no bare block is involved.
- `extend_block_ignore_directives` is correctly gated on `!directives.is_empty()`, so files without directives pay nothing.

**Performance — minor**
- The block-directive check in the diagnostic filter is an O(block-directives) linear scan **per diagnostic**, i.e. O(diagnostics x block-directives) worst case. Negligible for normal files (few directives), but if you expect files with many ignore directives + many diagnostics, consider indexing block directives by line range. Not a blocker.
- `extend_block_ignore_directives` walks the full AST once per file-with-directives — acceptable as a one-time per-file cost.

**Suggestion**
- Add an `if`/`for`/`while` body case to the "blocks only" tests (function is covered; the attachment mechanism is the same, but explicit coverage would lock in that `if (x) {` bodies aren't block-scoped).

**Housekeeping:** `mergeable: UNKNOWN` → rebase. Behavior change to ignore-directives → your call to sign off (your own PR).

_No changes requested. Well-scoped feature with the right guardrails and tests; good to merge once CI is green._

PR #1521, opened 2026-06-29
feat(erasing-op): add `erasing-op` rule

bartlomieju:
Ports oxlint's `oxc/erasing-op` rule (itself based on Clippy's `erasing_op`)
to a native Rust rule. It flags binary operations whose result is always
zero, which is almost always a mistake rather than intent:

- `x * 0` and `0 * x`
- `x & 0` and `0 & x`
- `0 / x`

It deliberately does not flag `0 / 0` (evaluates to `NaN`) or `x / 0`
(`Infinity` / `NaN`), since neither is zero. Parenthesized zeros such as
`x * (0)` are unwrapped and still flagged.

The rule reuses the existing `Handler`/`bin_expr` visitor and reports the
whole binary expression. Tests are ported from the oxc rule's own pass/fail
fixtures, plus a few extra valid cases that lock the operator scope (`+`,
`-`, `|` are not erasing) and the `0 / 0` / `x / 0` carve-outs.

This is the first of the oxc-correctness oxlint ports tracked in the lint
roadmap. One open decision for a maintainer: it is tagged `recommended`
(oxc classifies it as `correctness`), which makes it on-by-default for every
`deno lint` user. Happy to drop the `recommended` tag if you would rather
land it opt-in first.

Reference: https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/oxc/erasing_op.rs

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -19,6 +19,7 @@ pub mod camelcase;
 pub mod constructor_super;
 pub mod default_param_last;
 pub mod eqeqeq;
+pub mod erasing_op;
 pub mod explicit_function_return_type;
 pub mod explicit_module_boundary_types;
 pub mod for_direction;
@@ -263,6 +264,7 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(constructor_super::ConstructorSuper),
     Box::new(default_param_last::DefaultParamLast),
     Box::new(eqeqeq::Eqeqeq),
+    Box::new(erasing_op::ErasingOp),
     Box::new(explicit_function_return_type::ExplicitFunctionReturnType),
     Box::new(explicit_module_boundary_types::ExplicitModuleBoundaryTypes),
     Box::new(for_direction::ForDirection),
--- a/src/rules/erasing_op.rs
+++ b/src/rules/erasing_op.rs
@@ -1,0 +1,165 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+use super::{Context, LintRule};
+use crate::handler::{Handler, Traverse};
+use crate::tags::{self, Tags};
+use crate::Program;
+use deno_ast::swc::ast::BinaryOp;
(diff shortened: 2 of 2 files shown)


bartlomieju:
**Verdict: LGTM on the logic — needs the RECOMMENDED sign-off** you flagged + a rebase. Not auto-merging (new rule).

Faithful port of oxc/Clippy `erasing_op`.

**Correctness — verified**
- `Mul`/`BitAnd`: `is_zero(left) || is_zero(right)` correctly catches `x*0`, `0*x`, `x&0`, `0&x`.
- `Div`: `is_zero(left) && !is_zero(right)` flags `0/x` but correctly exempts the literal `0/0` (NaN) and `x/0` (Infinity/NaN). The valid tests lock these carve-outs and the non-erasing operators (`+`, `-`, `|`).
- `is_zero` unwraps `ParenExpr` recursively, so `x * (0)` is flagged; and it matches only `Lit::Num == 0.0`, so `+`/`-`/`|` and non-zero literals are untouched.

**Minor / faithful-to-reference edges (non-blocking)**
- Only a numeric literal `0` counts: `-0` (a `UnaryExpr`, not a `Num`) and BigInt `0n` (`Lit::BigInt`) are not flagged. That matches the reference and is reasonable (`x * 0n` would TypeError on mixed operands anyway) — just noting.
- The message "will always evaluate to zero" is slightly imprecise for `0 / x` when `x` is `0` at runtime (`NaN`); the rule intentionally flags the pattern as suspicious regardless, matching oxc. Could soften the wording, but fine as-is.

**Perf:** trivial — one `is_zero` check per `bin_expr`.

**Housekeeping:** `RECOMMENDED`/on-by-default is the maintainer call you raised (oxc classifies it `correctness`, so on-by-default is defensible). `mergeable: UNKNOWN` → rebase; a `docs/rules/*.md` isn't in the diff (new rules conventionally ship one).

_Logic is solid and well-tested. Blockers are just the tagging decision + rebase (+ docs)._

PR #1525, opened 2026-06-29
feat(prefer-set-size): add `prefer-set-size` rule

bartlomieju:
This ports the `prefer-set-size` rule from oxlint's unicorn plugin as a native
Rust rule.

The rule flags a `.length` member access on an array produced by converting a
`Set` to an array, and recommends using `Set#size` instead. It catches both
spread conversions (`[...set].length`) and `Array.from(set).length`, where the
`Set` is either a direct `new Set(...)` or a `const`-declared variable
initialized to `new Set(...)`.

It deliberately does not flag: optional chaining (`?.length`), computed access
(`[length]` / `["length"]`), non-`length` properties, `new NotSet(...)`, bare
`Set(...)` without `new`, `let`/`var`-declared or destructured set variables,
`Array.from(set, mapFn)` (a second map-fn argument), `Array?.from` /
`Array.from?.()`, and `NotArray.from`.

The pass/fail fixtures are ported verbatim from the oxc reference rule:
https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/unicorn/prefer_set_size.rs
(the autofix-only fixtures are dropped since deno_lint has no autofixer here).

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -121,6 +121,7 @@ pub mod prefer_ascii;
 pub mod prefer_const;
 pub mod prefer_namespace_keyword;
 pub mod prefer_primordials;
+pub mod prefer_set_size;
 pub mod react_no_danger;
 pub mod react_no_danger_with_children;
 pub mod react_rules_of_hooks;
@@ -369,6 +370,7 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(prefer_const::PreferConst),
     Box::new(prefer_namespace_keyword::PreferNamespaceKeyword),
     Box::new(prefer_primordials::PreferPrimordials),
+    Box::new(prefer_set_size::PreferSetSize),
     Box::new(react_no_danger::ReactNoDanger),
     Box::new(react_no_danger_with_children::ReactNoDangerWithChildren),
     Box::new(react_rules_of_hooks::ReactRulesOfHooks),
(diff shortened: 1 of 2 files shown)


bartlomieju:

Careful two-pass port (collect `const` Set vars → check `.length`). I verified the tricky parts:

**Correctness — verified**
- **Soundness via `const`-only:** only `const x = new Set(...)` with a simple `Ident` binding is collected, so the variable can't be reassigned to a non-Set. `let`/`var`/destructured bindings are (correctly) not flagged. Right call.
- **Hygiene:** `set_vars` is keyed by `Id` (`Atom` + `SyntaxContext`), so a `const set` in one scope won't false-match a same-named binding in another. Good.
- **Exclusions all hold:** `?.length` (parent is `OptChainExpr`), computed `[length]`/`["length"]`, non-`length` props, `new NotSet()`, bare `Set()` w/o `new`, `Array.from(set, mapFn)` (`is_array_from_call` requires `args.len() == 1`), `Array?.from`/`Array.from?.()` (not a plain `Expr::Member`/`Expr::Call`), and multi-element arrays (`[...set, x].length`) — each is correctly rejected and covered by valid tests.
- `get_set_node` requires a **single spread** element for the array form (`[set].length` with a non-spread element is not flagged), and `unwrap_parens` is applied at both the object and set-node levels, so `[...(set)].length` still resolves.

**Perf:** two full traversals (collector + handler), each O(nodes), with O(1) `HashSet` lookups — fine for a per-file pass.

_No code changes requested — the const-only soundness and the exclusion set are exactly right. Just tagging + rebase._

PR #1526, opened 2026-06-29
feat(no-await-in-promise-methods): add `no-await-in-promise-methods` rule

bartlomieju:
This ports the `no-await-in-promise-methods` rule from oxlint's unicorn
plugin to a native Rust rule.

The rule flags an `await` expression used directly as an element of the
array literal passed to `Promise.all([...])`, `Promise.allSettled([...])`,
`Promise.any([...])`, and `Promise.race([...])`. Awaiting a promise before
handing it to one of these methods defeats their purpose: the methods are
designed to accept pending promises and coordinate them concurrently, so
awaiting an element forces it to settle sequentially before the method even
runs. For example, `Promise.all([await foo()])` is almost always a mistake.

Matching mirrors oxc's behavior exactly: the callee object must be the
`Promise` identifier, the property must be a non-computed identifier naming
one of the four methods, the call must have exactly one argument, and that
argument must be an array literal (not a spread). Each array element that is
directly an `await` expression is reported; spread elements are ignored.

Reference:
https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/unicorn/no_await_in_promise_methods.rs

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -40,6 +40,7 @@ pub mod jsx_void_dom_elements_no_children;
 pub mod no_array_constructor;
 pub mod no_async_promise_executor;
 pub mod no_await_in_loop;
+pub mod no_await_in_promise_methods;
 pub mod no_await_in_sync_fn;
 pub mod no_boolean_literal_for_arguments;
 pub mod no_case_declarations;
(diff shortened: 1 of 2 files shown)


bartlomieju:

Faithful port; flags a direct `await` element inside `Promise.{all,allSettled,any,race}([...])`.

**Correctness — verified**
- Callee gating is exact: `Promise[all]` / `Promise["all"]` (computed → `MemberProp::Ident` fails), `Promise.all?.()` / `Promise?.all()` (optional chain → not a plain `Member` callee), `NotPromise.all`, `Promise.notListedMethod`, and `new Promise.all(...)` (a `NewExpr`, so `call_expr` isn't invoked) are all correctly excluded — each has a valid test.
- Arg gating: exactly one non-spread arg that is an array literal (`Promise.all(...[await p])`, `Promise.all([await p], extra)`, `Promise.all()`, `Promise.all(notArray)` all skipped).
- Element handling: `elems.iter().flatten()` skips elisions (`[,]`), spread elements are skipped, and only a **direct** `Expr::Await` element is flagged — so `[(await p, 0)]` (sequence, last value `0`) and `[foo(await p)]` (await nested in a call) are correctly not flagged.
- `globalThis.Promise.all([await p])` is not flagged (obj is a `Member`, not the `Promise` ident) — matches oxc's name-based approach; noting it as an intentional false-negative.

**Perf:** trivial — per `call_expr`, single pass over array elements.

_No code changes requested — edge handling matches the reference exactly. Just tagging + rebase._

PR #1527, opened 2026-06-29
feat(no-useless-fallback-in-spread): add `no-useless-fallback-in-spread` rule

bartlomieju:
This ports the unicorn `no-useless-fallback-in-spread` rule from oxlint as a
native deno_lint rule.

The rule flags a useless empty-object fallback inside an object spread, such as
`{ ...(foo || {}) }` or `{ ...(foo ?? {}) }`. Spreading a nullish (or otherwise
falsy) value into an object literal is a no-op, so providing an empty object
`{}` as a fallback via `||` or `??` adds nothing and can be removed. The rule
only triggers when the spread is inside an object literal (not an array literal
or call arguments), the logical operator is `||` or `??`, and the right operand
is an empty object literal, unwrapping any surrounding parentheses to match
oxc's behavior.

Reference implementation:
https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/unicorn/no_useless_fallback_in_spread.rs

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -110,6 +110,7 @@ pub mod no_unsafe_negation;
 pub mod no_unused_labels;
 pub mod no_unused_vars;
 pub mod no_unversioned_import;
+pub mod no_useless_fallback_in_spread;
 pub mod no_useless_rename;
 pub mod no_var;
 pub mod no_window;
@@ -358,6 +359,7 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(no_unused_labels::NoUnusedLabels),
     Box::new(no_unused_vars::NoUnusedVars),
     Box::new(no_unversioned_import::NoUnversionedImport),
+    Box::new(no_useless_fallback_in_spread::NoUselessFallbackInSpread),
     Box::new(no_useless_rename::NoUselessRename),
     Box::new(no_var::NoVar),
     Box::new(no_window::NoWindow),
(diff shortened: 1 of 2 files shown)


bartlomieju:

Clean port; flags `{ ...(foo || {}) }` / `{ ...(foo ?? {}) }`.

**Correctness — verified**
- **Scoping is elegant:** hooking `object_lit` means only *object-literal* spreads are checked, so array spreads (`[...(foo || {})]`), call-arg spreads (`call({}, ...(foo || {}))`), and destructuring patterns (`const {...foo} = obj`) are naturally excluded — all covered by valid tests.
- Operator gate is correct: only `LogicalOr`/`NullishCoalescing`, so `{...(foo && {})}` isn't flagged (matches oxc), and only the **right** operand is checked, so `{...({} || foo)}` isn't flagged.
- Empty-object check via `object.props.is_empty()` correctly treats `{...(foo || {...{}})}` (right object has a spread prop) and `{...(foo || {not: "empty"})}` as non-empty.
- `unwrap_parens` is applied to both the spread expr and the right operand, and non-`Bin` spread exprs (`{...(foo || {}).toString()}`, `{...fn(foo || {})}`, `{...(foo ? foo : {})}`) are correctly skipped.

**Perf:** trivial — per `object_lit`, linear over its props.

_No code changes requested — the `object_lit`-scoped approach cleanly handles every edge in the reference. Just tagging + rebase._

PR #1528, opened 2026-06-29
feat(no-useless-length-check): add `no-useless-length-check` rule

bartlomieju:
Ports the oxlint `unicorn/no-useless-length-check` rule to deno_lint as a
native rule.

The rule flags a redundant array length check that is combined with an
`.every()`/`.some()` call on the same array in a logical expression:

- `array.length === 0 || array.every(...)` — the `length === 0` check is
  useless because `Array#every()` already returns `true` for an empty array.
- `array.length !== 0 && array.some(...)` (and `array.length > 0 && ...`) — the
  non-empty check is useless because `Array#some()` already returns `false` for
  an empty array.

Matching oxc, the same identifier must be referenced in both the length check
and the method call, only the raw `0` literal counts, optional chains and
computed accesses are ignored, and chains of multiple `&&`/`||` operators are
flattened so every adjacent pair is checked. The diagnostic message and help
text mirror oxc exactly, and the complete pass/fail fixtures were ported (the
eslint-disable-specific commented-out cases were dropped).

Reference:
https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/unicorn/no_useless_length_check.rs

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -110,6 +110,7 @@ pub mod no_unsafe_negation;
 pub mod no_unused_labels;
 pub mod no_unused_vars;
 pub mod no_unversioned_import;
+pub mod no_useless_length_check;
 pub mod no_useless_rename;
 pub mod no_var;
 pub mod no_window;
@@ -358,6 +359,7 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(no_unused_labels::NoUnusedLabels),
     Box::new(no_unused_vars::NoUnusedVars),
     Box::new(no_unversioned_import::NoUnversionedImport),
+    Box::new(no_useless_length_check::NoUselessLengthCheck),
     Box::new(no_useless_rename::NoUselessRename),
     Box::new(no_var::NoVar),
     Box::new(no_window::NoWindow),
--- a/src/rules/no_useless_length_check.rs
+++ b/src/rules/no_useless_length_check.rs
@@ -5,7 +5,7 @@ use crate::handler::{Handler, Traverse};
 use crate::tags::{self, Tags};
 use crate::Program;
 use deno_ast::view::{
-  BinExpr, BinaryOp, Callee, Expr, Lit, MemberProp, NodeTrait,
+  BinExpr, BinaryOp, Callee, Expr, Lit, MemberProp, Node, NodeTrait,
 };
 use deno_ast::{SourceRange, SourceRanged};
 use derive_more::Display;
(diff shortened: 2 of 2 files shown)


bartlomieju:
**Verdict: needs changes** — duplicate diagnostics on nested same-operator chains. Not merging (new rule + bug).

The core logic (flatten a same-op chain, check adjacent pairs, match array name + `every`/`some` + raw `0`, exclude optional/computed) is correct and the exclusion tests are thorough. But there's a double-reporting bug.

**Correctness — blocking**
The `bin_expr` handler fires on **every** logical `BinExpr`, including the inner nodes of a chain, with no parent-operator guard. So for `A || B || C` = `((A || B) || C)`:
- the **outer** node flattens to `[A, B, C]` and checks pairs `(A,B)`, `(B,C)`;
- the **inner** node `A || B` flattens to `[A, B]` and checks `(A,B)` **again**.

When the first two operands are the useless pair, it's reported twice. Concrete repro (your own test):
`array.length === 0 || array.every(Boolean) || array.length === 0` — the leading `array.length === 0` is reported **twice at col 0** (the invalid test expects two `col: 0` entries plus one `col: 46`). A user would see the identical warning duplicated on the same code. The correct output is two diagnostics (first + third check), not three.

Note this means the ported test expectations **encode the bug** — they should be corrected to a single `col: 0` diagnostic once the handler is fixed (rather than left matching the duplicated output).

Fix — only process the root of a same-op chain by bailing when the parent is a logical `BinExpr` with the same operator (unwrapping parens), mirroring oxc's top-level-only handling:
```rust
// skip if this node is an operand of a same-op logical chain; the root handles it
let mut p = bin_expr.parent();
while let Node::Paren(_) = p { p = p.parent(); }
if let Node::BinExpr(parent) = p {
    if parent.op() == op { return; }
}
```
Then add/keep a 3-operand chain test asserting the de-duplicated count.

**Otherwise correct**
- `is_raw_zero` via `num.text() == "0"` correctly excludes `0.`, `0x0`, `"0"` (valid tests cover these).
- optional chain / computed / `notLength` / different-array-name / non-`every`,`some` exclusions all handled and tested.

**Perf:** `flatten` allocates a `Vec` per logical node; with the parent guard it's once per chain root, which is fine. Without the guard it also re-allocates for every sub-node — the fix helps here too.

**Housekeeping:** `RECOMMENDED` sign-off (you flagged it); `mergeable: UNKNOWN` → rebase.

_Blocker: add the same-op parent guard so chains report each useless check once, and fix the test expectations that currently bake in the duplicate._

PR #1529, opened 2026-06-29
feat(no-invalid-remove-event-listener): add `no-invalid-remove-event-listener` rule

bartlomieju:
Ports the `no-invalid-remove-event-listener` rule from oxlint's unicorn
plugin to a native deno_lint rule.

The rule flags calls to `removeEventListener` whose listener argument can
never match the listener originally passed to `addEventListener`, because it
is a freshly created value. Specifically, it reports when the second argument
is an inline arrow function, an inline function expression, or an inline
`.bind(...)` call. It only fires for static `removeEventListener` member
calls (e.g. `el.removeEventListener(...)`), including through optional member
access (`el?.removeEventListener(...)`), and ignores optional calls
(`el.removeEventListener?.(...)`), computed access, spread first arguments,
and listeners that are plain references or non-`.bind` calls. A trailing
options argument does not affect the result.

Reference:
https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/unicorn/no_invalid_remove_event_listener.rs

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -79,6 +79,7 @@ pub mod no_import_prefix;
 pub mod no_inferrable_types;
 pub mod no_inner_declarations;
 pub mod no_invalid_regexp;
+pub mod no_invalid_remove_event_listener;
 pub mod no_invalid_triple_slash_reference;
 pub mod no_irregular_whitespace;
 pub mod no_misused_new;
@@ -325,6 +326,7 @@ fn get_all_rules_raw() -> Vec<Box<dyn LintRule>> {
     Box::new(no_inferrable_types::NoInferrableTypes),
     Box::new(no_inner_declarations::NoInnerDeclarations),
     Box::new(no_invalid_regexp::NoInvalidRegexp),
+    Box::new(no_invalid_remove_event_listener::NoInvalidRemoveEventListener),
     Box::new(no_invalid_triple_slash_reference::NoInvalidTripleSlashReference),
     Box::new(no_irregular_whitespace::NoIrregularWhitespace),
     Box::new(no_misused_new::NoMisusedNew),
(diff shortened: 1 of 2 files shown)


bartlomieju:
**Verdict: LGTM — correct, with careful optional-chaining handling. Recommend merge** after the RECOMMENDED sign-off you flagged + rebase. Not auto-merging (new rule).

Flags `removeEventListener(..., <inline arrow|function|.bind()>)`. I verified the subtle parts:

**Correctness — verified**
- **Optional-chaining distinction is right:** `el?.removeEventListener(fn)` (optional *member*, non-optional call) is flagged — the `OptCall`'s parent `OptChainExpr.optional()` is `false`, so it isn't skipped and `member_from_expr` resolves the `OptChain(Member)` callee. Meanwhile `el.removeEventListener?.(fn)` (optional *call*) is skipped because the parent `OptChainExpr.optional()` is `true`. Both match the intended behavior and the tests.
- No double-processing: a plain call is a `CallExpr` (→ `call_expr`) and an optionally-chained one is an `OptCall` (→ `opt_call`); a node hits exactly one handler.
- Gating: non-computed `removeEventListener` prop, spread-first-arg bailout, `args.get(1)` (needs ≥2 args), and a trailing options arg correctly doesn't matter.
- Invalid-listener detection: `Arrow`/`Fn` inline, and `Expr::Call` where `is_bind_call` confirms a `.bind` member callee (`handler.bind(this)`). A bare reference or non-`.bind` call is correctly ignored.

**Minor / faithful edges**
- `is_bind_call` flags any `*.bind(...)` as listener regardless of receiver — matches oxc (any inline `.bind` yields a fresh function).

**Perf:** trivial (per call).

_No code changes requested — the optional-chain handling is the tricky bit and it's correct. Just tagging + rebase._

